PR #8, opened 2015-05-20
Fix function signature for cli.HelpPrinter

st3v:
Necessary due to github.com/codegangsta/cli/commit/e8425


asim:
Thanks for this.

PR #11, opened 2015-08-26
Change uuid library

kynrai:
Switch to using a different UUID library before code.google shuts down

I am not entirely sure on the version of UUID required by these methods but the proposed library supports them all. Whichever library is used, such as the one used in the rabbitmq module, code.google needs to change

### DIFF
--- a/broker/http/http.go
+++ b/broker/http/http.go
@@ -3,7 +3,7 @@ package http
 // This is a hack
 
 import (
-	"github.com/myodc/go-micro/broker"
+	"github.com/kynrai/go-micro/broker"
 )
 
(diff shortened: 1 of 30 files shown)


asim:
The uuid library was moved to https://github.com/pborman/uuid. So we can just switch to that quite easily. Also the PR includes 30 file changes for imports.

asim:
Switched to github.com/pborman/uuid which is basically just an import of the one from code.google.com

kynrai:
Yeah sorry, was changing on my fork but not in mown branch :P was only meant to submit the original change my bad

PR #18, opened 2015-11-11
add advertising option

asim:
server.Advertise allows you to specify a different address:port to the bind address to register with discovery.


crufter:
:+1: needed!

crackcomm:
:+1:

PR #19, opened 2015-11-13
Support for google pubsub

kenshiro-o:
- added support for Google [PubSub](https://cloud.google.com/pubsub/docs)
- added `Subscribe()` method to the `broker.Subscriber` interface (the interface was too asymmetric before)
- added comments to generate godoc


asim:
Google Pub/Sub should be added to go-plugins/broker. Go-micro will only include 3 implementations of each package interface in the library itself to keep downloads/importing/testing quick. registry/kubernetes was moved out to go-plugins for the same reason.

On the interface side. Not yet sure about the changes. Would appreciate feedback from others to gain a consensus. Was thinking about the use of Options in Publish and Subscribe rather than offloading to a secondary step of initialisation. broker.Subscribe(topic, handler, ...opts). Where opts could be broker.WithAckOption or something like that. Sort of leans towards the way the other packages are using options too. Thoughts?

kenshiro-o:
Yep I can move into the plugins. 
I'd actually suggest reducing go-micro further and only giving one implementation. There are loads of dependencies for a project that is supposed to be small.

I don't like the `Options` bit for this use case though. To me it sounds more like they should be used as part of the initialisation sequence of your subscriber. The ack phase happens on a per message basis. 

May actually write a pub/sub interface-only library. You could import it into go-micro in the future :stuck_out_tongue: .

PR #30, opened 2015-12-20
Experimental top level init

asim:
Top level initialisation. Syntactic sugar for all the underlying packages. Currently a WIP. Unsure of the interface and requirements.


go-micro.go
```diff
@@ -0,0 +1,47 @@
...
+	Server() server.Server
+	Run() error
+	String() string
+}
+
+type Option func(*Options)
+
+var (
+	HeaderPrefix   = "X-Micro-"
+	DefaultService = newService()
```
dancannon: Do you think this is needed? When would it be used?

service.go
```diff
@@ -0,0 +1,81 @@
...
+func (s *service) Client() client.Client {
+	return s.opts.Client
+}
+
+func (s *service) Server() server.Server {
+	return s.opts.Server
+}
+
+func (s *service) String() string {
+	return "go-micro"
```
dancannon, line 64: Is this correct? Should this be overridable?

dancannon:
I really like the look of this, helps make services much more testable.

go-micro.go
asim: The default service? Probably only useful if you just want to do things ad-hoc rather than initialisation a service.

service.go
asim, line 64: This is basically inline with all the other packages that implement an interface. Someone could implement a version of the service interface and call it something like "dancannon". The service name used for registration is actually embedded in the Server. Need to figure out a nice simple way to initialise that too.

go-micro.go
dancannon: Hmm ok, lately I have tried to avoid using default variables unless there is a good reason for them. As mentioned I find they make code very difficult to test.

service.go
dancannon, line 64: Cool :+1:

go-micro.go
asim: Depends on how you actually initialise. Defaults in go-micro don't normally result in go routines being spun up or anything magic. Its just a bit like DefaultServer and DefaultClient in net/http. It can be replaced, it can be left alone. It should not interfere but point noted. Want it to be elegant as opposed to cruft.

crackcomm: I'm with @dancannon on this one.

asim:
Removed the default service.

asim:
The interface still needs some work. Also getting screwed by initialisation. Setting stuff up in micro.NewService and then again in micro.Init seems like a bad idea. At the same time trying to integrate overriding with command line flags which proves to be a pain.

PR #33, opened 2015-12-23
Reverse the ping option

iheitlager:
Please reverse the ping option and disable it by default (or maybe even better to remove it completely).


asim:
Are you using go-micro?

iheitlager:
We are looking into it since yesterday and stumbled over this thingy

PR #50, opened 2016-02-24
Add Service in a Wrapper available to handlers

asim:
This adds Service within a wrapper making it available to handlers. So they can do `service, ok :=  micro.FromContext(ctx)`. Not yet sure if it should be added as it it delves into the realm of the user.

### DIFF
--- a/service.go
+++ b/service.go
@@ -26,9 +26,15 @@ func newService(opts ...Option) Service {
 		},
 	}
 
-	return &service{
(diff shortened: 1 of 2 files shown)


boyand:
I think this would be quite a good addition. We need that in many cases.

asim:
Backing this out for the moment. Doesn't feel like the right solution. The library should not be calling server.Init.

PR #49, opened 2016-02-24
Implement PortRange for transport.

jelmersnoeck:

PR #51, opened 2016-03-02
Internally handle server port range

asim:
This PR internally handles server port range for the HTTP transport. On further reflection of https://github.com/micro/go-micro/pull/49 it was clear that PortRange is a use case that needs to be solved but is very specific to the HTTP implementation. Adding a PortRange option did not feel like the right approach. Parsing the address allows existing --server_address or MICRO_SERVER_ADDRESS options to be set with "host:minPort:maxPort" without modifying any other code.

### DIFF
--- a/transport/http_transport.go
+++ b/transport/http_transport.go
@@ -5,11 +5,14 @@ import (
 	"bytes"
 	"crypto/tls"
 	"errors"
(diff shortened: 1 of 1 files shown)


jelmersnoeck:
:+1:

Issue #62, opened 2016-03-16
Consul warning: WARN Service tag "..." will not be discoverable via DNS due to invalid characters. Valid characters include all alpha-numerics and dashes.


asim:
This is a warning but an issue people continually bring up. Fix this so it no longer occurs. Most likely requiring encoding of tag values using gzip and hex.

basilex:
Hi!
Very quickly regarding the topic issue:
github.com/micro/go-micro/registry/consul_registry.go

[consul_registry.go.diff.gz](https://github.com/micro/go-micro/files/176173/consul_registry.go.diff.gz)

Probably we've to implement according to rfc2782?
Here's details: https://www.consul.io/docs/agent/dns.html

Alex

basilex:
Hi!
Very quickly regarding the topic issue:
github.com/micro/go-micro/registry/consul_registry.go

Probably we've to implement according to rfc2782?
Here's details: https://www.consul.io/docs/agent/dns.html https://www.consul.io/docs/agent/dns.html
Alex

asim:
The only comment Consul provides on this is that tags are opaque strings and must be of the alpha-numeric format. Converting to hex is probably going to be the easiest solution. I welcome any more appropriate solutions to convert JSON to something of a valid format.

asim:
https://github.com/micro/go-micro/pull/63

asim:
https://github.com/micro/go-micro/pull/63 now merged. Next step will be to stop registering old encoding entirely, while still decoding them if they exist. That should completely remove the warnings but will be the point at which any old service cannot read new services.

asim:
It's been sometime since the encoding fix was merged. In the next few weeks I will strip the old encoding so it removes the warnings.

asim:
now stripped https://github.com/micro/go-micro/commit/06ee80fd0a9594fc38e68d16d2e7a7e9b2ed0ecf

PR #63, opened 2016-03-16
Fix Consul WARN issues for tags

asim:
Fixes issue https://github.com/micro/go-micro/issues/62 BUT old clients cannot decode data for new apps which means version is hex encoded and endpoints/metadata is not returned.

### DIFF
--- a/registry/consul_registry.go
+++ b/registry/consul_registry.go
@@ -2,7 +2,6 @@ package registry
 
 import (
 	"crypto/tls"
-	"encoding/json"
 	"errors"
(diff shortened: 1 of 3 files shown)


asim:
Encoding updating to be backwards compatible. It will set both old and new formats and try to read whichever is available. As new builds of services are deployed the encoding will be updated. We'll strip out the old encoding code after a few months which will remove the Consul warnings and provide a non-breaking path to updating.

boyand:
:+1:

PR #67, opened 2016-03-23
return a formatted codec error

asim:


asim:
Not appropriate. We should not format errors as there's no indication as to whether the user is actually using our own error handling.

PR #69, opened 2016-04-01
HTTPTransport: implement queue for listener.

jelmersnoeck:
Before this patch, when an error occurs in trying to accept a connection
from the listener, the error would be returned. This also happened on
temporary issues like `too many open files`.

Temporary issues are "self-healing" and will resolve over time. This
means that we can put the requests in a queue to wait until the issue is
resolved and start processing the connections once it is resolved.

This patch implements such mechanism, as copied from the standard
library http package. It will retry temporary errors but will return
permanent errors (or errors that are not from the net.Error type).

### DIFF
--- a/transport/http_transport.go
+++ b/transport/http_transport.go
@@ -8,12 +8,14 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"log"
 	"net"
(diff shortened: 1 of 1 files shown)


asim:
LGTM

PR #70, opened 2016-04-05
Add Backoff func option

asim:
This enables backoff to be specified as a function to the client. The function is executed in the retry cycle. We execute before the first request for completely sake. This allows initial delays to be added or basically kicking out of the cycle before even beginning. There is however notable overlap in functionality between this and the Selector.

### DIFF
--- a/client/backoff.go
+++ b/client/backoff.go
@@ -1,0 +1,18 @@
+package client
+
+import (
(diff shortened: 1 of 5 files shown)


boyand:
Looks good to me!

PR #82, opened 2016-05-09
unmarshal a number into an interface{} as a Number

lcg635:


asim:
Does this even work? The build broke. Please add a test.

PR #83, opened 2016-05-10
unmarshal a number into an interface{} as a Number

lcg635:


asim:
From our discussion. You issue was more to do with the API returning float64 rather than int64. That's because its unmarshalling into map[string]interface{} in the api because it doesn't know what the response type is. I've changed this to json.RawMessage which preserves the number. https://github.com/micro/micro/commit/742fc8a7794d5e79092afab4ac4736ae671cb2dd

I would rather not make the change to the codec if that solves the issue.

PR #86, opened 2016-05-13
Connection Pooling

asim:
This is a PR for connection pooling. Its a WIP. There's some issues to deal with before this can be merged. Edge cases where if the request failed we might still attempt to use the connection. Need to battle test this and make it work as a stable thing.


client/rpc_client.go
```diff
@@ -73,10 +73,15 @@ func (r *rpcClient) call(ctx context.Context, address string, req Request, resp
 		return errors.InternalServerError("go.micro.client", err.Error())
 	}
 
-	c, err := r.opts.Transport.Dial(address, transport.WithTimeout(opts.DialTimeout))
+	var grr error
```
netixen: should be 'err'

PR #88, opened 2016-05-17
index always equal zero, always select first node

btfak:
current code index always equal zero, always select first node. so move them to global variable

### DIFF
--- a/selector/strategy.go
+++ b/selector/strategy.go
@@ -30,6 +30,11 @@ func Random(services []*registry.Service) Next {
 	}
 }
 
+var (
+	rrIndex = 0
+	rrMutex sync.Mutex
+)
+
 // RoundRobin is a roundrobin strategy algorithm for node selection
 func RoundRobin(services []*registry.Service) Next {
 	var nodes []*registry.Node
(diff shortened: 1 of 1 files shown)


asim:
This change is not ideal as it basically means the counter is incremented for ANY use of the strategy across any number of concurrent requests, not to mention across ALL services selected. This doesn't provide any additional benefit. RoundRobin strategy is setup so that when a request fails, on the next attempt it will try the next node. What you're maybe looking for is it actually distribute requests across calls to a service. I'll think about a better way but your change is unlikely to be merged as is.

btfak:
that's ok

asim:
I thought of a better solution. The problem is that we're always start within index 0 right. That's what you're saying. Well we can just start with an random index `idx := rand.Int() % len(nodes)` and then cycle through in round robin fashion. The behaviour we want out of the round robin strategy is to ensure we cycle through nodes one by one but that's not to say we need to start in the same place each time a new Next function is created for a pool.

Thoughts?

btfak:
sorry for my english, I guess I don't explain it clearly.
when we use round robin strategy,
if we run two nodes(A and B), we expect node serve request like this

```
A,B,A,B
```

A and B should both  serve half request.
and current strategy always select A, behave like this

```
A,A,A,A
```

asim:
I understand what you're saying but you have to understand the usage of the Strategy. When you do client.Call the selector retrieves a list of nodes for that service from the registry, then executes the Strategy function with that pool of nodes. That returns a Next function which is then called to retrieve a node from the list. If the call fails and retries are set it will cycle the pool by executing the next function again.

I understand your expectation of the strategy. Without it being a preinitialised pool, that behaviour wont occur but the pool of nodes is dynamic from the selectors standpoint as its retrieving a new list on every call, there's no state management within the Client. 

Your change basically added a global index that would be applied across every single service call ever made. So if you called Service A and Service B, the index would be 0 for service A, 1 for service B and then 2 for service A if you called that again. The behaviour isn't correct. The idea is that the strategy returns a pool that then cycles nodes.

If you have a suggestion for a better change, I'm interested to hear it. Otherwise my only thought would be to randomly set the index.

PR #95, opened 2016-05-28
MockClient.CallRemote now returns nil on success

ThatsMrTalbot:
MockClient.CallRemote was not returning nil on success as it should due to a missing return. This can cause tests to appear broken that are not.


asim:
Thanks

PR #101, opened 2016-06-08
Broker Codec implementation

asim:
This uses a codec rather than forcing json marshalling. Not entirely sure about this yet. It highlights some problems with the codec. The codec is a legacy artifact of net/rpc. It works but its not "clean".


asim:
Not merging. Unhappy with the interface.

PR #102, opened 2016-06-17
Add registry.OAuth2ClientCredentials option

st3v:
@asim: Not sure what you had in mind with regards to naming but what do you think about this? Happy to wait with this until https://github.com/hudl/fargo/pull/38 has been merged.

I've got a corresponding change in https://github.com/micro/go-plugins/tree/master/registry/eureka but wanted to get some feedback on the registry option first.

### DIFF
--- a/registry/options.go
+++ b/registry/options.go
@@ -8,10 +8,11 @@ import (
 )
 
 type Options struct {
-	Addrs     []string
-	Timeout   time.Duration
-	Secure    bool
-	TLSConfig *tls.Config
+	Addrs                   []string
+	Timeout                 time.Duration
+	Secure                  bool
+	TLSConfig               *tls.Config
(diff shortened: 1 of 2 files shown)


asim:
Hey Stev. The naming is fine. I was thinking the implementation should live in the eureka package itself since it's directly related to that and not all registries. You can see that Options has a Context field which can be used to add extra options without having to modify it further. Here's an example of how it's used in the gossip registry https://github.com/micro/go-plugins/blob/master/registry/gossip/options.go and how it's extracted https://github.com/micro/go-plugins/blob/master/registry/gossip/gossip.go#L474

st3v:
I see, that makes sense. I'll pull it there over the weekend and submit a PR against the Eureka plugin. Thanks!

PR #108, opened 2016-07-04
add quorum registry option

phifty:


asim:
Go-micro options should not be modified. Extra options should be added using Context for the specific plugin. Here's an example from the eureka plugin https://github.com/micro/go-plugins/blob/master/registry/eureka/options.go#L19L29

PR #114, opened 2016-07-30
Use struct tags if available

asim:
This uses struct tags where available for naming of endpoint values. Only really applicable with protobuf generated types.


bkrahmer:
Looks good to me.  Thanks Asim!

PR #126, opened 2016-10-23
add set cmd description option

qAison:


asim:
cmd.Init triggers command line parsing. I'm sorry this cannot be supported in an Option.

qAison:
@asim Can be achieved in other ways?

asim:
https://github.com/micro/go-micro/blob/master/cmd/cmd.go#L204

You can set internal cmd.App.Usage.

qAison:
@asim Micro.NewService After initializing the instance, unable to update settings cmd.app.Usage.

asim:
```
command := cmd.NewCmd()
command.App.Usage = "whatever"

service := micro.NewService(
  micro.Cmd(command),
)
```

or

```
service := micro.NewService()
service.Options().Cmd.App.Usage = "whatever"
```

qAison:
@asim service.Options() is struct, non-pointer,so it can not be updated.

asim:
Then pass in your own command as shown.

qAison:
@asim Thank you for your enthusiastic response, you are the best!(_^__^_)

PR #130, opened 2016-10-29
RoundRobin strategy

qAison:
No need to start from the first.

### DIFF
--- a/selector/strategy.go
+++ b/selector/strategy.go
@@ -21,12 +21,12 @@ func Random(services []*registry.Service) Next {
 	}
 
 	return func() (*registry.Node, error) {
-		if len(nodes) == 0 {
+		if count := len(nodes); count == 0 {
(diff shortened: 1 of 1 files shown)


asim:
What's the reason for this change?

qAison:
``` go
services = { s1, s2, s3, s4 }
next := RoundRobin(services)
```

For example, there are 1000 different client first call, the first will be selected s1.
We can provide a random order list when initializing, to avoid the first hit the same focus on the server.

asim:
While the RoundRobin function should likely start with a different index each time, shuffling the nodes is not ideal, that's what a completely random function is useful for. Also not really a fan of the superficial change to the random function. Thanks for the change but I'll likely just set "i" to be a random integer.

qAison:
Yes, that would be simpler.

PR #133, opened 2016-11-03
Adding the ability to specify a function to check if micro should retry a failed rpc call

ynori7:
I've run into the issue that micro is always retrying failed requests even if the error is on the client's side (such as when you provide a bad ID and the service returns a 404). It makes no sense to retry these, so I've added an additional option to specify a checkIfRetriable function, so now you can say something like this:

```
rsp, err := repo.personReadClient.FindPersonByID(
	ctx,
	req,
	microclient.WithRetries(client.DefaultRetries),
	microclient.WithRequestTimeout(client.DefaultRequestTimeout),
	microclient.WithCheckIfRetriable(microclient.Only500Errors),
)
```

By default it uses the AlwaysRetry function which just returns true (and therefore does the same thing which is currently done).


asim:
I've spent a little while thinking about this. While I understand the problem, the solution proposed is not something I'm sure of at the moment. There's a long standing issue of needing more fine grained control over the call functionality. While I've added retries and backoff, it feels like every other thing is a little hack on top. Ideally this should be addressed by providing a configurable `CallFunc` and having the current method be the default behaviour. I need to spend a little bit more time thinking about it.

If this was to be merged the naming would have to change. 

DefaultCheckIfRetriable = DefaultRetry
WithCheckIfRetriable = WithRetry
IsRetriableFunc = RetryFunc

Only500Errors would have to be removed and the AlwaysRetry be a private function.

asim:
Ok looks better. I can see how this will be useful but let's address a couple things.

RetryFunc is more useful if it looks like BackoffFunc:
`type RetryFunc func(context.Context, Request, int, error) (bool, error)`

RetryFunc should take the context, request, retry number and error. It should return bool for whether the request should be retried or not but also an error. Returning an error allows the RetryFunc to customise the response back to the caller if need be or indicates whether an error occurred in the retry logic. 

RetryFunc should also be executed inside the Stream method.

ynori7:
I just pushed another change, is that what you had in mind?

client/rpc_client.go
```diff
@@ -400,7 +405,12 @@ func (r *rpcClient) Stream(ctx context.Context, request Request, opts ...CallOpt
 			if rsp.err == nil {
 				return rsp.stream, nil
 			}
-			grr = rsp.err
+
+			if retry, err := callOpts.Retry(ctx, request, i, rsp.err); !retry {
+				return nil, err
+			}
+
+			grr = err
```
asim: What error is this using? Why not rsp.err?

asim:
Yea looks fine. Just one comment about stream error. 

Little confused by the logic around checking the function response. Idiomatic go method is to check error value.

```
case err := <-ch:
    // if the call succeeded lets bail early
    if err == nil {
        return nil
    }

    retry, rerr := callOpts.Retry(ctx, request, i, err)
    if rerr != nil {
        return rerr
    }

    if !retry {
        return err
    }

    gerr = err
}
```

In this case alwaysRetry should return true, nil

ynori7:
I adjusted the logic around the returned error from the retry function now and added the comment to the retry.go file to make it clear that returning either true or a non-nil error will result in the call not being retried.

client/rpc_client.go
```diff
@@ -299,6 +299,17 @@ func (r *rpcClient) Call(ctx context.Context, request Request, response interfac
...
+			retry, rerr := callOpts.Retry(ctx, request, i, err)
+			if rerr != nil {
+				return rerr
+			}
+
+			if !retry {
+				return err
+			}
+
+
```
asim: Unnecessary extra space

client/rpc_client.go
```diff
@@ -400,7 +411,17 @@ func (r *rpcClient) Stream(ctx context.Context, request Request, opts ...CallOpt
...
+			retry, rerr := callOpts.Retry(ctx, request, i, rsp.err)
+			if rerr != nil {
+				return nil, rerr
+			}
+
+			if !retry {
+				return nil, rsp.err
+			}
+
+			grr = err
```
asim: This should be `grr = rsp.err`

asim (commented):
Please run go fmt after addressing the comments.

asim:
Looks good. Thanks for the contribution.

PR #138, opened 2016-11-12
Use once instead of chan on service init and optimize

amorwilliams:
1.Use `sync.Once` instead of `chan` for service init once, just better.
2.I don't understand why create a temp options to keep cmd instance, it still will be override in the loop

### DIFF
--- a/service.go
+++ b/service.go
@@ -63,13 +63,9 @@ func (s *service) Init(opts ...Option) {
 		inited = true
 
 		// We might get more command flags or the action here
-		// This is pretty ugly, find a better way
-		options := newOptions()
-		options.Cmd = s.opts.Cmd
 		for _, o := range opts {
-			o(&options)
+			o(&s.opts)
 		}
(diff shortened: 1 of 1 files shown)


asim:
"Just better" is not a valid reason for this change. It's just style preference.

amorwilliams:
First commit is ok. but second. Could you explain for me? thanks

asim:
Options contain a lot of nested interfaces. Init provides a way to set some of those Options but in certain cases it cannot be done cleanly due to the nature of how the entire library is built. Starting with a new set of options then cleanly applying changes to that works better. There's also the ability to additionally add flags and arguments related to command line parsing at this stage so again best way to handle this is preserving the existing cmd and making it part of the new options.

amorwilliams:
I got it, here only to set the cmd interface, regardless of other parameters. Setting and overriding of other parameters  is at the last for loop, so here need a  clean options for cmd.I was so stupid.

PR #139, opened 2016-11-17
Add ability to pass consul client config via registry options.

simplechris:
As discussed in the slack channel earlier:

This can be used when you need to setup the internal consul client to use a specific configuration (E.G to pass a Token if the consul cluster is using ACL policies)

### DIFF
--- a/registry/consul/options.go
+++ b/registry/consul/options.go
@@ -1,0 +1,16 @@
+package consul
+
+import (
+	consul "github.com/hashicorp/consul/api"
+	"github.com/micro/go-micro/registry"
+	"golang.org/x/net/context"
+)
+
+func Config(c *consul.Config) registry.Option {
+	return func(o *registry.Options) {
+		if o.Context == nil {
+			o.Context = context.Background()
+		}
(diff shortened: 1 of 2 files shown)


registry/consul/options.go
```diff
@@ -0,0 +1,25 @@
...
+		o.Context = context.WithValue(o.Context, "consul_config", c)
+	}
+}
+
+func Token(t string) registry.Option {
+	return func(o *registry.Options) {
+		if o.Context == nil {
+			o.Context = context.Background()
+		}
+		o.Context = context.WithValue(o.Context, "consul_token", t)
```
jelmersnoeck, line 23: Would it be worth appending that to a default config and then use the `Config` method as above? This way you only have 1 key that does this work.

~~On the key part, would it be feasible to add a `NewContext` and `FromContext` method (like with metadata) which sets/gets the config from a private key (Have a look [here](https://github.com/micro/go-micro/blob/master/metadata/metadata.go#L7))?~~

_Edit: nvm, the key part won't work otherwise you'll have to refer to this package from the registry._

asim:
Looks good. Thanks

PR #140, opened 2016-11-21
consul registry configuration

simplechris:
Along the lines of: https://github.com/micro/go-micro/commit/8a25723fe07c02d47308a32e629b2fe11184c91a, which added a `Token` helper method to configure a consul registry. 

For completeness, this PR adds the remaining methods for configuring the [consul configuration struct](https://github.com/hashicorp/consul/blob/c744792fc4d665363dba0ecfc7d05fdedc9cab32/api/api.go#L126..L150), should they differ from the default settings. (`Address`, `Scheme` `Datacenter` etc). 

https://github.com/micro/go-micro/pull/139 *did* add the `Config` method enabling you to pass the complete config object, but for consistency these individual methods can be used in the same manner as the `Token` helper.

Something along the lines of:

```go

    import(
        "github.com/micro/go-micro"
        "github.com/micro/go-micro/registry/consul"
    )

   // .....

    service := micro.NewService(
        micro.Name("example"),
        micro.Version("latest"),
        micro.Registry(
            consul.NewRegistry(
                consul.Address("127.0.0.1:9000"), // consul running on a non-standard port
                consul.Token("TOKEN_HERE"),       // consul ACL token
            ),
        ),
    )
```

### DIFF
--- a/registry/consul/options.go
+++ b/registry/consul/options.go
@@ -1,6 +1,9 @@
 package consul
 
 import (
+	"net/http"
+	"time"
+
 	consul "github.com/hashicorp/consul/api"
 	"github.com/micro/go-micro/registry"
 	"golang.org/x/net/context"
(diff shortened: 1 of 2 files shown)


asim:
It seems likely redundant to do this considering Config can be set. Are you using these settings individually?

simplechris:
Yeah, I agree that this may be redundant or undesirable . 

The only use-case I've ran into is setting the token. I can imagine some folk may run consul on a non-default address, so I guess that could be common(ish) too.

You can indeed pass the config which is probably the best way to do things, but where I (selfishly) snuck the `Token` helper in the last PR for my use-case (that saved me a few lines for a slightly cleaner code and avoided importing the consul api package) and allowed me to do this:

```go
import(
    "github.com/micro/go-micro"
    "github.com/micro/go-micro/registry/consul"
)

// ...

micro.Registry(
    consul.NewRegistry(
        consul.Token("TOKEN_HERE"),
    ),
)
```

over this:

```go
import(
    "github.com/micro/go-micro"
    "github.com/micro/go-micro/registry/consul"
    consulapi "github.com/hashicorp/consul/api"
)

// ...

consulconf := consulapi.DefaultConfig()
consulconf.Token = "TOKEN_HERE"

micro.Registry(
    consul.NewRegistry(
        consul.Config(consulconf),
    ),
)
```

Where I'd introduced a 'individual' way of solving my configuration use-case for `Token` I was just adding the remaining helpers for the settings that I wasn't using, but for the sake of consistency.

Upon reflection, this arguably just adds unwanted bloat. A different way of achieving the 'consistency' I was aiming for is removing the `Token` helper I introduced (and *not* adding all of the other helpers), and just keep the `Config` method as the single way of passing in non-standard config options.

asim:
I would prefer to minimise bloat. I appreciate that being able to pass a single option is cleaner than setting up the consul config but this also means if any new fields are added to the config, maintaining parity would mean adding more options here.

simplechris:
@asim Yeah. I agree - makes complete sense. 

Should I keep the `Token` method introduced in #139, or should I update this PR to remove it (and all of the other individual helpers) to make the `Config` helper the way to manually configure?

asim:
Please close this PR and create a new one to just strip the Token option. Don't want to add unnecessary commits. Appreciate the effort though. You've demonstrated the process for configuring a secure Consul.

simplechris:
Closing in favour of #141

PR #148, opened 2017-01-19
Add Qos to broker

peti2001:
Right now it is not possible to fetch only one element from the queue and continue after the message has been acknowledged. The RabbitMQ supports this with Qos where we can define the number of messages that will be fetched without acknowledgment. And we also have set the AutoAck to FALSE.
I also had to change the createSubHandler function to send the acknowledgment after the handler has been executed if the AutoAck is false. I had no better idea where to do this. Actually like this it is still auto-acknowledged but done by go-micro and only after the handler.
If you need a queue without Qos you can do it like this:
```
service.Server().Subscribe(
		service.Server().NewSubscriber(
			"topic.go.micro.srv",
			subscriber.RequestHandler,
			server.DisableAutoAck(),
			server.Qos(2, 0),
		),
)
```

### DIFF
--- a/broker/options.go
+++ b/broker/options.go
@@ -32,6 +32,8 @@ type SubscribeOptions struct {
 	// will create a shared subscription where each
 	// receives a subset of messages.
 	Queue string
+	PrefetchCount int
+	PrefetchSize int
 
 	// Other options for implementations of the interface
 	// can be stored in a context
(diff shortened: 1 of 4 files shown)


asim:
Discussion occurring in Slack but this change will not be accepted. The broker is meant as a high level abstraction, this adds QoS specifics for RabbitMQ.

PR #149, opened 2017-01-20
Add ack() after executing the subscriber handler

peti2001:
Right now we cannot catch the event when a subscriber handler finishes the execution. Something would be nice to solve this. Here I used the broker.Publication(), but maybe something else that allow us to inject our code to handle this event would be even better.


asim:
Acking cannot be supported in server subscribe due to the fact that multiple handlers can handle a single message in parallel. For acking users should use the broker directly.

PR #152, opened 2017-03-01
allow micro framework to query and return a thirdparty service

gjrtimmer:
From discussion on slack.

The micro framework currently cannot return services which are not registered by ```micro```.
However within environments which uses the micro framework, it can be usefull or expected that I go-micro service has to use a already configured service within the consul registry.

Example:
Postgres service registered within the consul registry by its own agent; with the following consul config.
```json
{
    "datacenter": "datacore",
    "leave_on_terminate": true,
    "start_join": [ "172.30.0.2" ],
    "retry_join": [ "172.30.0.2" ],
    "node_name": "micro.db",
    "bind_addr": "0.0.0.0",
    "client_addr": "0.0.0.0",
        "service": {
                "name": "micro.db",
                "tags": [ "primary" ],
                "port": 5432,
                "checks": [{
                        "id": "pg_alive",
                        "notes": "Check connection",
                        "script": "/usr/bin/check_postgres --dbname micro --action connection",
                        "interval": "10s"
                }]
        }
}
```

A micro service could use this to create a database connection url and connect to the database.

Within the current code a service is excluded from the registry.GetService(...) when a 'version' tag is missing; however it might be because it is a third party service.

Furthermore; when switching to the go-micro framework, it wouldn't be uncommon to have the go-micro service connect to `for example serveral Java micro services` by protobuf until all business logic have been moved to new go-micro services.

The ListServices() already returns all services within a consul registry; this PR simply updates GetService(...) to include all requested services even if they are non go-micro.

### DIFF
--- a/registry/consul_registry.go
+++ b/registry/consul_registry.go
@@ -215,9 +215,40 @@ func (c *consulRegistry) GetService(name string) ([]*Service, error) {
 		// address is service address
 		address := s.Service.Address
 
-		// if we can't get the version we bail
-		// use old the old ways
+		// if we can't get the version and still got a service
+		// we are probably dealing with a non-micro framework service
+		// return the service as a thirdparty service
 		if !found {
+			svc, ok := serviceMap[key]
+			if !ok {
+				svc = &Service{
+					Name:       s.Service.Service,
+					ThirdParty: true,
+				}
+				serviceMap[key] = svc
+			}
+
+			var del bool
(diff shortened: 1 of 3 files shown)


gjrtimmer:
Start of slack conversation: https://micro-services.slack.com/archives/general/p1488361908601230

Additional thread: https://micro-services.slack.com/conversation/general/p1488363950685960

PR #159, opened 2017-04-13
Lock metadata when reading or writing

Taik:
Addresses the error noted at https://github.com/golang/go/issues/18137:

```bash
fatal error: concurrent map iteration and map write
```

### DIFF
--- a/client/rpc_client.go
+++ b/client/rpc_client.go
@@ -17,18 +17,20 @@ import (
 )
 
 type rpcClient struct {
-	once sync.Once
-	opts Options
-	pool *pool
+	once   sync.Once
+	mdLock sync.RWMutex
+	opts   Options
+	pool   *pool
 }
 
 func newRpcClient(opt ...Option) Client {
 	opts := newOptions(opt...)
 
 	rc := &rpcClient{
-		once: sync.Once{},
-		opts: opts,
-		pool: newPool(opts.PoolSize, opts.PoolTTL),
+		once:   sync.Once{},
+		mdLock: sync.RWMutex{},
+		opts:   opts,
+		pool:   newPool(opts.PoolSize, opts.PoolTTL),
 	}
 
 	c := Client(rc)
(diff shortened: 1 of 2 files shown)


asim:
This is not ideally the way to fix this problem. Go-micro was optimised to avoid locking the metadata, instead reading from it and making a copy then operating on the copy and essentially avoiding the concurrent access issue and leaving it to the user to deal with that. A more ideal solution would be to define a sync.RWMutex on metadata.Metadata so that it could be locked by any user of metadata.

Taik:
@asim I agree, this was more or less a hack to bypass the problem.

What do you think about returning shallow copies of the metadata instead? Would that suffice?

asim:
What problem are you actually trying to solve? Concurrent access within the library or concurrent access within and externally? Within the library the simplest solution is to make a copy and only write changes to the copy. Immediately solves all problems. If you're talking about access both inside and outside the library then that requires a public lock on metadata that everyone must use.

Taik:
I'm writing server and client wrappers to trace incoming and outgoing RPC calls with Datadog's APM (basically their implementation of distributed tracing). I've modeled the wrappers similar to the [opentracing wrapper](https://github.com/micro/go-plugins/blob/master/wrapper/trace/opentracing/opentracing.go). 

While rolling out to production and doing some basic load tests, [panics](https://gist.githubusercontent.com/Taik/91007ab3899b6b576fdd302b882808af/raw/eb7a727ae9378a2fbc2d997a44e19cd4564bd1fd/panic.log) started happening around concurrent map iteration and map write.

Would the tracing wrapper be considered 'inside of the go-micro library'?

asim:
Does it give you line numbers?

Taik:
@asim yep, it's here: https://github.com/Taik/go-micro/blob/master/client/rpc_client.go#L64

(I'm linking to my fork as that's the version the panic is generated from).

asim:
It panics on your fork? So basically the change you've made does nothing? Its hard to go on this information. It sounds like you are personally modifying the metadata that the client reads from. Why not make a copy?

PR #166, opened 2017-05-11
Transport callbacks no longer silently recovering from error

mier85:
These days we had problems with some of our services that got an "unexpected EOF" error from endpoints on services using the micro framework. We couldn't find any log messages to what causes these unexpected EOFs. After some digging in the framework I found this spot with a TODO on it. Now we can give an option of a callback so we can control if we want to log the error or if we even want to restart the service etc.


asim:
Closing in favour of https://github.com/micro/go-micro/pull/167

PR #167, opened 2017-05-11
make use of micro/go-log

asim:
In favour of "libraries not logging" I've defined a universal log interface https://github.com/go-log/log to help solve this problem and a micro scoped logger https://github.com/micro/go-log for all micro libraries. This is the first PR to shift to using that logger. This change actually changes nothing since the underlying implementation still uses the stdlib log package. In the grand scheme of things people can freely set the logger using `micro/go-log.SetLogger`.

### DIFF
--- a/broker/http_broker.go
+++ b/broker/http_broker.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"log"
 	"math/rand"
 	"net"
 	"net/http"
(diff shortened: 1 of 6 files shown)


mier85:
Looking good. I don't know if  I am overlooking something, but I don't see that you are setting the DefaultLogger. So shouldn't the logger by default be the noopLogger and it won't log anything?

asim:
https://github.com/micro/go-log/blob/master/log.go#L10

PR #170, opened 2017-05-31
Improve the UX for users of the errors package.

svanharmelen:
By making the error helper functions variadic functions, you can pass either a string (like the existing behaviour) or a format and verbs that will then be formatted for you.

This doesn’t break any existing API’s, but it prevents people using this package to have to call `fmt.Sprintf` whenever they would like to format the error detail.

Also fixed as much linter warning as possible without making backwards incompatible changes...

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -1,14 +1,14 @@
-// Package errors is an interface for defining detailed errors
+// Package errors provides a way to return detailed information
+// for an RPC request error. The error is normally JSON encoded.
 package errors
 
 import (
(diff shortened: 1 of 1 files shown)


svanharmelen:
FYI I noticed Github is having some issues today, so I'm afraid this pending check will not be updated properly... It ended almost 10 minutes ago and passed the tests.

image

svanharmelen:
Thanks!

asim:
Thank you!

PR #171, opened 2017-05-31
Wait for requests to finish before closing transport

asim:


server/rpc_server.go
```diff
@@ -44,6 +46,7 @@ func newRpcServer(opts ...Option) Server {
 		handlers:    make(map[string]Handler),
 		subscribers: make(map[*subscriber][]broker.Subscriber),
 		exit:        make(chan chan error),
+		wg:          sync.WaitGroup{},
```
svanharmelen: It doesn't hurt, but this is not needed. The zero value of a `sync.WaitGroup` is the same as `sync.WaitGroup{}`.

PR #176, opened 2017-06-20
Use HTTP statuses for errors and add 422

gillesfabio:


asim:
I'm sorry but no new errors are being added to this package. The predefined ones are there as convenience. It's expected that users define their own bespoke errors as needed.

gillesfabio:
That's pretty clear. Thanks @asim. I close this PR.

BTW, this package should not be used in [examples](https://github.com/micro/examples/blob/master/booking/api/hotel/main.go#L23). This can be confusing for newcomers that can be tempted to use it everywhere in their projects and send PRs again :)

PR #178, opened 2017-06-23
 switch tip to 1.8.3

orvice:


asim:
Likely to change all the travis files to be the same and include 1.9.

PR #179, opened 2017-06-24
Add MessagePack broker codec

elthariel:
Hi,

This PR adds a message pack broker codec. I do understand that this probably should be in the go-plugins repo, but due to the way the msgp go implementation works (afaict, it needs to generate methods on the serialized data structure), this is not cleanly possible.

That being said, the serialization performance (as documented here https://github.com/alecthomas/go_serialization_benchmarks) are preaching for its inclusion into the project :)

Any comment is is highly appreciated :)

### DIFF
--- a/broker/broker.go
+++ b/broker/broker.go
@@ -1,6 +1,9 @@
 // Package broker is an interface used for asynchronous messaging
 package broker
 
+// This is required by the MsgPack broker codel
+//go:generate msgp -io=false -tests=false
+
 // Broker is an interface used for asynchronous messaging.
 // Its an abstraction over various message brokers
 // {NATS, RabbitMQ, Kafka, ...}
--- a/broker/broker_gen.go
+++ b/broker/broker_gen.go
@@ -1,0 +1,98 @@
+package broker
+
+// NOTE: THIS FILE WAS PRODUCED BY THE
+// MSGP CODE GENERATION TOOL (github.com/tinylib/msgp)
+// DO NOT EDIT
+
+import (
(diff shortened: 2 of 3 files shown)


asim:
This is unlikely to be merged. Like you said, plugins live in go-plugins. I'm not quite sure as to why it needs to be here. You say msgp has some sort of dependency which means it needs to be like this. Why? Why can't your codec just take the broker.Message, convert to a type of the same format and encode it?

elthariel:
MessagePack is supported in Go by this library: https://github.com/tinylib/msgp

It works by generating methods (MarshalMsg/UnmarshalMsg) on the structs your want to serialize. The upside is it avoids any kind of reflection, the downside is that it requires to touch the package where the struct lives. 

As for the struct duplication approach, and with my limited go knowledge, I assumed it would require copying the data from the broker.Message to my msgp_broker.MyMessageClone, and it felt quite inelegant ?

asim:
Yes it would involve copying the data. I will reiterate that it's unlikely this will be merged into go-micro itself. Go-micro is a set of interfaces with a default reference architecture. Plugins are where separately maintained.

elthariel:
If by 'unlikely', you mean 'no'. It's faster if you just say 'no' :)

I might go with the data copying approach, even if it defeats the purpose of having very efficient marshalling.

While I have a hold on you, at which path would you recommend I store this plugin in the go-plugins repo ? There isn't any broker codec plugin yet.

asim:
Create the directory go-plugins/broker/codec/msgp

elthariel:
Thanks

PR #189, opened 2017-07-11
server/rpc_codec: friendlier behavior if c.codec.Write fails

hlian:
When developing go-micro services, it is frequently possible to set invalid results in the response pointer. When this happens (as I and @trushton personally experienced), `sendResponse()` returns an error correctly explaining what happened (e.g. protobuf refused to encode a bad struct) but the `call()` function one above it in the stack ignores the returned error object.

Thus, invalid structs go un-encoded and the _client side times out_.

@trushton and I first caught this in our CI builds when we left a protobuf.Empty field uninitialized (nil) instead of setting it to `&ptypes.Empty{}`. This resulted in an `proto: oneof field has nil value` error, but it was dropped and became a terribly confusing client timeout instead.

This patch is two independent changes:

* In rpc_codec, when a serialization failure occurs serialize an error message, which will correctly become a 500 for HTTP services, about the encoding failure. This means rpc_codec only returns an `error` when a socket failure occurs, which I believe is the behavior that rpc_service is expecting anyway.

* In rpc_service, log any errors returned by sendResponse instead of dropping the error object. This will make debugging client timeouts less of a hassle.

### DIFF
--- a/server/rpc_codec.go
+++ b/server/rpc_codec.go
@@ -7,6 +7,7 @@ import (
 	"github.com/micro/go-micro/codec/jsonrpc"
 	"github.com/micro/go-micro/codec/protorpc"
 	"github.com/micro/go-micro/transport"
+	"github.com/pkg/errors"
 )
 
 type rpcPlusCodec struct {
@@ -96,7 +97,11 @@ func (c *rpcPlusCodec) WriteResponse(r *response, body interface{}, last bool) e
 		Header: map[string]string{},
 	}
 	if err := c.codec.Write(m, body); err != nil {
-		return err
+		c.buf.wbuf.Reset()
+		m.Error = errors.Wrapf(err, "Unable to encode body").Error()
+		if err := c.codec.Write(m, nil); err != nil {
+			return err
+		}
 	}
 
 	m.Header["Content-Type"] = c.req.Header["Content-Type"]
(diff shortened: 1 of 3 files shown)


server/rpc_codec.go
```diff
@@ -96,7 +97,17 @@ func (c *rpcPlusCodec) WriteResponse(r *response, body interface{}, last bool) e
 		Header: map[string]string{},
 	}
 	if err := c.codec.Write(m, body); err != nil {
-		return err
+		c.buf.wbuf.Reset()
+		m = &codec.Message{
```
asim: Considering codec.Message is defined just a few lines earlier and a coded should not modify the message in the Write method could you just do `m.Error = errors.Wrapf(err, "Unable to encode body").Error()`.

server/rpc_codec.go
```diff
@@ -96,7 +97,17 @@ func (c *rpcPlusCodec) WriteResponse(r *response, body interface{}, last bool) e
...
+			Method: r.ServiceMethod,
+			Id:     r.Seq,
+			Error:  errors.Wrapf(err, "Unable to encode body").Error(),
+			Type:   codec.Response,
+			Header: map[string]string{},
+		}
+		if err := c.codec.Write(m, struct{}{}); err != nil {
```
asim: Do you need to send an empty struct or can you just pass in nil? Nil seems cleaner if possible.

asim (commented):
Can you write a test for this please. Nothing fancy, just enough to check the success and error cases do what's expected.

asim:
LGTM

PR #199, opened 2017-08-21
support tcp check on consul

wlibo666:
support tcp check on consul


asim:
I'm sorry, this change will not be accepted. It adds consul specific options. Go-micro is a pluggable abstraction. Implementation specific details are not added in this way. Please see the Config option here https://github.com/micro/go-micro/blob/master/registry/consul/options.go which demonstrates how a consul specific change may be added.

PR #204, opened 2017-08-25
fixing typo

wzhliang:


asim:
top level statements must start with "Package". It should say "Package transport"

PR #205, opened 2017-09-20
only return healthy service with consul backend

wzhliang:
Currently it's returning all services from consul, including ones that are in `critical` state, which is not usable.

### DIFF
--- a/registry/consul_registry.go
+++ b/registry/consul_registry.go
@@ -203,20 +203,22 @@ func (c *consulRegistry) GetService(name string) ([]*Service, error) {
 	}
 
 	serviceMap := map[string]*Service{}
-	isHealthy := func(s *consul.ServiceEntry) bool {
-		for _, chk := range s.Checks {
-			if chk.Status != "passing" {
-				return false
-			}
-		}
-		return true
-	}
 
 	for _, s := range rsp {
 		if s.Service.Service != name {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


asim:
I don't understand what you're doing. Do you realise that a few lines down this feature already exists. https://github.com/micro/go-micro/blob/master/registry/consul_registry.go#L237L244

wzhliang:
hmmm. But I did get `critical` services in return from `getService()`. Let me check.

asim:
Have you tried debugging the existing code rather than just adding the exact same thing a few lines up?

wzhliang:
It looks like the service itself is already added to the list (`serviceMap`) and the existing check only removes `Nodes` from the service. Is this expected behavior?

asim:
So it should technically return a service with no nodes. The selector would then parse this and say no nodes available. What's the actual behaviour you're getting?

wzhliang:
Yes I do get a service with `Nodes == nil`. If that's the expected behavior, then let's forget about this. But returning a service with no nodes seems just complicating client code for more checks.

asim:
It's not complicating anything because the functionality is there and it works. We have the expected behaviour.

wzhliang:
Cool. Related question: do you think the `check.Status == "critical"` should be `!check.Status == "passing"`?

PR #206, opened 2017-09-28
Added Options() to registry interface

darren-west:
I have added the Options flag to the registry.

At the moment I cannot extract any registry option from the Registry implementation.
Yes if i set this option from command line I can get it via an action or env variable but if it is a default option I cannot.

For example if I do not set the registry address via command line the default consul implementation will use an address of 127.0.0.1:8500, so i could have a copy of this address in my service but what if I decide to switch to a different implementation that uses address x.x.x.x my code now needs to change. 

I have made the change in go-plugins see [here](https://github.com/darren-west/go-plugins). go-os would also need to be updated.

### DIFF
--- a/registry/consul_registry.go
+++ b/registry/consul_registry.go
@@ -17,7 +17,7 @@ import (
 type consulRegistry struct {
 	Address string
 	Client  *consul.Client
-	Options Options
+	opts    Options
 
 	sync.Mutex
 	register map[string]uint64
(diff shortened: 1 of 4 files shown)


asim:
Why are the registry options required?

Based on the discussion in Slack it looks as though you want the consul registry address so you can do some sort of querying against the consul agent. My advice would be to default that value to localhost:8500 and look for the value set by flags or env vars to override. 

Making this change essentially breaks any registry implementation out there.

Those are my initial thoughts.

asim:
Ok it's been a while since this has come up. Willing to accept the change if you can PR the additional update in go-plugins

PR #209, opened 2017-10-09
use sync.Once instead of chan

uffy:
sync.Once is more clear and faster than chan.


service.go
```diff
@@ -10,12 +10,13 @@ import (
 	"github.com/micro/go-micro/cmd"
 	"github.com/micro/go-micro/metadata"
 	"github.com/micro/go-micro/server"
+	"sync"
```
asim: please inline this with the other go stdlib packages

uffy: OK

asim:
"import sorting" is a completely incorrect comment

Issue #212, opened 2017-10-25
any things to do when Server.Register() err 


weisd:
https://github.com/micro/go-micro/blob/53554d98cd6f32ba7561391fe0d6fd743ec63e31/service.go#L47

asim:
Merged

PR #213, opened 2017-10-25
add log when register err

weisd:
#212


weisd:
@asim

asim:
Not so sure about this. Horrible comment. Might fix it myself.

asim:
Also will be one that floods the log if theres an error

weisd:
Any good ideas ？

Issue #218, opened 2017-11-07
micro can not support "protobuf oneof feature"?


hunterhug:
I just change the protobuf file in https://github.com/micro/examples/tree/master/greeter/srv:

```
syntax = "proto3";

package go.micro.srv.greeter;

service Say {
    rpc Hello (Request) returns (Response) {
    }
}

message Request {
    string name = 1;
    oneof a {
        OneOfRequest b = 2;
    }
}

message OneOfRequest {
    string c = 1;
}

message Response {
    string msg = 1;
}
```

the result:

```
XXXX:~/code/src/github.com/micro/examples/greeter$ micro get service go.micro.srv.greeter
service  go.micro.srv.greeter

version 1.0.0

Id      Address Port    Metadata
go.micro.srv.greeter-36211938-c385-11e7-90ec-509a4c048c1a       172.16.13.78    45708   transport=http,broker=http,server=rpc,registry=consul

Endpoint: Say.Hello
Metadata: stream=false

Request: {
        name string
        is_request__a isRequest_A
}

Response: {
        msg string
}
```

It seem that `oneof` feature can not support:

```
XXXX:~/code/src/github.com/micro/examples/greeter$ micro query go.micro.srv.greeter Say.Hello '{"name": "John","a":{"b":"test"}}'
error calling go.micro.srv.greeter.Say.Hello: json: cannot unmarshal object into Go struct field Request.A of type go_micro_srv_greeter.isRequest_A

XXXX:~/code/src/github.com/micro/examples/greeter$ micro query go.micro.srv.greeter Say.Hello '{"name": "John","a":{"b":{"c":"test"}}}'
error calling go.micro.srv.greeter.Say.Hello: json: cannot unmarshal object into Go struct field Request.A of type go_micro_srv_greeter.isRequest_A
```

how can I write the json? I try a long time....

but I do this, It works!
```
XXXX:~/code/src/github.com/micro/examples/greeter$ micro query go.micro.srv.greeter Say.Hello '{"name": "John","a":null}'
{
        "msg": "Hello John"
}
```

Micro can not support "protobuf  oneof feature"?

asim:
The oneof feature requires the use of [jsonpb](https://github.com/micro/protobuf/tree/master/jsonpb) which micro does not currently support.

hunterhug:
Thank you very much. We're preparing to use micro as our microservice framework. Many of our old services are using `oneof`(we use gRPC). It's a bit of a sad. When do you start supporting?

[...]

PR #220, opened 2017-11-18
Added jsonpb rpc codec (stub?)

tudurom:
I think that I nailed the jsonpb codec but I don't know how to test if it actually works.

Might be useful for #218


asim:
Can you find a way to test this please. You will have to set the content-type on the request to make use of the codec or set it as the default.

asim:
It's also likely better put into go-plugins

asim:
Can you please move this to go-plugins

tudurom:
Will move tomorrow

asim:
Cheers. Closing this then.

liilac:
Seeing as there hasn't been a PR made to go-plugins, I'm working on one at the moment

PR #221, opened 2017-11-20
add https support for consul

gaxxx:
fix issue that will never get a chance to setup tls even the environment variables set.
using enviroment variables

for example:
export CONSUL_HTTP_SSL=1
export CONSUL_HTTP_ADDR="https://example.com"
export CONSUL_CLIENT_CERT="/Users/foo/.ssh/consul/consul.cert"
export CONSUL_CLIENT_KEY="/Users/foo/.ssh/consul/consul.key"
export CONSUL_CACERT="/Users/foo/.ssh/consul/ca.cert"

### DIFF
--- a/registry/consul_registry.go
+++ b/registry/consul_registry.go
@@ -59,14 +59,6 @@ func newConsulRegistry(opts ...Option) Registry {
 			config = c
 		}
 	}
-	if config.HttpClient == nil {
-		config.HttpClient = new(http.Client)
-	}
-
-	// set timeout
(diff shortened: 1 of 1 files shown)


asim:
You've removed line 63 which does a nil check for the http client

gaxxx:
@asim Yes. 
consul.NewClient(config) will read the env variables and create a http client with correct tls configuration.
if we create a httpclient outside the consul.NewClient,  the consul.NewClient will ignore the env variables and use the httpclient which is incorrect.

PR #224, opened 2018-01-01
fix possible deadlock since code can return without unlocking the Mutex

dh1tw:
Hi,

If a subscriber already exists the code returns without unlocking the rpcServer's Mutex. This might cause a deadlock later on.


asim:
Thanks for the fix

PR #226, opened 2018-01-28
registry: export consul registry

tudurom:
Export the consul registry to allow direct manipulation of the consul client used by it for more complex use cases. This way, services can declare consul tags or use the K/V storage.

### DIFF
--- a/registry/consul_registry.go
+++ b/registry/consul_registry.go
@@ -14,7 +14,7 @@ import (
 	hash "github.com/mitchellh/hashstructure"
 )
 
-type consulRegistry struct {
+type ConsulRegistry struct {
 	Address string
 	Client  *consul.Client
 	Options Options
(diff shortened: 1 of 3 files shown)


asim:
This change basically breaks the current design pattern of only returning the registry interface. It's unlikely we do this. If you'd like to expose consul to do things specific to it, please write a plugin for that purpose.

tudurom:
Then I propose another change: add a function, say `GetClient() interface{}` that returns the registry's underlying client, if any. Then, depending on the registry, the client can type assert the result and do complex things.

tudurom:
Indeed the bring your own client is the better idea. The use case that I imagined is that I want a consul client connection to the same instance that micro is using but I don't want to deal [with the fuss of creating it](https://github.com/micro/go-micro/blob/master/registry/consul_registry.go#L48). Writing my own registry works but it's pointless duplicate code. The code wouldn't be duplicated if micro handles the initialization of the consul client, otherwise I might just as well write my own registry.

PR #233, opened 2018-03-03
switch to stdlib context

asim:
This switches to the stdlib context. As of Go 1.9 this is a non-breaking change. We'll always try support the two latest releases. Ideally we don't want to make breaking changes but users should be updating Golang versions.


EwanValentine (approved):
Looks good!

PR #238, opened 2018-03-16
Add option to enable TCP check with Consul registry

shuLhan:
One disadvantage of using TTL based health check is the high network
traffic between Consul agent (either between servers, or between server
and client).

In order for the services considered alive by Consul, microservices must
send an update TTL to Consul every n seconds (currently 30 seconds).

Here is the explanation about TTL check from Consul documentation [1]

    Time to Live (TTL) - These checks retain their last known state for a
    given TTL. The state of the check must be updated periodically over
    the HTTP interface. If an external system fails to update the status
    within a given TTL, the check is set to the failed state. This
    mechanism, conceptually similar to a dead man's switch, relies on the
    application to directly report its health. For example, a healthy app
    can periodically PUT a status update to the HTTP endpoint; if the app
    fails, the TTL will expire and the health check enters a critical
    state. The endpoints used to update health information for a given
    check are the pass endpoint and the fail endpoint. TTL checks also
    persist their last known status to disk. This allows the Consul agent
    to restore the last known status of the check across restarts.
    Persisted check status is valid through the end of the TTL from the
    time of the last check.

Hint:

    TTL checks also persist their last known status to disk. This allows
    the Consul agent to restore the last known status of the check
    across restarts.

When microservices update the TTL, Consul will write to disk. Writing to
disk means all other slaves need to replicate it, which means master need
to inform other standby Consul to pull the new catalog. Hence, the
increased traffic.

More information about this issue can be viewed at Consul mailing list [2].

[1] https://www.consul.io/docs/agent/checks.html
[2] https://groups.google.com/forum/#!topic/consul-tool/84h7qmCCpjg

### DIFF
--- a/registry/consul/options.go
+++ b/registry/consul/options.go
@@ -18,13 +18,13 @@ func Config(c *consul.Config) registry.Option {
 }
 
 //
-// RegisterTCPCheck will tell the service provider to check the service address
+// TCPCheck will tell the service provider to check the service address
 // and port every `t` interval. It will enabled only if `t` is greater than 0.
 // See `TCP + Interval` for more information [1].
 //
 // [1] https://www.consul.io/docs/agent/checks.html
 //
-func RegisterTCPCheck(t time.Duration) registry.Option {
+func TCPCheck(t time.Duration) registry.Option {
 	return func(o *registry.Options) {
 		if t <= time.Duration(0) {
 			return
@@ -32,7 +32,6 @@ func RegisterTCPCheck(t time.Duration) registry.Option {
 		if o.Context == nil {
 			o.Context = context.Background()
 		}
-		o.Context = context.WithValue(o.Context,
-			registry.ConsulRegisterTCPCheckKey, t)
+		o.Context = context.WithValue(o.Context, "consul_register_tcp_check", t)
 	}
 }
--- a/registry/consul/options.go
+++ b/registry/consul/options.go
@@ -32,6 +32,6 @@ func TCPCheck(t time.Duration) registry.Option {
 		if o.Context == nil {
 			o.Context = context.Background()
 		}
-		o.Context = context.WithValue(o.Context, "consul_register_tcp_check", t)
+		o.Context = context.WithValue(o.Context, "consul_tcp_check", t)
 	}
 }
--- a/registry/consul/options.go
+++ b/registry/consul/options.go
@@ -2,6 +2,7 @@ package consul
 
 import (
 	"context"
+	"time"
 
 	consul "github.com/hashicorp/consul/api"
 	"github.com/micro/go-micro/registry"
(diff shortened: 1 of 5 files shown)


asim:
This cannot be merged in its current form. This is a consul specific change. Consul specific options should be added here https://github.com/micro/go-micro/blob/master/registry/consul/options.go and then accessed through the Context.

Your check is also incorrectly setup. It specifies the consul address as the tcp check address. This makes no sense.

shuLhan:
Done.

registry/consul/options.go
```diff
@@ -15,3 +16,23 @@ func Config(c *consul.Config) registry.Option {
...
 }
+
+//
+// RegisterTCPCheck will tell the service provider to check the service address
+// and port every `t` interval. It will enabled only if `t` is greater than 0.
+// See `TCP + Interval` for more information [1].
+//
+// [1] https://www.consul.io/docs/agent/checks.html
+//
+func RegisterTCPCheck(t time.Duration) registry.Option {
```
asim: Change this to TCPCheck

registry/consul/options.go
```diff
@@ -15,3 +16,23 @@ func Config(c *consul.Config) registry.Option {
...
+	return func(o *registry.Options) {
+		if t <= time.Duration(0) {
+			return
+		}
+		if o.Context == nil {
+			o.Context = context.Background()
+		}
+		o.Context = context.WithValue(o.Context,
+			registry.ConsulRegisterTCPCheckKey, t)
```
asim: Put this on one line

registry/consul_registry.go
```diff
@@ -14,6 +14,10 @@ import (
 	hash "github.com/mitchellh/hashstructure"
 )
 
+const (
+	ConsulRegisterTCPCheckKey = "consul_register_tcp_check"
```
asim: Do not declare this. We don't need a const for this.

registry/consul_registry.go
```diff
@@ -125,6 +142,15 @@ func (c *consulRegistry) Register(s *Service, opts ...RegisterOption) error {
 		o(&options)
 	}
 
+	if c.opts.Context != nil {
+		tcpCheckInterval, ok := c.opts.Context.
+			Value(ConsulRegisterTCPCheckKey).(time.Duration)
```
asim: Put on one line

registry/options.go
```diff
@@ -18,7 +18,9 @@ type Options struct {
 }
 
 type RegisterOptions struct {
-	TTL time.Duration
+	TCPCheck bool
```
asim: We will not accept TCPCheck as an Option value here. Put the TCPCheck in the Context as with option Consul specific options.

registry/consul_registry.go
shuLhan: Three reasons why I use constant in my code are: (1) minimize typo, e.g. "consul_register_tpc_check"  (2) readability, (3) Easy on refactoring (e.g. changing from "consul_register_tcp_check" to "register_tcp_check" will not effect other files).

registry/consul/options.go
```diff
@@ -15,3 +16,22 @@ func Config(c *consul.Config) registry.Option {
...
+//
+func TCPCheck(t time.Duration) registry.Option {
+	return func(o *registry.Options) {
+		if t <= time.Duration(0) {
+			return
+		}
+		if o.Context == nil {
+			o.Context = context.Background()
+		}
+		o.Context = context.WithValue(o.Context, "consul_register_tcp_check", t)
```
asim: Rename to `consul_tcp_check`

registry/consul_registry.go
```diff
@@ -115,16 +115,39 @@ func (c *consulRegistry) Deregister(s *Service) error {
...
+	var regTCPCheck bool
+	var regInterval time.Duration
+
 	var options RegisterOptions
 	for _, o := range opts {
 		o(&options)
 	}
 
+	if c.opts.Context != nil {
+		if tcpCheckInterval, ok := c.opts.Context.Value("consul_register_tcp_check").(time.Duration); ok {
```
asim: Rename to `consul_tcp_check`

asim:
LGTM, will merge once tests pass. Thanks

shuLhan:
Thank you for accepting the PR and for this repository.

Issue #234, opened 2018-03-06
Handler result error is ignore


onrik:
https://github.com/micro/go-micro/blob/master/server/subscriber.go#L232
Currently it's impossible to handle return value from handler. Do you have some solution?

asim:
What return value? This is the subscriber. A subscriber does not return anything

onrik:
Function [fn](https://github.com/micro/go-micro/blob/master/server/subscriber.go#L232) returns error, but it's handle nowhere.

asim:
What should be the appropriate behaviour? Please PR the change

asim:
Now fixed with this commit https://github.com/micro/go-micro/commit/b39ec4472c0b6da5bcb6c80ab7acabe1da9d4e20

PR #241, opened 2018-03-21
Fix issue #234

onrik:
Check handler result error.


asim:
By removing the waitgroup the server can no longer support graceful shutdown. By inlining the function it will now execute multiple method sequentially. I'm not sure this is the best fix.

onrik:
Ok. I'll rework it.

asim:
Closing until something better is PRed

PR #248, opened 2018-04-14
Rework Interfaces

asim:
This is a long overdue rework of go-micro interfaces. 

What's changing/needs to change:

- simplified client/codec
- server with ServeRequest handler


simplechris:
much cleaner 🎉

👍

Issue #267, opened 2018-06-13
Deadlock issue, when HTTPBrocker Connect is called.


jasimmk:
I was working on a pub sub  poc, 
a. `server` (single binary) is responsible for generating a proto payload and send to `worker` via a topic
b. `worker` (single binary) is responsible for subscribing to the topic, accept the payload, regenerate another payload and publish it
c. `server` is subscribed to the second type of events generated by `worker`

While testing locally, using mDNS registry, Events are published from `server` and received at `worker` properly, But at the moment,  publish second event, from `worker` to `server`, deadlock happens.

On debugging, could see 

```
(h *httpBroker) ServeHTTP(w http.ResponseWriter, req *http.Request)
      .......
	h.RLock()
	for _, subscriber := range h.subscribers[topic] {
		if id == subscriber.id {
			// sub is sync; crufty rate limiting
			// so we don't hose the cpu
			subscriber.fn(p)
		}
	}
	h.RUnlock()
```
is creating a read lock, and `RUnlock` never happens

while 
```
func (h *httpBroker) Connect() error {
	h.Lock()
	defer h.Unlock()

	if h.running {
		return nil
	}
```
Creates a full lock, and waits forever to release the read locks.

<img width="1680" alt="deadlock" src="https://user-images.githubusercontent.com/759996/41349989-a410e300-6f22-11e8-8fc4-4a093de38ed9.png">

asim:
Merged your PR.

PR #268, opened 2018-06-13
Fixing httpBroker dead lock; If publish is called from a subscription #267

jasimmk:
Solving Issue  #267 

Due to a read lock from other threads, and connect is trying to create a write lock.

We can do `broker` is running in a read lock, and if the broker is already connected, it won't do the write locks, and simply exits the method.

Implemented similar logic for `Disconnect` and `Init`  methods as well.

### DIFF
--- a/broker/http_broker.go
+++ b/broker/http_broker.go
@@ -276,12 +276,16 @@ func (h *httpBroker) Address() string {
 }
 
 func (h *httpBroker) Connect() error {
-	h.Lock()
-	defer h.Unlock()
 
+	h.RLock()
 	if h.running {
+		h.RUnlock()
 		return nil
 	}
+	h.RUnlock()
+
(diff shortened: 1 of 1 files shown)


asim:
Thanks for the update. Out of curiousity, is go-micro being used at Careem?

jasimmk:
`go-micro` provides necessary options for running distributed applications, with simple integration, without much-complications. Thanks for the library.  At present  We are evaluating some option to include `golang` and `go-micro` at our SRE stack. Business critical applications are mostly on Java though.

asim:
Good to know. Careems a big company so if you do decide to use it would be really cool to have you as a sponsor https://www.patreon.com/microhq/memberships. Otherwise I'm launching support subscription soon if that would be a useful option.

PR #278, opened 2018-07-20
set DefaultRetries to 0

jiyeyuran:
By default, the client should not retry when error happens.


asim:
I know I broke the behaviour. I did it on purpose. If a request succeeds there's no retry, if it fails we attempt it when it's under the request timeout. This is much like the behaviour of nginx or other web servers that try to retry if there's a failure. I don't know if I plan to remove this

PR #279, opened 2018-07-20
Retries can be 0

jiyeyuran:
Giving a chance to set retries to 0. The errors from micro service may not be just some network error, but also created by application. For example, we notify user the password is wrong by returning error. In this situation, I don't want the gateway to request twice.


asim:
What if it's a negative number? The logic should stay the same just >=

asim:
Did you gofmt?

jiyeyuran:
I editted the file using the online editor.

PR #288, opened 2018-08-01
Missing default values in CMD

sysradium:
Because right now I get 0 retries by default which causes an issue with GRPC.

The GRPC problem has been fixed here https://github.com/micro/go-plugins/commit/4ff3c6956c31bbfb99558f99a87871d60ef198f8#diff-210c773dbf19e356c9da5ee119131afb

However I still think the defaults have to be set.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -56,26 +56,31 @@ var (
 			Name:   "client",
 			EnvVar: "MICRO_CLIENT",
 			Usage:  "Client for go-micro; rpc",
+			Value:  "rpc",
 		},
 		cli.StringFlag{
 			Name:   "client_request_timeout",
 			EnvVar: "MICRO_CLIENT_REQUEST_TIMEOUT",
 			Usage:  "Sets the client request timeout. e.g 500ms, 5s, 1m. Default: 5s",
+			Value:  "5s",
 		},
 		cli.IntFlag{
 			Name:   "client_retries",
 			EnvVar: "MICRO_CLIENT_RETRIES",
 			Usage:  "Sets the client retries. Default: 1",
+			Value:  1,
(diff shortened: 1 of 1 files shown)


asim:
This doesn't make sense. When the client is created all the defaults are set https://github.com/micro/go-plugins/blob/master/client/grpc/grpc.go#L463L471

sysradium:
@asim when you call `service.Init()` it calls `cmd.Init(...)` which resets the values.
Basically:
```golang
fmt.Println(service.Client().Options().CallOptions.Retries) // 1
service.Init()
fmt.Println(service.Client().Options().CallOptions.Retries) // 0
```

asim:
The inconsistencies with value setting should be resolved but we should do it in one place. I prefer to do it in the default options and then remove any flag vars. Do you agree?

sysradium:
@asim fine. However it looks just a little bit weird that you don't use your own flags as others should use them :)

And this confused me:
https://github.com/micro/go-micro/blob/1d2e0accdd85172ccb23992ab49fbb701fbec867/cmd/cmd.go#L123

But I think that's an exception.

asim:
That's most definitely an exception to avoid nil pointer errors.

asim:
Also, we personally don't personally set flag values because this should be left to the user as an override. Options should be default values set in code. Flags and env vars let you override at runtime.

asim:
If you want to fix this crap, more than welcome. Basically set all flags to nothing. Set all default options to sane values. Ensure it works. Profit.

asim:
Ok so looking at the code now, all flag values are not set. The metadata flag specifies an empty string slice to avoid panics. Otherwise everything looks fine. Closing issue.

PR #289, opened 2018-08-04
http2 support

asim:
The first steps towards supporting http2 by default


asim:
This PR merged server side support for h2 and h2c. If the server encounters http/1 requests it will serve as expected. h2c upgrade is available and h2 also. Client side will be a difficult thing to solve.

Issue #283, opened 2018-07-24
Consul service mesh


alexbilbie:
I'm just wondering if there any thoughts/plans to integrate with Consul's new service mesh (released in Consul v1.2)?

As an alternative, have you got any recommendations on how a permission layer could be integrated with Micro so service X could be given permission to call method Y on service Z?

asim:
There is plan to integrate consul connect, it's just unclear what the best path forward is right now. 

Consul connect requires configuration in 3 ways. 

1. A service must register with consul specifying that it's a connect enabled application
2. A service must use the connect api to enable mutual tls authentication
3. A service must use the connect api to retrieve other connect services

I believe given service mesh is becoming a central technology we must define an abstraction for it. While service mesh is supposed to operate transparently, I think only the data plane offers this capability. In the event that we need to integrate with a control plane something must be specified.

I propose a new abstraction or potential option "proxy" whereby we proxy a number of pieces of functionality through the specified system. In the case of connect we'd enable this by the flag `MICRO_PROXY=connect` or flag `--proxy=connect`.

I don't believe this is a package level definition in go-micro because I think go-micro itself has a focus on solving these distributed systems problems as a library. I do believe this is a higher order function that we could define at the top-level alongside Service and Function. 

Interested to get feedback on this.

```
// Proxy defines a type which configures go-micro to use service mesh
type Proxy interface {
        // Creates a new proxied micro service
        Service(opts ...Option) Service
        // name of proxy e.g micro, connect, envoy
        String() string
}
```

asim:
Something in the pipeline https://github.com/micro/go-micro/pull/295

Closing the issue for now

PR #295, opened 2018-08-14
Proxy interface

asim:
https://github.com/micro/go-micro/issues/283


asim:
Due to some missing fields in crypto/x509.CertificateRequest we may not be able to merge this until we can support 2 independent versions of Go.

PR #310, opened 2018-10-03
Allow new subscriptions after initial registration.

norganna:
# Description

Currently while subscriptions can be attempted after initial registration of the service without issue, the subscription never actually happens because of a check of the `registered` property.

The next call to `Register()` then publishes the subscription in the endpoints list to the registry, even though the subscription is not actually subscribed.

When the server exits, the endpoint is subsequently deregistered, so all this works as expected, apart from the fact that the application isn't actually receiving anything on the subscription because it's never created.

The following is a minimal test case that shows the issue:
https://gist.github.com/norganna/763ffe795602f37fdc6d16cf78378301

# Proposal

It seems that it would be simple to allow subscriptions after first registration by checking if registered and then simply adding the subscription.

At the next `RegisterInterval` the registration will be updated with the correct endpoints.

The MR as attached will create the subscription and add it to the `subscriptions` map.

### DIFF
--- a/server/rpc_server.go
+++ b/server/rpc_server.go
@@ -185,16 +185,14 @@ func (s *rpcServer) Subscribe(sb Subscriber) error {
 	config := s.Options()
 
 	s.Lock()
+	defer s.Unlock()
 	_, ok = s.subscribers[sub]
 	if ok {
-		s.Unlock()
 		return fmt.Errorf("subscriber %v already exists", s)
 	}
 	s.subscribers[sub] = nil
-	defer s.Unlock()
 
-	registered := s.registered
-	if registered {
+	if s.registered {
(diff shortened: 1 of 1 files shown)


asim:
So I've given this thought and I think the behaviour should actually be simplified to error on registration of a handler or subscriber where a server is already running. This feels like a side effect that we should really not allow. Otherwise the more appropriate behaviour would be to flip the registered flag if any server changes actually occur. I'm more in favour of the first option because its the clean approach of setup before running. Post run subscription seems like a bad idea.

asim:
Closing.  I think the fix we need is to prevent handler and subscriber registration after the server is running. I think this is a simpler and saner approach. If you'd like to contribute that please PR separately. Thanks.

PR #311, opened 2018-10-05
Update Readme

andyfen:
should client```NewGreeterService``` be ```NewGreeterClient```  ?


asim:
No because if you used protoc-gen-micro you would know this is correct

PR #313, opened 2018-10-09
Fix tcp check no ttl error

lovelly:
Register too many references, otherwise you can consider using 
image
to wait for the state change of the service


asim:
This fix isn't really useful. You're now calling the Checks method for no reason. Considering your issue, in the case of TTL not being set for tcp check should we not just return nil if its a tcp check?

asim:
Please `go fmt`

PR #333, opened 2018-11-22
Only check if the service is in Consul once every deregister interval

sneat:
Following on from https://github.com/micro/go-micro/pull/313

Making use of the TCPCheck option currently calls `c.Client.Health().Checks` more often than is required, putting unnecessary load on Consul. This change makes it only check if it has been at least `DeregisterTTL` since it last checked.

This is a combination of your original comment (just return nil) and needing to make sure that Consul hasn't deregistered the service due to failing health checks for greater than the `DeregisterTTL` duration

 I'd additionally like to know your thoughts on allowing (presumably via an option/ENV var)

```go
&consul.QueryOptions{
    AllowStale: true,
}
```

so that requests can (optionally) be serviced by non-leader Consul servers, effectively reducing load on the Consul leader (with the tradeoff being potentially out of date data based on the synchronisation time between Consul servers).

### DIFF
--- a/registry/consul_registry.go
+++ b/registry/consul_registry.go
@@ -24,6 +24,8 @@ type consulRegistry struct {
 
 	sync.Mutex
 	register map[string]uint64
+	// lastChecked tracks when a node was last checked as existing in Consul
+	lastChecked map[string]time.Time
 }
 
 func getDeregisterTTL(t time.Duration) time.Duration {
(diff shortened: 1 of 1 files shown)


asim:
LGTM. We should probably add an option to disable reading stale nodes here https://github.com/micro/go-micro/blob/master/registry/consul/options.go with the options but otherwise I see that consul defaults this to true for dns so we should do the same.

PR #335, opened 2018-11-23
Add ability to specify Consul options

sneat:
Add two new Options to allow turning off `AllowStale` (defaults to on) and allows specifying the entire `QueryOptions` struct too for granular control.

```golang
reg := registry.NewRegistry(consulOptions.AllowStale(false))
```

```golang
reg := registry.NewRegistry(consulOptions.QueryOptions(&consul.QueryOptions{
    RequireConsistent: true,
}))
```

I've put together a [second branch](https://github.com/sneat/go-micro/blob/consul-options-test/registry/consul_registry_test.go#L208-L298) that I don't recommend merging to show a couple of tests. Unfortunately import cycles make standard tests difficult without a large refactor which I wasn't willing to do without your go-ahead.

Finally, would you consider a refactor to support non-string `context.WithValue` keys as per [https://godoc.org/context#WithValue](https://godoc.org/context#WithValue). As it stands currently with `consul_registry.go` being part of the `registry` package, converting them away from strings would create an import cycle unless the keys were part of a third package. What are your thoughts here?

### DIFF
--- a/registry/consul/options.go
+++ b/registry/consul/options.go
@@ -27,6 +27,40 @@ func Config(c *consul.Config) registry.Option {
 	}
 }
 
+// AllowStale sets whether any Consul server (non-leader) can service
+// a read. This allows for lower latency and higher throughput
+// at the cost of potentially stale data.
+// Works similar to Consul DNS Config option [1].
+// Defaults to true.
+//
+// [1] https://www.consul.io/docs/agent/options.html#allow_stale
+//
+func AllowStale(v bool) registry.Option {
+	return func(o *registry.Options) {
+		if o.Context == nil {
+			o.Context = context.Background()
+		}
+		o.Context = context.WithValue(o.Context, "consul_allow_stale", v)
+	}
+}
+
+// QueryOptions specifies the QueryOptions to be used when calling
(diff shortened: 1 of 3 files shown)


asim:
I'm not sure it's worth the refactor just to try support non strings. To be honest I gave it a couple of tries but the import cycles result because we need to set a default registry and the consul registry has to implement the registry interface. There's no way around that without breaking default behaviour which is a bigger issue requiring a longer discussion. 

What benefit do you seem moving to non-strings? We do this elsewhere because we can.

sneat:
It was purely based on the description of `WithValue` in the docs:

`The provided key must be comparable and should not be of type string or any other built-in type to avoid collisions between packages using context.`

From my understand of it, if someone was using the micro packages and also another package that also used raw strings then they might collide as far as the equality checks go.

I agree that it's not worth refactoring unless someone comes up with a use-case that outweighs it.

asim:
Yea lets revisit if its an issue.

Are you using micro at a company? Can I add to the users page? https://micro.mu/docs/users.html

asim:
Well in that case thanks very much for the sponsorship!

PR #336, opened 2018-11-23
client: add call option to set maximum received message size

### LINKED ISSUE #219: Provide ability to modify GRPC Max Message Size
GRPC has an Option to allow to modify the Max Message Size. It is currently defaulted at 4MB.

```
const (
	defaultClientMaxReceiveMessageSize = 1024 * 1024 * 4
	defaultClientMaxSendMessageSize    = math.MaxInt32
)

...

// WithMaxMsgSize returns a DialOption which sets the maximum message size the client can receive. Deprecated: use WithDefaultCallOptions(MaxCallRecvMsgSize(s)) instead.
func WithMaxMsgSize(s int) DialOption {
	return WithDefaultCallOptions(MaxCallRecvMsgSize(s))
}
```

Go Micro should have the ability to allow the user to pass a new Max Message Size limit (and internally micro would set the new limit using GRPC's Options)

shuLhan:
The MaxRecvMsgSize is an option for gRPC call which allow client to
customize maximum message that their can receive.

The default maximum received is 4 MB.

This resolve #219.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -77,6 +77,9 @@ var (
 	DefaultPoolSize = 1
 	// DefaultPoolTTL sets the connection pool ttl
 	DefaultPoolTTL = time.Minute
+	// DefaultMaxRecvMsgSize maximum message that client can receive
+	// (4 MB).
+	DefaultMaxRecvMsgSize = 1024 * 1024 * 4
 )
 
 // Makes a synchronous call to a service using the default client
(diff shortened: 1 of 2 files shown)


asim:
This is a grpc option and non go-micro specific. Please add to the grpc plugins, not here.

shuLhan:
The thing is "NewClient" on "go-plugins/client/grpc" use "go-micro/client.Options" as parameter. [1] How can I set the option if the struct itself does not contains the required field?  Can you point me the right line where to add it?

[1] https://github.com/micro/go-plugins/blob/master/client/grpc/grpc.go#L507

PR #347, opened 2018-12-15
Fix cache-selector concurrency problem of locked time too long 

jichre:
When i ab-test my program, found the blocked time was so long at selector/cache:get.
So i changed the mutex type to rwmutex, and fetch from registry async.

### DIFF
--- a/selector/cache/cache.go
+++ b/selector/cache/cache.go
@@ -15,9 +15,10 @@ type cacheSelector struct {
 	ttl time.Duration
 
 	// registry cache
-	sync.Mutex
+	sync.RWMutex
 	cache map[string][]*registry.Service
 	ttls  map[string]time.Time
+	isRequesting bool
 
 	watched map[string]bool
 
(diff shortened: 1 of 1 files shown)


asim:
So the cache miss is only on first request right, so thats slow and the rest fast? Is this optimisation necessary?

jichre:
I changed go-rcache the same first, test commad :
ab -n 100000 -c 1000 "my gateway url"
qps will add 5-10%
I used "go tool trace" to analysis find this 
fb6b3a85-3f9b-46dd-9b9f-3f102ebce7a8

asim:
Interesting. I can see the value but the implementation needs work. I think we can likely move to RWMutex without the need for the async call. First cache miss would be a slow call going to the registry but everything else would be fast.

jichre:
OK!
Hope for better

asim:
Please test this https://github.com/micro/go-micro/pull/352

asim:
Merged the fix as release 0.15.1

PR #353, opened 2018-12-18
implement some gossip options

vtolstov:
* fix panic, when context is nil
* provide Address, Advertise, Config options

### DIFF
--- a/registry/gossip/gossip.go
+++ b/registry/gossip/gossip.go
@@ -2,9 +2,12 @@
 package gossip
 
 import (
+	"context"
 	"encoding/json"
 	"io/ioutil"
+	"net"
 	"os"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -12,7 +15,7 @@ import (
 	"github.com/golang/protobuf/proto"
 	"github.com/google/uuid"
 	"github.com/hashicorp/memberlist"
-	"github.com/micro/go-log"
+	log "github.com/micro/go-log"
 	"github.com/micro/go-micro/registry"
 	pb "github.com/micro/go-micro/registry/gossip/proto"
 	"github.com/mitchellh/hashstructure"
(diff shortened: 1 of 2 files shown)


vtolstov:
don't understand, but found this issue:
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0xefa89f]

goroutine 1 [running]:
github.com/unistack-org/go-micro/registry/gossip.configure(0xc0001ba8c0, 0xc0001b4038, 0x1, 0x1, 0xc0001a46f0, 0x7f14eec286c0)
        vendor/github.com/unistack-org/go-micro/registry/gossip/gossip.go:112 +0x30f
github.com/unistack-org/go-micro/registry/gossip.NewRegistry(0xc0001b4038, 0x1, 0x1, 0xc0001aa6a0, 0xc0001a46f0)
        vendor/github.com/unistack-org/go-micro/registry/gossip/gossip.go:580 +0x19f
```

asim:
Context is likely nil. It needs to be preset in options. Otherwise looks good. Except please don't change the password.

asim:
Ok thats fine but lets call it something useful like `micro-gossip-key`

asim:
My advice would be to create the context when the options are created in NewRegistry. Likely fixes the panic.

vtolstov:
Ok im change it, im eechek context issue and dont see that it nil.

ср, 19 дек. 2018 г., 6:33 Asim Aslam notifications@github.com:

asim:
Well something is invalid and throwing a nil pointer so something is by definition nil.

asim:
`	if optConfig, ok := g.options.Context.Value(contextConfig{}).(*memberlist.Config); ok && optConfig != nil {
`

This is the line, the only possibility here is either context is nil or the value itself is nil. You should not cast before it's actually checked for availability.

Issue #358, opened 2018-12-28
add go.mod and go.sum to support go modules function


caofangkun:
Go Modules:  https://github.com/golang/go/wiki/Modules

![image](https://user-images.githubusercontent.com/1931407/50500835-d6fc3f00-0a8e-11e9-8fbe-92ef7441fc5f.png)

caofangkun:
```
→  go build     
# github.com/micro/go-micro/registry/gossip/proto
registry/gossip/proto/gossip.pb.go:21:11: undefined: proto.ProtoPackageIsVersion3
```
`go get github.com/golang/protobuf@master`

asim:
We won't support go modules until it's a defacto standard. This is likely after a 1.12 release.

PR #359, opened 2018-12-28
#358 add .gitignore, go.mod and go.sum

caofangkun:
Go Modules:  https://github.com/golang/go/wiki/Modules

image


asim:
We won't support go modules until it's a defacto standard. This is likely after a 1.12 release.

PR #368, opened 2019-01-07
 fix strategy roundrobin bug

yincongcyincong:
roundrobin doesn‘t work, so i fix the bug


asim:
It's not a bug. The current client does not provide stateful load balancing. Patching the selector is the incorrect method. Use this instead https://github.com/micro/go-plugins/tree/master/wrapper/select/roundrobin

PR #387, opened 2019-01-16
optimise for range

fengpf:


asim:
Prefer the explicit definition in case syntax changes in future. Sorry.

PR #389, opened 2019-01-17
Fix concurrent access to config metadata

mmwtsn:
This adds in a patch Pete wrote which copies over the metadata maps to allow concurrent access.


mmwtsn:
Ah, I meant to open this against our fork.

PR #397, opened 2019-01-23
pass to broker method broker options

vtolstov:


asim:
Unfortunately this is a breaking change to the API and will not be accepted. Have a look at how we pass SelectOptions into the client Call. That's likely a more appropriate solution.

PR #399, opened 2019-01-24
add context to SubscriberOptions

vtolstov:

### DIFF
--- a/broker/options.go
+++ b/broker/options.go
@@ -111,3 +111,10 @@ func TLSConfig(t *tls.Config) Option {
 		o.TLSConfig = t
 	}
 }
+
+// SubscribeContext set context
+func SubscribeContext(ctx context.Context) SubscribeOption {
+	return func(o *SubscribeOptions) {
+		o.Context = ctx
+	}
+}
--- a/server/handler.go
+++ b/server/handler.go
@@ -1,13 +1,20 @@
 package server
 
+import "context"
+
+type HandlerOption func(*HandlerOptions)
+
 type HandlerOptions struct {
 	Internal bool
 	Metadata map[string]map[string]string
 }
 
+type SubscriberOption func(*SubscriberOptions)
+
 type SubscriberOptions struct {
 	Queue    string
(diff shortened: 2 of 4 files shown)


broker/options.go
```diff
@@ -111,3 +111,10 @@ func TLSConfig(t *tls.Config) Option {
 		o.TLSConfig = t
 	}
 }
+
+// Context set context
+func Context(ctx context.Context) SubscribeOption {
```
asim: Considering this is for Subscribe it should be SubscribeContext

server/handler.go
```diff
@@ -1,13 +1,30 @@
...
+import "context"
+
 type HandlerOptions struct {
 	Internal bool
 	Metadata map[string]map[string]string
 }
 
+type SubscriberOption func(*SubscriberOptions)
+
+func NewSubscriberOptions(opts ...SubscriberOption) SubscriberOptions {
```
asim: Move to bottom of file

vtolstov:
i'm fix issues

server/handler.go
```diff
@@ -35,9 +34,36 @@ func InternalSubscriber(b bool) SubscriberOption {
...
+	}
+
+	for _, o := range opts {
+		o(&opt)
+	}
+
+	return opt
+}
+
+type SubscriberOptions struct {
```
asim: Types should remain at the top. Only move NewSubscriberOptions.

server/rpc_server.go
```diff
@@ -357,6 +357,9 @@ func (s *rpcServer) Register() error {
 		if queue := sb.Options().Queue; len(queue) > 0 {
 			opts = append(opts, broker.Queue(queue))
 		}
+		if cx := sb.Options().Context; cx != nil {
+		opts = append(opts, broker.Context(cx))
```
asim: Please `go fmt`. Also broker.Context is no longer valid.

server/handler.go
vtolstov: ok

Issue #404, opened 2019-01-29
gossip registry


vtolstov:
I'm experiment with gossip registry and may be found some specific issue. When i'm start first service1 with
specify gossip.Address("172.16.1.254:4223") and on other server service2 with registry.Addrs("172.16.1.254:4223") and gossip.Address("172.16.1.1:0")

i see that members equal 2 on both sides, but the second service does not registered in registry.
But  if i stop service1 and start it with registry.Addrs("172.16.1.254:4223") and gossip.Address("172.16.1.1:xxx") where xxx is port provided by service2 all works fine.
So issue appears only on first service1 when it start first without other members.

vtolstov:
also i don't see broadcast update messages in the first case

vtolstov:
also then the service connected to already running registry service it receives updates from it, but i don't see any updates send to first service

vtolstov:
root case of the issue, because service2 then connect broadcast sync event, and serivce2 receives all data from service1, but service1 don't receive any service info from service2

vtolstov:
and LocalState func for connected service does not have any services data in channel, because registry not created when gossip join happening.

vtolstov:
next investigation:
```
g.queue.QueueBroadcast(&broadcast{
  update: up,
  notify: nil,
})
```
this is not send service data when calls Register in gossip.
I'm check this by providing channel to notify and check when read from it returned.

service1 that start first after boadcast returns, but service2 not.

vtolstov:
@asim , gentle ping

asim:
I do not have time to investigate this right now. Feel free to PR a fix.

vtolstov:
nice, i think that enterprise SLA helps with such cases, can you write in enterprise repo you test system , so that other can understand risks and what you autotest for each commit?

vtolstov:
now i have only one workaround, remove check for join in LocalState and in MergeRemoteState
so after first pull/push service data updated on both sides.But this is very ugly. As i see in all cases broadcast not worked for me.

vtolstov:
I'm write test case for gossip registry. And it works fine, also i'm try to run two micro services with the same registry params and service info not propagated to  to each other. Does it possible that some issue present in micro/server code?

vtolstov:
https://github.com/unistack-org/go-micro/blob/gossip/registry/gossip/gossip_test.go

vtolstov:
[...]

PR #411, opened 2019-02-01
registry: gossip unify registry option passing, optimize

vtolstov:


vtolstov:
related to #404

vtolstov:
now with enabling compression in memberlist each service transfer about 100-150 bytes.

asim:
Nice PR

PR #415, opened 2019-02-10
registry: gossip add Reconnect and Timeout

vtolstov:


vtolstov:
i think that i have bug..

PR #433, opened 2019-03-10
docs(rpc_client.go): timeout in seconds

alexesDev:
Hi. I tried to find docs about timeouts and find this, but its wrong.
Also `micro call` doesn't have a timeout arg =(

### DIFF
--- a/client/rpc_client.go
+++ b/client/rpc_client.go
@@ -75,7 +75,7 @@ func (r *rpcClient) call(ctx context.Context, node *registry.Node, req Request, 
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


asim:
We want the timeout in nanoseconds so if that's not the case we need to fix this. We're printing a digit which means it's going to print duration so I think it should be ok.

alexesDev:
Today I checked it and in practice this timeout in seconds. Ok, no problem.
```bash
curl -H 'Timeout: 360' -d 'service=go.micro.srv...
```

PR #466, opened 2019-05-02
send requestBody to mock function if it can handle it

Ak-Army:
The change is backward compatible


asim:
Can you name it request rather than reqBody? Thanks.

Ak-Army:
Yes, but the function get just the request.Body()

PR #473, opened 2019-05-12
add RegisterCheck server option for internal health checks

vtolstov:

### DIFF
--- a/server/options.go
+++ b/server/options.go
@@ -25,6 +25,8 @@ type Options struct {
 	HdlrWrappers []HandlerWrapper
 	SubWrappers  []SubscriberWrapper
 
+	// RegisterCheck runs a check function before registering the service
+	RegisterCheck func(context.Context) error
 	// The register expiry time
 	RegisterTTL time.Duration
 	// The interval on which to register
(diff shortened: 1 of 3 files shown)


vtolstov:
i'm test it and it works fine for me (deregister forcible when error happened)

server/options.go
```diff
@@ -163,6 +165,13 @@ func Metadata(md map[string]string) Option {
 	}
 }
 
+// RegisterCheck run func before registry service
```
asim, line 172: "RegisterCheck runs a check function before registering the service"

server/rpc_server.go
```diff
@@ -518,8 +527,25 @@ func (s *rpcServer) Start() error {
...
+				// if register check func defined, use it
+				if fn := s.opts.RegisterCheck; fn != nil {
+					if err = fn(s.opts.Context); err != nil {
+						log.Logf("Server %s-%s register error: %s", config.Name, config.Id, err)
+						s.RLock()
+						registered := s.registered
+						s.RUnlock()
+						if registered {
+							// deregister self in case of error
+							if err := s.Deregister(); err != nil {
```
asim: You've not deregistered and not set s.registered to false. This code block is also a bit unclean in a huge if statement. Maybe there should be a default check function that always returns nil so we don't actually have to test the function exists.

vtolstov: i think that i fix this with new commit

vtolstov:
i'm fix all, i think now it less spams in case of repeated errors in check func

server/options.go
```diff
@@ -163,6 +169,13 @@ func Metadata(md map[string]string) Option {
...
```
asim, line 172: Can you make this comment match the above

vtolstov: yes, sorry. fixed

PR #478, opened 2019-05-16
Use existing consul client for watcher

sneat:
# Description

This change utilises the existing Consul Client for connecting the Watcher to Consul rather than creating a new Client during the `wp.Run()` call. The definition of the log entry is taken from the `.Run()`→ `.RunWithConfig()` → `.RunWithClientAndLogger()` flow that exists.

## Why does this matter

In our setup, we run Consul with multiple Datacenters to support multiple environments. We configure the micro Registry to connect to Consul specifying the Datacenter to use which correctly facilitates service discovery. Some of our services need to be able to access services in other environments which we do by maintaining a separate Client connection with the Datacenter specified to use within the service discovery. This all works well until a service deregisters, we never get the watch event firing (by design within Consul due to Datacenters).

## Solution

By making use of the `consul.Client` available on the `consulRegistry` we can piggyback on any configuration already applied to the Client, including Datacenter specifications.

### DIFF
--- a/registry/consul/watcher.go
+++ b/registry/consul/watcher.go
@@ -2,6 +2,8 @@ package consul
 
 import (
 	"errors"
+	"log"
+	"os"
 	"sync"
 
 	"github.com/hashicorp/consul/api"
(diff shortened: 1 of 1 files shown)


asim:
Thanks

PR #484, opened 2019-05-24
add AutoAck support for Server

printfcoder:


BaileyJM02:
I don't know whether this is the cause but when running: `go get github.com/micro/go-grpc`, I'm getting this error:
```
# github.com/micro/go-plugins/server/grpc
../../../micro/go-plugins/server/grpc/grpc.go:532:19: sb.Options().AutoAck undefined (type server.SubscriberOptions has no field or method AutoAck)
../../../micro/go-plugins/server/grpc/subscriber.go:39:10: unknown field 'AutoAck' in struct literal of type server.SubscriberOptions
```

asim:
go get -u probably fixes that for you

PR #485, opened 2019-05-24
Add grpc transport

asim:
We are not yet merging this!

We want to add grpc as a default transport but its unclear in what form. People mostly want it for transparent proxying but I'm not really sure we can support that. Go Micro and grpc-go are competing frameworks. The interop may never be clean so we need to identify how we're moving forward.

This PR adds the grpc transport from go-plugins. We define a client/server which takes a transport message. This has good performance characteristics but doesn't really deal with the proxy case people want.

Open to discussion.


starsea:
When is code merged?

asim:
Not for a while, we're assessing the best way to support grpc

PR #489, opened 2019-05-29
Add nats broker as a default

asim:


vtolstov:
nice to see this merged, with this by defaults micro have more mature defaults to local and production development (gossip, nats...)
i'm thinking about some discovery mechanism with priority like mdns,dns-srv,udp broadcast (fallback if multicast is blocked).

asim:
What I wonder is whether it's wrong to play favourites with adding more defaults. Maybe there should only be one default of each and every thing else should live in go-plugins. 

It feels like go-micro should not be the center of gravity. That something higher level should construct the various libraries together and add the plugins.

PR #496, opened 2019-06-05
fix grpc import package error

cbwfree:
fix grpc import package error


asim:
Nothing in go-micro should ever reference go-plugins

PR #500, opened 2019-06-06
cleanup go.mod remove replace section

leaxoy:
This PR cleanup go.mod file, remove `replace` and `exclude` section, and work well. This will solve deps hell, I think.


asim:
Are these manual edits to the modules?

leaxoy:
No， I use `go mod tidy` maintain this file, while go version is 1.12.5

leaxoy:
In fact, first remove `replace` and `exclude`, then run `go mod tidy`

asim:
And if you run go get -u ./...?

leaxoy:
Only `go mod tidy` been run, I try run `go get -u ./...` on my local pc, it success without `replace`, but add new deps to `go.mod`, should I add those deps to this PR?

vtolstov:
Looks good

asim:
Why is it importing grpc-gateway and related things now? There are deps here that seem totally unrelated.

vtolstov:
this is indirect dep, some of used dep uses github.com/grpc-ecosystem/grpc-gateway

vtolstov:
i'm do myself go mod tidy so i'm compare my results with this pr.
i think that this is etcd dep (grpc-gateway)

leaxoy:
Sorry for the delay due to time difference， like @vtolstov  says，  grpc-gateway is an indirect dep， it should be imported by other dep.

asim:
Strangely if I remove the file and do it now I don't even get that dependency

vtolstov:
Strange

пт, 7 июн. 2019 г., 10:05 Asim Aslam <notifications@github.com>:

PR #509, opened 2019-06-11
fix etcd error

outshow:
1. fix etcd dep error [https://github.com/micro/go-config/issues/105](url), this is another solution. Please refer to [https://github.com/etcd-io/etcd/issues/10398#issuecomment-469439257](url).
2. add dialTimeout option for etcd client, program will hang there if the endpoints are not reachable without setting this option.

### DIFF
--- a/config/source/etcd/etcd.go
+++ b/config/source/etcd/etcd.go
@@ -6,9 +6,9 @@ import (
 	"net"
 	"time"
 
+	cetcd "github.com/coreos/etcd/clientv3"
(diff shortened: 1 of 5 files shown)


vtolstov:
Looks reasonable

asim:
We need to move this plugin out to go-plugins. We're trying to remove etcd usage inside go-micro itself for now since we're not personally using it at the company and without that use we can't ultimately support. I'll merge but look to see the entire thing move to go-plugins.

vtolstov:
Yes, i think that moving to go-plugins is better now.

PR #510, opened 2019-06-11
[WIP] add multi registry

vtolstov:


vtolstov:
This allows combine multiple registries. Main idea to have readonly registry and read/write.
So when you call register it registries only in read/write registries.

asim:
As discussed in slack. Multi-X should be in go-plugins.

PR #514, opened 2019-06-12
remove mock data from memory registry

vtolstov:
memory registry can be used as fast inprocess registry,
so mock data needs to be in tests only

### DIFF
--- a/broker/common_test.go
+++ b/broker/common_test.go
@@ -1,0 +1,51 @@
+package broker
+
+import (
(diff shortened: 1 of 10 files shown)


asim:
I have a feeling that breaks some things that rely on that for tests

vtolstov:
yes, i'm try to find all stuff that uses it

vtolstov:
all ok now

PR #517, opened 2019-06-14
fix wrong path of nats.go

printfcoder:
seems nats-io already mved it ago
err:

```
stat github.com/nats-io/nats.go: no such file or directory
```


vtolstov:
This is wrong.see readme in nats.go repo you miss slash

PR #521, opened 2019-06-17
New registry interface

asim:
DO NOT MERGE

This moves Result to Event and adds a SendEvent method to the registry. We also add a channel for events, switch Action to Type, add Interval and a Message type.


vtolstov:
Can you switch event type from string to dedicated type , and provide types via const and create String() method for it type?

asim:
We need to think about this. Because every additional event would require modifying the registry to add it. What's your reason for this?

vtolstov:
because if i want to check event type i need to know all strings of events, but in case of "enum" i'm always have all types. Also if i want to store this data or transfer via wire, int types is preferred.
Also this is idiomatic go representation of something that must have concrete value.

vtolstov:
also i don't think that something more like CRUD provides needed for event type.

vtolstov:
Also best to do - NewRegistry returns (Registry, error), and not Registry, but this is needs to modify some other pieces of code too

asim:
That's more related to updating all initialisation related to the cmd package

vtolstov:
@asim, yes. but this is good, we can avoid log.Fatal/log.Fatalf in all registry code

asim:
Ok. Point noted about event type. If we can change initialisation to add error it needs to be everywhere. That's a separate and more painful change.

vtolstov:
yes, i think so (about changing NewRegistry, also this needs to modify all go-plugins registries too)

asim:
Tbh we should be moving cmd into config/cmd and then be changing how plugins are registered

vtolstov:
@asim why?

asim:
I'll PR when I have a complete solution.

PR #528, opened 2019-06-19
Adds router package

milosgajdos:
This PR introduces `router` package which allows to implement micro network router. 

It provides two basic primitives:
* `Table` - routing table which allows to CRUD route operations
* `Router` -  maintains a routing table and advertises local services to "network" registry

The package provides basic e.g. default implementation of both `Router` and `Table` objects.


router/default_router.go
```diff
@@ -0,0 +1,322 @@
...
+	"time"
+
+	"github.com/micro/go-micro/registry"
+	"github.com/olekukonko/tablewriter"
+)
+
+var (
+	// AdvertiseToNetworkTick defines how often in seconds do we scal the local registry
+	// to advertise the local services to the network registry
+	AdvertiseToNetworkTick = 5 * time.Second
```
asim: AdvertiseTick

router/default_router.go
```diff
@@ -0,0 +1,322 @@
...
+	// AdvertiseNetworkTTL defines network registry TTL in seconds
+	// NOTE: this is a rather arbitrary picked value subject to change
+	AdvertiseNetworkTTL = 120 * time.Second
```
asim: AdvertiseTTL

router/options.go
```diff
@@ -0,0 +1,105 @@
...
+	NetworkAddress string
+	// LocalRegistry is router local registry
+	LocalRegistry registry.Registry
+	// NetworkRegistry is router remote registry
+	// NOTE: we need some abstraction on top of gossip.Registry
+	NetworkRegistry registry.Registry
+	// Table is routing table
+	Table Table
+	// RIB is Routing Information Base
+	RIB RIB
```
asim: Remove RIB as currently unused

router/options.go
```diff
@@ -0,0 +1,105 @@
...
+	// DefaultNetworkAddress is default network bind address
+	DefaultNetworkAddress = ":9094"
+)
+
+// Options are router options
+type Options struct {
+	// ID is router id
+	ID string
+	// Address is router micro service address
+	Address string
```
asim, line 20: Router has no address because it's not a service, its just a library

router/options.go
```diff
@@ -0,0 +1,105 @@
...
+	// GossipAddress is router gossip address
+	GossipAddress string
```
asim: Change gossip address to Address as no one knows its a gossip registry and the assumption is just this is the router's address to talk to other routers. We just happen to use gossip for it.

router/options.go
```diff
@@ -0,0 +1,105 @@
...
+	// NetworkAddress is micro network address
+	NetworkAddress string
```
asim: Change NetworkAddress to Advertise. This is the address which is advertised.

router/options.go
```diff
@@ -0,0 +1,105 @@
...
+	// LocalRegistry is router local registry
+	LocalRegistry registry.Registry
```
asim: LocalRegistry to Registry, its assumed to be the local registry

router/options.go
```diff
@@ -0,0 +1,105 @@
...
+	// NetworkRegistry is router remote registry
+	// NOTE: we need some abstraction on top of gossip.Registry
+	NetworkRegistry registry.Registry
```
asim: NetworkRegistry changed to Network.

router/query.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+}
+
+// QueryOption sets routing table query options
+type QueryOption func(*QueryOptions)
+
+// QueryOptions are routing table query options
+type QueryOptions struct {
+	// DestAddr is destination address
+	DestAddr string
```
asim: DestAddr to Destination

router/rib.go
```diff
@@ -0,0 +1,30 @@
+package router
+
+// RIB is Routing Information Base.
+// RIB is used to source the base routing table.
+type RIB interface {
```
asim: RIB is being removed

router/route.go
```diff
@@ -0,0 +1,140 @@
...
+
+var (
+	// DefaultLocalMetric is default route cost for local network
+	DefaultLocalMetric = 1
+	// DefaultNetworkMetric is default route cost for micro network
+	DefaultNetworkMetric = 10
+)
+
+// AddPolicy defines routing table addition policy
+type AddPolicy int
```
asim: Remove AddPolicy. Or change the Policy to be a function that takes a Route and allows the caller to define how to handle it e.g like Retry and Backoff Func in client https://github.com/micro/go-micro/blob/master/client/options.go#L51 then provide a default like AlwaysOverride

router/table_watcher.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// WatchOption is used to define what routes to watch in the table
+type WatchOption func(*WatchOptions)
+
+// Watcher defines routing table watcher interface
+// Watcher returns updates to the routing table
+type Watcher interface {
+	// Next is a blocking call that returns watch result
+	Next() (*Event, error)
```
asim, line 56: Also add `Chan() (<-chan *Event, error)`. Been bitten by lack of Chan in registry watcher and we're going to add one.

router/table_watcher.go
```diff
@@ -0,0 +1,140 @@
...
+type WatchOptions struct {
+	// Specify destination address to watch
+	DestAddr string
+	// Specify network to watch
+	Network string
+}
+
+// WatchDestAddr sets what destination to watch
+// Destination is usually microservice name
+func WatchDestAddr(a string) WatchOption {
```
asim: WatchDestination

router/table_watcher.go
```diff
@@ -0,0 +1,140 @@
...
+	// Next is a blocking call that returns watch result
+	Next() (*Event, error)
+	// Stop stops watcher
+	Stop()
+}
+
+// WatchOptions are table watcher options
+type WatchOptions struct {
+	// Specify destination address to watch
+	DestAddr string
```
asim: Destination

router/query.go
```diff
@@ -0,0 +1,104 @@
...
+	// NetworkAddress is network address
+	Network string
+	// Gateway is gateway address
+	Gateway Router
```
asim: Change to Router

router/options.go
milosgajdos: We've agreed to "move" the `GossipAddress` into this field i.e. `Address` is the router's "gossip address"

asim: Agreed

PR #534, opened 2019-06-20
consul config source support 1st-level array

magodo:
Check whether the 1st level encoded json is array or not, to
support 1st level array in consul config.

During debug, i suspected the incapability of arrray is caused by
json reader, so i added test for array. I think it makes no harm
to also check that in.


magodo:
WIP, don't merge..

asim:
Please PR went complete and can be merged.

Issue #532, opened 2019-06-20
Can a go-micro stream be like a grpc stream?


lockp111:
go-micro stream must send then recv,but grpc not.
like this:
```go
go func() {
        for {
	    select {
		case <-ctx.Done():
		        log.Println("close...")
			return
		default:
			log.Println("send...")
			stream.Send(&steam.Request{
				Msg: time.Now().String(),
			})
		}
	}
}()

go func() {
	for {
		req, err := stream.Recv()
		if err != nil {
                         stream.Close()
			log.Println("err:", err)
                         cancel()
		}
		log.Println("recv:", req.Data)
	}
}()
```

lpxxn:
yes go-micro can do this. 
[example](https://github.com/micro/examples/tree/master/stream)

lockp111:
> yes go-micro can do this.
> [example](https://github.com/micro/examples/tree/master/stream)

Sorry,I don't think so.
I think you have never read the code.

lpxxn:
???
Go-micro bidirectional stream processing steps is that
client-->connect to server
server or client -->  rev or send data 

the server side stream  function finished and  return ,  go-micro will send  EOF to client ,then the client close the stream;

lockp111:
> ???
> Go-micro bidirectional stream processing steps is that
> client-->connect to server
> server or client --> rev or send data
> 
> the server side stream function finished and return , go-micro will send EOF to client ,then the client close the stream;

Do you know Recv() and Send() have lock?
**Please read the source code.**
If you receiving but nothing then you can not send anything.
That is the reason why I open this issue.

lpxxn:
sorry i Misunderstanding it. i have read the source code.  the default stream Rev() and Send() have lock.
perhaps you can use the [grpc plugin](https://github.com/micro/go-plugins/tree/master/server/grpc) replace the default processing
it has no lock.  [grpc client](https://github.com/micro/go-plugins/blob/master/client/grpc/stream.go) and [grpc server](https://github.com/micro/go-plugins/blob/master/server/grpc/stream.go)

asim:
As I see it the question is stating that that the rpc_stream.go code has a mutex around sending and receiving data. When Send is called, it locks the mutex. When Recv is called, it locks the mutex. There for only one or the other operation can be honoured at once. This essentially limits bidirectional streaming to the ability to process requests over the connection. Such that if Recv is blocking waiting for a response from the server nothing can be sent to the server. My assumption is the server code probably looks similar.

[...]

PR #535, opened 2019-06-21
remove: default rpc stream mutex lock

lockp111:
#532


vtolstov:
do you check changes via race flag?

asim:
Please don't PR commented code. Also please do not PR something that's clearly untested. Removing mutexes can introduce significant race conditions.

PR #536, opened 2019-06-21
Config source consul support 1st array

magodo:
Check whether the 1st level encoded json is array or not, to
support 1st level array in consul config.

During debug, i suspected the incapability of arrray is caused by
json reader, so i added test for array. I think it makes no harm
to also check that in.


asim:
The conflicts for this need to be resolved.

magodo:
Conflict resolved.

PR #538, opened 2019-06-21
config consul source supports slash as prefix

magodo:
`config.NewConfig()` with consul source will both read from consul
and watch consul for changes. Hence, the `prefix` is used in these
2 cases:

- read case: it is used to strip path based on the `KVPair` returned
from consul `kv.List()` method
- watch case: it is used as the `key` of watch query (`keyprefix` type)

So for *watch case*, the `key` is leagal to be `/` for watching change
on root. While for *read case*, because `KVPair.Key` is always stripped
off the leading slash, so if user specified some `prefix` with leading
slash, we should strip it also.

An extream case would be: user want's to read & watch node in root dir.
One would specify `prefix` as `/`, and it should work then.

### DIFF
--- a/config/source/consul/README.md
+++ b/config/source/consul/README.md
@@ -4,7 +4,7 @@ The consul source reads config from consul key/values
 
 ## Consul Format
 
-The consul source expects keys under the default prefix `micro/config`
+The consul source expects keys under the default prefix `/micro/config`
 
 Values are expected to be json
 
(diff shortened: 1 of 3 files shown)


magodo:
There is expected conflict against #536, which i will resolve once you merge one of them.

asim:
So are you saying your previous PR to strip prefix broke this functionality?

magodo:
no, that pr did nothing more than modify the default ptefix. so the issue always exist.
with this pr, i can just roll back last pr, which is not necessary any more.

PR #539, opened 2019-06-21
[WIP] mdns registry cleanup

vtolstov:
* avoid allocation new mdns.Server for each go-micro service
* serve records directly from services map
* extend for feature watcher async updates


vtolstov:
i think that i'm lost ability to list all services, rename issue to WIP before it fixed.

asim:
Please resubmit a functioning PR

PR #542, opened 2019-06-25
Add consul-specific option for config (as registry)

magodo:
Consul default `http.Client` use `Timeout` as 0, which might cause hang issue if server outage occur, as indicated by [this blog](https://medium.com/@nate510/don-t-use-go-s-default-http-client-4804cb19f779).

User should have the ability to do finer grained config when using consul as config backend.

### DIFF
--- a/config/source/consul/options.go
+++ b/config/source/consul/options.go
@@ -12,7 +12,7 @@ type prefixKey struct{}
 type stripPrefixKey struct{}
 type dcKey struct{}
 type tokenKey struct{}
-type consulConfigKey struct{}
+type configKey struct{}
 
 // WithAddress sets the consul address
 func WithAddress(a string) source.Option {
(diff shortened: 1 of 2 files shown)


config/source/consul/options.go
```diff
@@ -61,3 +63,13 @@ func WithToken(p string) source.Option {
 		o.Context = context.WithValue(o.Context, tokenKey{}, p)
 	}
 }
+
+// WithConsulConfig set consul-specific options
+func WithConsulConfig(c *api.Config) source.Option {
```
asim: Please specify as WithConfig considering this is already in the consul package

config/source/consul/options.go
```diff
@@ -11,6 +12,7 @@ type prefixKey struct{}
 type stripPrefixKey struct{}
 type dcKey struct{}
 type tokenKey struct{}
+type consulConfigKey struct{}
```
asim: configKey

PR #553, opened 2019-06-27
Changed router interface. Added table watcher. Advertise routes

milosgajdos:
**DO NOT MERGE IN YET**

* Changed router interface to return Advertisement channel
* Added default gateway route to the routing table if supplied
* Watch table for updates and advertise to the network
* We hash the routes on 3-tuple (Destination, Gateway, Network)

### DIFF
--- a/network/router/default_router.go
+++ b/network/router/default_router.go
@@ -32,7 +32,7 @@ func newRouter(opts ...Option) Router {
 
 	return &router{
 		opts:       options,
-		status:     Status{Error: nil, Code: Stopped},
+		status:     Status{Error: nil, Code: Init},
 		advertChan: make(chan *Update),
 		exit:       make(chan struct{}),
 		wg:         &sync.WaitGroup{},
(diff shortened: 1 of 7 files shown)


network/router/default_router.go
```diff
@@ -69,43 +72,110 @@ func (r *router) Network() string {
...
+			errChan <- r.watchTable(tableWatcher)
+		}()
+
+		r.wg.Add(1)
+		go func() {
+			defer r.wg.Done()
+			select {
+			// wait for exit chan
+			case <-r.exit:
+				r.status.Code = Stopped
```
asim: You should lock when updating the status. There is a lock call at top of the function but this function exits after all the go routines are fired.

network/router/default_router.go
```diff
@@ -69,43 +72,110 @@ func (r *router) Network() string {
...
+			case err := <-errChan:
+				r.status.Code = Error
+				r.status.Error = err
+			}
+			// close the advertise channel
+			close(r.advertChan)
```
asim: If you close this here, if I call Advertise again is it creating a new advertChan?

network/router/default_router.go
milosgajdos: Yep, I know. It shall be fixed shortly.

network/router/default_router.go
milosgajdos: This question will hopefully be answered as "yes" in the next commit I'm working on.

network/router/default_router.go
```diff
@@ -190,13 +167,186 @@ func (r *router) manageServiceRoutes(w registry.Watcher, metric int) error {
...
+		u := &Update{
+			ID:        r.ID(),
+			Timestamp: time.Now(),
+			Event:     event,
+		}
+
+		select {
+		case <-r.exit:
+			return nil
+		case r.advertChan <- u:
```
milosgajdos, line 201: `advertChan` is a blocking channel; so this will block until update `u` is read by something

Furthermore what happens when we close the `advertChan`? this will most likely `panic` as writing to closed channel causes `panic`

milosgajdos (commented):
This still needs proper thinking, so **DO NOT MERGE THIS IN**

PR #563, opened 2019-07-03
export registry util function to safe copy registry data

vtolstov:
this functions is duplicated in many registries


asim:
I get the need to reuse it but it's really ugly under registry package as util. It should either be in cache or memory as exported methods.

vtolstov:
why in cache? for example gossip and mdns and memory registry need this functions (may be other registries in plugins too)

vtolstov:
firstly i think about export this function under top level registry package. what you say?

vtolstov:
i'm minimize count of exported func, only Del,Add and Copy.
also fix the issue when adding multiple versions of the same service

vtolstov:
@asim can you merge it?

asim:
I need to review it first

vtolstov:
no problem

PR #565, opened 2019-07-04
grpc: if unmarshal target is proto.Message, using jsonpb

sunfuze:
Make grpc json codec unmarshal behavior like codec/json, support json unmarshal some special protobuf type like google.protobuf.Struct

### DIFF
--- a/client/grpc/codec.go
+++ b/client/grpc/codec.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


vtolstov:
can you rewrite pr to utilize https://golang.org/pkg/bytes/#NewReader ? with string(data) you allocate additional memory to hold data.

asim:
Going to merge as is. Change for more memory efficient version can be done after.

PR #570, opened 2019-07-08
grpc: using jsonpb.Marshaler to do Marshal, map to jsonpb.Unmarsh

sunfuze:
Correspond to grpc jsonCodec.Unmarshal, using jsonpb.Marshaler to do Marshal.


asim:
Please look at the server side and use the same method.

sunfuze:
> Please look at the server side and use the same method.

Server side use jsonpb.Unmarshal, client use jsonpb.Marshaler to Marshal? This is not right?

PR #584, opened 2019-07-12
transport memory: try to find free address 100 times

vtolstov:
on busy system with many listeners running i get error
like "already listening on []:16264"
add additional try to find free address


asim:
Doesn't adding 100 tries seem a bit absurd?

vtolstov:
Yes, 100 times is ugly.. but what is the real fix?
Increase randomisation interval?

vtolstov:
drop in favor of #585

Issue #586, opened 2019-07-15
the registry cache performance problem


webliupeng:
Hi ,I use latest version go-micro with consul registry. I got a performance problem when high traffic(about 5000 request per second and 50 nodes ). 

I found these code:
```golang
func (c *cache) get(service string) ([]*registry.Service, error) {
	// read lock
	c.RLock()

	// check the cache first
	services := c.cache[service]
	// get cache ttl
	ttl := c.ttls[service]

	// got services && within ttl so return cache
	if c.isValid(services, ttl) {
		// make a copy
		cp := registry.Copy(services)
		// unlock the read
		c.RUnlock()
		// return servics
		return cp, nil
	}
```
it's looks like copy services cause alloc a lot of memory. I have use `go tool ` to analyze alloc space got these result:

```
    0.51GB  0.36%  0.49%    54.96GB 39.09%  github.com/micro/go-micro/client.(*rpcClient).Call
         0     0%  0.49%       54GB 38.41%  github.com/micro/go-micro/client.(*rpcClient).next
    0.06GB 0.043%  0.54%       54GB 38.41%  github.com/micro/go-micro/selector.(*registrySelector).Select
         0     0%  0.54%    53.32GB 37.92%  github.com/micro/go-micro/registry/cache.(*cache).GetService
         0     0%  0.54%    53.32GB 37.92%  github.com/micro/go-micro/registry/cache.(*cache).get
         0     0%  0.54%    48.44GB 34.45%  github.com/micro/go-micro/registry/cache.(*cache).get.func1
    0.01GB 0.009%  0.55%    48.43GB 34.44%  github.com/micro/go-micro/registry/consul.(*consulRegistry).GetService
```

My question is these is a way to reduce memory allocation?

milosgajdos:
Thanks for report this @webliupeng 

I've opened #587 to address some of the performance issues in `registry.Copy`. 

Going forward we will have to revise the code more, but for the time being, the PR should at the very least address the reallocations related to `append()` which reallocates the whole slice as soon as it hits its capacity, which in high volume microservices projects is very likely to happen.

vtolstov:
I think main problem that we copies data twice. And have 3 copies one in registry one in cache and one in client selector.

milosgajdos:
We need to think this through and come up with a proper fix.

Like I said #587 is a partial solution, not a complete one. It will alleviate *some* number of allocations, but you've raised a great point @vtolstov. Thanks!

[...]

PR #587, opened 2019-07-15
Preallocate slices in registry.Copy() to avoid append() reallocations when copying data

milosgajdos:
[Partially] addresses #586

### DIFF
--- a/registry/util.go
+++ b/registry/util.go
@@ -51,33 +51,32 @@ func delNodes(old, del []*Node) []*Node {
 
 // Copy makes a copy of services
 func Copy(current []*Service) []*Service {
-	var services []*Service
-
-	for _, service := range current {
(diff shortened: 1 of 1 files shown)


vtolstov (approved):
looks good, changes are trivial but very useful.

vtolstov:
@milosgajdos83 can you change also https://github.com/micro/go-micro/blob/master/registry/util.go#L7 ?

milosgajdos:
Yep, happy to @vtolstov.  

I'm thinking...we might as well make the same prealloc change across the whole `util.go` file. The initial PR fix was a quick early-morning 😴fix addressing only the `Copy()` function.

vtolstov (approved):

vtolstov:
thanks, @asim can you merge it?

PR #591, opened 2019-07-16
[WIP] Fixes advert route event dampening behaviour

milosgajdos:
**DO NOT MERGE THIS YET**

This PR adds the following changes:
* `routeAdvert` now stores a slice of route events as opposed to just the last one
* we attempt to deduplicate route events when a new event is received before appending it to the `routeAdvert` events
* we now have a max suppress time threshold for the long time suppressed adverts
* we now correctly decay event penalties on every advert tick

Originally we werent decaying event penalties on every advert tick. Not decaying the event penalties on every tick or not having the max suppression threshold could potentially lead to DoS by growing the router memory ending in out of memory crash.

Furthermore, some events ended up being accumulated in advert events for a long time equally 
causing memory leaks. By introducing `MaxSuppressTime` we now delete these long held suppressed events. This might be subject to further change, though.

Previously we were also overriding the last received route event which was causing incorrect sequence of events to be applied when received by a receiver. Think of the following sequence:
`Create`, `Delete` would be "squashed" into `Delete` **only** which would be nonsensical in its own right since the `Create` event would never get advertised, hence we would end up trying to delete nonexistent routes.

Finally, some changes have been made to the names of some **unexported** functions and some variables.

The route dampening algorithm has been inspired by [1]

[1] [BGP Route Flap Dampening Algorithms](http://www.ensc.sfu.ca/~ljilja/cnl/pdf/steve_thesis.pdf)

### DIFF
--- a/network/router/default.go
+++ b/network/router/default.go
@@ -273,10 +273,8 @@ func (r *router) advertiseTable() error {
 
 			// advertise all routes as Update events to subscribers
 			if len(events) > 0 {
-				go func() {
-					r.advertWg.Add(1)
-					r.advertiseEvents(Update, events)
-				}()
+				r.advertWg.Add(1)
+				go r.advertiseEvents(Update, events)
 			}
 		case <-r.exit:
 			return nil
--- a/network/router/default.go
+++ b/network/router/default.go
@@ -18,17 +18,24 @@ const (
 	// AdvertiseTableTick is time interval in which router advertises all routes found in routing table
 	AdvertiseTableTick = 1 * time.Minute
 	// AdvertSuppress is advert suppression threshold
-	AdvertSuppress = 2000
+	AdvertSuppress = 2000.0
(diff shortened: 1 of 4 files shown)


network/router/default.go
```diff
@@ -594,5 +609,5 @@ func (r *router) Stop() error {
 
 // String prints debugging information about router
 func (r *router) String() string {
-	return "router"
+	return "default router"
```
asim: This should either be called default or router. We only ever return single word strings for implementations to be stored in metadata.

milosgajdos: Yah, I hear ya, Maxi. This one escaped my attention somehow! I had a few routers running off different branches and this one got away. Nice spot!

milosgajdos: Come to think of it, I think we should return `"default"` and not `"router"` as we originally did. WDYT?

asim: Sounds good

PR #594, opened 2019-07-17
fix ipv6 addr parsing

vtolstov:


leeson1995:
lgtm

vtolstov:
@leeson1995 can you test this pr in your env?

vtolstov:
@asim i think that this is ready to merge. I can't test it with pure ipv6 addresses, but with ipv6+ipv4 all works fine.

PR #595, opened 2019-07-17
[WIP] DONT MERGE now transport: memory fix race on channel close

vtolstov:

### DIFF
--- a/transport/memory/memory.go
+++ b/transport/memory/memory.go
@@ -20,6 +20,7 @@ type memorySocket struct {
 	// listener exit
 	lexit chan bool
 
+	once   sync.Once
 	local  string
 	remote string
 	sync.RWMutex
@@ -80,14 +81,18 @@ func (ms *memorySocket) Send(m *transport.Message) error {
 }
 
 func (ms *memorySocket) Close() error {
-	ms.Lock()
-	defer ms.Unlock()
-	select {
-	case <-ms.exit:
-		return nil
-	default:
-		close(ms.exit)
+	onceBody := func() {
+		ms.Lock()
+		defer ms.Unlock()
+		select {
+		case <-ms.exit:
+			return
+		default:
+			close(ms.exit)
+		}
 	}
+	ms.once.Do(onceBody)
+
 	return nil
 }
 


vtolstov:
@asim i'm check this by running million of requests with more then 5 minutes.
tests complited, but i have error
race: limit on 8128 simultaneously alive goroutines is exceeded, dying
that not related to this pr

vtolstov:
hm, on next run i hit the same issue

milosgajdos:
I was thinking about this last night. The original code causing the double `close` seems wrong to start with. From channel semantics in [Go language specification](https://golang.org/ref/spec)

> If one or more of the communications can proceed, a single one that can proceed is chosen via a uniform pseudo-random selection. Otherwise, if there is a default case, that case is chosen. If there is no default case, the "select" statement blocks until at least one of the communications can proceed.

Btw, @vtolstov, about `race: limit on 8128 simultaneously alive goroutines is exceeded, dying`. 

Are you getting this error when running `go build -race`? I've found this  [go cl](https://go-review.googlesource.com/c/go/+/103160/) raised in [here](https://github.com/golang/go/issues/23611). There seems to be a goroutine cap when running `-race` detector 🤷‍♂️

vtolstov:
yes , memory transport issue brings by invalid use of channel to close.
about race - yes my code build with race flag and i'm stress test it. i think that somewhere present goroutine leak.

asim:
Why is this using sync.Once?

vtolstov:
This is experiment. With sync.Once problem appears rarely.

asim:
Should locking not be enough?

vtolstov:
Locking not helps. Panic happened often

asim:
Do you have the line it panics on because technically this will not panic when locked unless something else is sending on the channel or closing elsewhere.

vtolstov:
Panic happened when try to close closed channel. In Close() func

сб, 20 июл. 2019 г., 22:11 Asim Aslam <notifications@github.com>:

asim:
If you are locking then by definition access is serialised. Once you call close the channel is effectively closed. Once the mutex unlocks the next thread can proceed and it should not block and basically return without attempting to close. Please provide a trace of the panic and ensure you are using the most up to date code. 

I cannot otherwise see why it would panic with a lock unless you're running old code.

vtolstov:
why next goroutine don't close?
https://github.com/micro/go-micro/blob/master/transport/memory/memory.go#L87

why it can't go to default case in select?
also as i see https://golang.org/ref/spec#Close
we can use multivalue select case to check that channel closed
what you think?

milosgajdos:
Guys, this has absolutely nothing to do with locking.

This has to do with Close() method NOT being thread safe (it can’t be safely called by multiple threads/goroutines) not in the way it is written right now.

As I mentioned in my first comment, Go spec explains the behaviour pretty well, but let’s walk through this:

> If one or more of the communications can proceed, a single one that can proceed is chosen via a uniform pseudo-random selection

1. First goroutine which successfully calls Close() will close the channel

Now the next goroutine which calls Close() AFTER the channel has been closed previously has two possible execution paths which can proceed:
- exit clause
- default clause

Go spec says: if there are more than one paths possible the one that can proceed is chosen pseudo randomly — it can be either the exit or default clause: **Go does not guarantee which one**, hence when Close() is called multiple times (by multiple goeoutines) the race is inevitable.

Now, I do not have a solution to how the Close() should be written to make it thread safe, unfortunately, but would like to give it a proper thought next week

vtolstov:
i think that we must have only one case clause with multivalue select if value received and ok - close channel, if value received and !ok - don't close channel but only return

milosgajdos:
Yeah, that’s an interesting approach actually: having some boolean flag (struct member) which must be set by the first goroutine that manages to close the channel; the subsequent callers must be required to check it before attempting to close the exit channel. 

IMO, thats a viable approach and should solve the problem.

In fact, if I remember ancient versions of Go had a IsClosed builtin function which you could use to check if a channel is closed but it had a performance hit or something like that so Go authors removed it.

milosgajdos:
Sorry for accidentally closing the issue 🤦‍♂️ Reopening it now

milosgajdos:
Take this example for some inspiration:

```Go
package main

import (
	"fmt"
	"sync"
)

type Foo struct {
	exit     chan struct{}
	isClosed bool
	sync.RWMutex
}

func (f *Foo) Close() {
	f.Lock()
	defer f.Unlock()

	select {
	case <-f.exit:
		return
	default:
		if f.isClosed {
			return
		}
		close(f.exit)
		f.isClosed = true
	}

	return
}

func main() {

	foo := &Foo{
		exit:     make(chan struct{}),
		isClosed: false,
	}

	var wg sync.WaitGroup

	for i := 0; i < 100; i++ {
		wg.Add(1)
		go func() {
			defer wg.Done()
			foo.Close()
		}()
	}

	wg.Wait()
	fmt.Printf("Done\n")
}
```

asim:
I feel like I'm missing something. The whole thing around pseudo random selection you're talking about applies to multiple channels being used in a select. Not the default case. The default is the fall through in the event everything else is blocked. It's not included in the random choice. I can show you code that demonstrates this very clearly.

In the case of thread safety. We know select/close evaluated by multiple go routines is not thread safe. If anything this code was never meant to be accessed by multiple go routines. But in the case that it is. We have a mutex there. All threads except one should block on Lock. At which point use of the select is serialised across all callers. Meaning it's not evaluated by multiple callers until the unlock. 

Now if you're saying select inside a function is pre-evaluated by Go and the lock is meaningless in that case then yes I am with you and we'd need some variable to protect it. In which case let's do that and close this pr.

milosgajdos:
@vtolstov can you provide us with some sample cause where we can observe the `panic` please? Otherwise we'll end up coming up with solutions and theories to "made up" problems.

asim:
I'm going to close this for now. It's a long set of comments for a PR that's not going to be merged. I think our point of consensus is to use an isClosed style variable to solve the issue. @vtolstov please test and PR that solution otherwise Milos or I will do it on Monday.

PR #606, opened 2019-07-23
[WIP] router and mucp proxy integration

milosgajdos:
**DO NOT MERGE YET: THIS IS WIP aka POC**

This PR shows a prototype implementation of `router` and `mucp.Proxy` integration.

It adds the following changes to `router` package:
* it refactors `Advertise()` function which now does **only** what it claims to do: **advertising**
* various `router` packages functions/methods have been renamed to make their functionality more obvious and more in line with what they actually do
* function documentation changes related to the above bullet points

It adds the following changes to `mucp.Proxy` package:
* it initializes the internal state of the proxy when creating it
* it does the absolute minimum necessary when looking up the routes in `getRoute()`
* it adds un-exported `watchRoutes()` method which watches routes in remote router
* it reimplements the proxy cache using a similar approach the in-memory implementation of the default routing table uses

### DIFF
--- a/network/proxy/mucp/mucp.go
+++ b/network/proxy/mucp/mucp.go
@@ -134,8 +134,11 @@ func (p *Proxy) getRoute(service string) ([]string, error) {
 
 	// no router is specified we're going to set the default
 	if len(name) == 0 && len(addr) == 0 {
+		// create default router and start it
 		p.Router = router.DefaultRouter
-		go p.Router.Advertise()
+		if err := p.Router.Run(); err != nil {
+			return nil, err
+		}
 
 		// recursively execute getRoute
 		return p.getRoute(service)
@@ -171,43 +174,40 @@ func (p *Proxy) getRoute(service string) ([]string, error) {
 		return nil, selector.ErrNoneAvailable
 	}
 
-	var pbRoutes *pb.LookupResponse
-	var gerr error
-
 	// set default client
 	if p.RouterService == nil {
 		p.RouterService = pb.NewRouterService(name, p.Client)
 	}
 
+	var resp *pb.LookupResponse
+	var err error
+
 	// TODO: implement backoff and retries
 	for _, addr := range addrs {
 		// call the router
-		proutes, err := p.RouterService.Lookup(context.Background(), &pb.LookupRequest{
+		resp, err = p.RouterService.Lookup(context.Background(), &pb.LookupRequest{
 			Query: &pb.Query{
 				Service: service,
 			},
 		}, client.WithAddress(addr))
 		if err != nil {
-			gerr = err
 			continue
 		}
-		// set routes
-		pbRoutes = proutes
 		break
 	}
 
 	// errored out
-	if gerr != nil {
-		return nil, gerr
+	if err != nil {
+		return nil, err
 	}
 
 	// no routes
-	if pbRoutes == nil {
+	if resp == nil || len(resp.Routes) == 0 {
 		return nil, selector.ErrNoneAvailable
 	}
 
 	// convert from pb to []*router.Route
-	for _, r := range pbRoutes.Routes {
+	for _, r := range resp.Routes {
 		routes = append(routes, table.Route{
 			Service: r.Service,
 			Address: r.Address,
@@ -346,8 +346,6 @@ func NewProxy(opts ...options.Option) proxy.Proxy {
 	r, ok := p.Options.Values().Get("proxy.router")
 	if ok {
 		p.Router = r.(router.Router)
-		// TODO: should we advertise?
-		go p.Router.Advertise()
 	}
 
 	return p
--- a/network/proxy/mucp/mucp.go
+++ b/network/proxy/mucp/mucp.go
@@ -3,6 +3,7 @@ package mucp
 
 import (
 	"context"
+	"errors"
 	"io"
 	"os"
 	"strings"
@@ -19,6 +20,10 @@ import (
 
 	pb "github.com/micro/go-micro/network/router/proto"
 	"github.com/micro/go-micro/network/router/table"
+)
+
+var (
+	ErrRouteWatcher = errors.New("route watcher stopped")
 )
 
 // Proxy will transparently proxy requests to an endpoint.
@@ -231,11 +236,16 @@ func (p *Proxy) getRoute(service string) ([]string, error) {
 
 // watchRoutes watches remote router service routes and updates proxy cache
 func (p *Proxy) watchRoutes() {
+	// this is safe to do as the only way watchRoutes returns is
+	// when some error is written into error channel - we want to bail then
+	defer close(p.errChan)
+
 	stream, err := p.RouterService.Watch(context.Background(), &pb.WatchRequest{})
 	if err != nil {
 		p.errChan <- err
 		return
 	}
+	defer stream.Close()
 
 	for {
 		event, err := stream.Recv()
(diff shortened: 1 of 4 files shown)


network/proxy/mucp/mucp.go
```diff
@@ -79,155 +84,126 @@ func readLoop(r server.Request, s client.Stream) error {
...
+	return nodes
+}
 
+func (p *Proxy) getRoute(service string) ([]string, error) {
 	// lookup the route cache first
 	p.RLock()
 	routes, ok := p.Routes[service]
-	// got it!
 	if ok {
-		p.RUnlock()
```
asim, line 101: I think removing this will now cause it to hold read locks since it's supposed to return here

network/proxy/mucp/mucp.go
```diff
@@ -79,155 +84,126 @@ func readLoop(r server.Request, s client.Stream) error {
...
 		return toNodes(routes), nil
 	}
+	p.Routes[service] = make(map[uint64]table.Route)
```
asim, line 109: This is being created while using a read lock so I think it introduces some sort of race

network/proxy/mucp/mucp.go
```diff
@@ -342,13 +329,41 @@ func NewProxy(opts ...options.Option) proxy.Proxy {
...
 	// get router
 	r, ok := p.Options.Values().Get("proxy.router")
 	if ok {
 		p.Router = r.(router.Router)
-		// TODO: should we advertise?
-		go p.Router.Advertise()
 	}
 
+	// create default router and start it
+	if p.Router == nil {
```
asim, line 368: Looking at the getRoute method, the p.Router interface is never nil, so it always uses that. So if MICRO_ROUTER or MICRO_ROUTER_ADDRESS env vars are set these are ignored unless p.Router returns zero routes. So think thats a bug with the logic.

Previous logic:

1. If p.Router interface exists use it
2. If MICRO_ROUTER/MICRO_ROUTER_ADDRESS do not exist, set the default router interface and use that
3. If MICRO_ROUTER or MICRO_ROUTER_ADDRESS are present then use those

network/proxy/mucp/mucp.go
```diff
@@ -79,155 +84,126 @@ func readLoop(r server.Request, s client.Stream) error {
...
+			Address: event.Route.Address,
+			Gateway: event.Route.Gateway,
+			Network: event.Route.Network,
+			Link:    event.Route.Link,
+			Metric:  int(event.Route.Metric),
+		}
+		if _, ok := p.Routes[route.Service]; !ok {
+			p.Routes[route.Service] = make(map[uint64]table.Route)
+		}
+		p.Routes[route.Service][route.Hash()] = route
```
asim: I'm assuming the table event could be a delete or update event along with create so should it act on those?

network/proxy/mucp/mucp.go
```diff
@@ -79,155 +84,126 @@ func readLoop(r server.Request, s client.Stream) error {
...
-	// set default client
-	if p.Client == nil {
-		p.Client = client.DefaultClient
+// watchRoutes watches remote router service routes and updates proxy cache
+func (p *Proxy) watchRoutes() {
+	// this is safe to do as the only way watchRoutes returns is
+	// when some error is written into error channel - we want to bail then
+	defer close(p.errChan)
+
+	stream, err := p.RouterService.Watch(context.Background(), &pb.WatchRequest{})
```
asim, line 196: Should watchRoutes be doing the equivalent of Lookup? Using p.Router.Watch if the interface is present or otherwise p.RouterService.Watch.

asim:
Reading back my own comments and the code now, it feels like having both Router interface and RPC router calls in the same code path complicates things A LOT (entirely my fault for writing shit code).  An alternative approach might be to only use the Router interface and strip out the RPC calls. What do you think?

We could make it a rule that anything with RPC calls between services {proxy, router, ...} should be kept entirely in the github.com/micro/micro toolkit itself. In which case we should shift the RPC related logic from this proxy code to there as a separate Router interface implementation just used by the `micro proxy` in the event MICRO_ROUTER or MICRO_ROUTER_ADDRESS is present.

Otherwise with the fixes above should be able to merge this.

network/proxy/mucp/mucp.go
milosgajdos: Oh, amazing catch! I'll blame this one on my dyslexia 😜

network/proxy/mucp/mucp.go
milosgajdos: Not sure what you mean here -- the `Unlock()` has moved below to line 108 once the cache map has been modified.

network/proxy/mucp/mucp.go
milosgajdos: Another great catch; funnily enough I was in the middle of fixing this after making more tests this morning. Yeah, this is by all means a bug and needs to be fixed (the bug fix is in progress rn).

network/proxy/mucp/mucp.go
milosgajdos: This is a good question and I must admit it's something I had thought about. The thing is, the `p.Router` as it is **only watches routing table events when router.Advertise() has been called**. 

Unless the `router` advertises, it only watches and handles local registry events. If you dont call `Advertise()` what's the point of  watching your own routing table events? That seemed illogical to me as well as calling `Advertise()` here, hence only the remote routes watch...

milosgajdos:
We both sort of suspected having both interfaces (RPC and `go-micro` one) in the `proxy` code would inevitably lead to complexity. 

It also opens an interesting question of coupling and de-coupling of the code between toolkit, framework and in between the packages of the framework: discussion we never had, alas you personally may have thought of that 🤷‍♂️

Oh man, and stop saying `it's my fault, shitty code etc.`. Youre da man 👊

We need to come up with some way to figure out the best trade off between complexity, usability and dev experience. I'll have a think about this today.

Meanwhile, **PLEASE, DO NOT MERGE THIS. ITS A POC**

milosgajdos:
I'm thinking, though...the `router` package code changes maybe could be `cherry-picked` and merged in. That cleanup was badly needed, but then the idiot I am I've made the cardinal `git` rookie mistake: not putting code changes to separate files into separate git commits 🤦‍♂️

I'm pretty sure no cherry picking is possible. Ugh, I'm actually deservedly a moron for doing that.

network/proxy/mucp/mucp.go
milosgajdos: Ah, I always forget about the preference of the `env` variables in `micro`. This whole mixing and matching of the config, env vars etc creates so much `if` jumps in `go-micro` package code; it's not easy to navigate. Or maybe it's just me.

Either way, this is a great catch wrt `micro` way of dealing with env and config stuff. It makes me think we should have some sort of glue code which handles this separately so that this logic does not leak into every single piece of code we write or something 🤔

network/proxy/mucp/mucp.go
milosgajdos: Oh, I got it now. Noticed that as soon as I started modifying code. Nice catch, thanks!

network/proxy/mucp/mucp.go
milosgajdos: I need to have a think about this one. `MICRO_ROUTER` and `MICRO_ROUTER_ADDRESS` have the absolute precedence I take it. I need to come up with some non-hacky/crufty way of doing that in code.

PR #610, opened 2019-07-24
Adds route watcher to mucp.Proxy

milosgajdos:


network/proxy/mucp/mucp.go
```diff
@@ -119,6 +123,55 @@ func (p *Proxy) getRoute(service string) ([]string, error) {
...
+		event, err := w.Next()
+		if err != nil {
+			p.errChan <- err
+			return
+		}
+
+		p.Lock()
+		if err := p.manageRouteCache(event.Route, fmt.Sprintf("%s", event.Type)); err != nil {
+			// TODO: should we bail here?
+			continue
```
asim, line 170: This will need p.Unlock since continue will continue to the next iteration

milosgajdos: FML, good catch.

PR #617, opened 2019-07-27
fix bug

three-zhang:
https://github.com/micro/micro/issues/293
Send request failed using micro Content-Type application/grpc+json


three-zhang:
I am using the consul test without problems
This change seems to have an impact on MDNS, but I don't know why.

Issue #631, opened 2019-08-01
goroutine leak in registry cache when watching a non-exists server 


lqczzz:
here is the example  from [example/gretter](https://github.com/micro/examples/blob/master/greeter/cli/main.go)

i changed the registry to zk registry
```Go
func main() {
	// create a new service
	service := micro.NewService(
		micro.Registry(zookeeper.NewRegistry(
			registry.Addrs("127.0.0.1:2181"),
		)),
	)

	// parse command line flags
	service.Init()

	// Use the generated client stub
	cl := hello.NewSayService("go.micro.srv.greeter", service.Client())

	// Make request
	rsp, err := cl.Hello(context.Background(), &hello.Request{
		Name: "John",
	})
	if err != nil {
		fmt.Println(err)
		// return
	}
	go func() {
		log.Println(http.ListenAndServe("localhost:8090", nil))
	}()
	for {
		time.Sleep(1 * time.Second)
		fmt.Println("num of goroutine is : ", runtime.NumGoroutine())
	}
	fmt.Println(rsp.Msg)
}
```

![image](https://user-images.githubusercontent.com/8279787/62264330-0d999e00-b452-11e9-9f9e-6b80d2a045d6.png)

from above we note that goroutines from cache keep increasing.

it is because the loop-manage goroutine does not exists after the watcher stop.

PR #632, opened 2019-08-01
fixed goroutine leak in registry cache

lqczzz:
fixed
https://github.com/micro/go-micro/issues/631

### DIFF
--- a/registry/cache/rcache.go
+++ b/registry/cache/rcache.go
@@ -328,9 +328,20 @@ func (c *cache) watch(w registry.Watcher) error {
 	defer w.Stop()
 
 	// manage this loop
+	quitLoop := make(chan interface{})
+	defer func() {
+		select {
+		case quitLoop <- nil:
+		default:
+		}
+	}()
 	go func() {
 		// wait for exit
-		<-c.exit
+		select {
+		case <-c.exit:
+		case <-quitLoop:
+		}
+
 		w.Stop()
 	}()
 


registry/cache/rcache.go
```diff
@@ -328,9 +328,20 @@ func (c *cache) watch(w registry.Watcher) error {
 	defer w.Stop()
 
 	// manage this loop
+	quitLoop := make(chan interface{})
+	defer func() {
```
vtolstov, line 332: can you comment this code, i don't understand it

vtolstov (commented):
@asim can you look at this? i Know that in some go-micro codebase we have goroutine leaks (my tests says that, after massive req/rsp is about hour race tests stopped with more that 8129 goroutine exists in the same time)

but i don't understand this pr, how it fix the issue if it present.

asim:
I'm not really sure what this fixes. I know this user issued a PR in the zookeeper registry to fix a go routine leak and we merged that, so that likely solves the issue. This however just duplicates some functionality for exiting. If you look at the overall code, when the exit channel is closed we stop the watcher which should cause it to error and then return. This code specifically should not really leak.

If there are go routine leaks elsewhere we should look into it. I'm closing this since I think the other PR solved the problem.

lqczzz:
> I'm not really sure what this fixes. I know this user issued a PR in the zookeeper registry to fix a go routine leak and we merged that, so that likely solves the issue. This however just duplicates some functionality for exiting. If you look at the overall code, when the exit channel is closed we stop the watcher which should cause it to error and then return. This code specifically should not really leak.
> 
> If there are go routine leaks elsewhere we should look into it. I'm closing this since I think the other PR solved the problem.

even the watcher stops, this goroutine will leak too. because no signal would make it stop
https://github.com/micro/go-micro/blob/master/registry/cache/rcache.go#L331

registry/cache/rcache.go
lqczzz: ok, actually it is because that even the watcher stops, [this goroutine](https://github.com/micro/go-micro/blob/master/registry/cache/rcache.go#L331) wont stop ,and we will make a new watcher and watch again, which leads to the number of goroutines increasing

milosgajdos: @vtolstov this function is a `defer` function so it's called when `watch()` returns. It unblocks (exits it) the following goroutine:

```Go
go func() {
		// wait for exit
		select {
		case <-c.exit:
		case <-quitLoop:
		}

		w.Stop()
	}()
```

which originally looked like this:
```Go
// manage this loop
	go func() {
		// wait for exit
		<-c.exit
		w.Stop()
	}()
```

which would otherwise be only unblocked only if `close(c.exit)` is called somewhere. Unless `close(c.exit)` is called somewhere this goroutine might really be "leaking".

My issue here is though. if we DO unblock the goroutine shown above, we will have **double** `w.Stop()` call, which I'm not sure can be done. What I mean by double:
* `watch` returns
* it calls `w.Stop()` (the first line in the `watch` function has the `defer` call)
* the "leak" goroutine gets unblocked and it **also** calls `w.Stop()`

Just my thoughts here...

lqczzz: great, i think the w.Stop should be re-entrantable(call repeatedly)。

or

```
case <-quitLoop:
    return // wont call w.Stop() again
```

emm, seems double call `w.Stop()` will wtill happen
(the original code will cause double call too)

PR #634, opened 2019-08-01
Stop a goroutine leak in registy

### LINKED ISSUE #631: goroutine leak in registry cache when watching a non-exists server

asim:
We have a go routine leak in the registry cache. When the watcher errors it never kills the go routine thats waiting on exit, so we slow leak. The issue is here https://github.com/micro/go-micro/issues/631 and a previous PR fix is here https://github.com/micro/go-micro/pull/632. 

The code is basically identical but a bit cleaner by only caller w.Stop once inside the go routine.

Fixes https://github.com/micro/go-micro/issues/631

### DIFF
--- a/registry/cache/rcache.go
+++ b/registry/cache/rcache.go
@@ -325,18 +325,27 @@ func (c *cache) run(service string) {
 // watch loops the next event and calls update
 // it returns if there's an error
 func (c *cache) watch(w registry.Watcher) error {
-	defer w.Stop()
+	// used to stop the watch
+	stop := make(chan bool)
 
 	// manage this loop
 	go func() {
+		defer w.Stop()
+
+		select {
 		// wait for exit
-		<-c.exit
-		w.Stop()
+		case <-c.exit:
+			return
+		// we've been stopped
+		case <-stop:
+			return
+		}
 	}()
 
 	for {
 		res, err := w.Next()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


milosgajdos:
Nice! We are also avoiding double `watcher.Stop()` 👍

Good catch @lqczzz in #631

PR #641, opened 2019-08-05
fix Address more then one

514366607:
when consul Node down
new service can work up.

### DIFF
--- a/registry/consul/consul.go
+++ b/registry/consul/consul.go
@@ -17,9 +17,11 @@ import (
 )
 
 type consulRegistry struct {
-	Address string
-	Client  *consul.Client
+	Address []string
 	opts    registry.Options
+
+	client       *consul.Client
+	consulConfig *consul.Config
 
 	// connect enabled
 	connect bool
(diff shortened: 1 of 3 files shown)


asim:
This PR introduces a second call to check the consul agent every time we use the client. Does this not seem excessive?

514366607:
Only the first time an object is created is it checked , when client is nil。

asim:
How does this solve the problem of something working when the agent is down? Surely you'd have to rotate through the client list on every call to Client() no?

514366607:
Node consul A   IP : 192.168.0.11
Node consul B   IP : 192.168.0.12
Node consul C   IP : 192.168.0.13

service A
```
re := consul.NewRegistry(
		registry.Addrs("192.168.0.11:8500", "192.168.0.12:8500", "192.168.0.13:8500"),
		consul.TCPCheck(time.Second),
	)
```

When consul Node A goes down
The new service can Register at consul Node B | C

514366607:
When consul Node A goes down
then leader is consul Node B | C.

If Service A goes down will exit after 1 second(HealthCheck)

asim:
Alright, cool

PR #651, opened 2019-08-07
wait nats drain since it's asynchronous

magodo:
1. nats subscription draining is removed, since it is asynchronous,
   and there is no reliable way to detect when it is finished.
   Remove this option to avoid confusion.
2. nats connection draining is kept, and use 2 callbacks to detect
   draining timeout (timeout is set via `nats.Options`) or finish.
3. Also honour options passed in `broker.Init()` (previously only
   `broker.New()` is honoured).

### DIFF
--- a/broker/nats/nats.go
+++ b/broker/nats/nats.go
@@ -13,18 +13,19 @@ import (
 )
 
 type natsBroker struct {
+	sync.Once
 	sync.RWMutex
-	addrs []string
-	conn  *nats.Conn
-	opts  broker.Options
-	nopts nats.Options
-	drain bool
+	addrs   []string
+	conn    *nats.Conn
+	opts    broker.Options
+	nopts   nats.Options
+	drain   bool
+	closeCh chan (error)
 }
 
 type subscriber struct {
-	s     *nats.Subscription
-	opts  broker.SubscribeOptions
-	drain bool
+	s    *nats.Subscription
+	opts broker.SubscribeOptions
 }
 
 type publication struct {
@@ -54,9 +55,6 @@ func (s *subscriber) Topic() string {
 }
 
 func (s *subscriber) Unsubscribe() error {
-	if s.drain {
-		return s.s.Drain()
-	}
 	return s.s.Unsubscribe()
 }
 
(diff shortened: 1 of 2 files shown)


asim:
@vtolstov can you look at this?

vtolstov:
please wait, i'm try to look

broker/nats/nats.go
```diff
@@ -207,39 +197,59 @@ func NewBroker(opts ...broker.Option) broker.Broker {
...
+
+func (n *natsBroker) setOption(opts ...broker.Option) {
+	for _, o := range opts {
+		o(&n.opts)
 	}
 
-	var drain bool
-	if _, ok := options.Context.Value(drainSubscriptionKey{}).(bool); ok {
-		drain = true
+	n.Once.Do(func() {
```
vtolstov, line 213: why you guard this? setOption called in Init and in NewBroker

vtolstov (changes requested):
looks good, but i don't understand why using sync.Once ? Save one call to GetDefaultOptions ? or i miss something?

broker/nats/nats.go
magodo: This is because we need only set default option in the first time. Otherwise, the nats option set in `New` will be reset in `Init`

vtolstov: but New and Init both have opts and needs to allow overwrite options stored in broker.... @asim i'm right?

asim: Yea, Init is for reinitialising options

asim: But I guess what this code is saying is that we only want to fetch default options once. We don't want to overwrite existing options with default options. So it sort of makes sense.

asim: e.g New sets default options and adds addrs then Init adds something else, you don't want init to replace addrs

vtolstov: > But I guess what this code is saying is that we only want to fetch default options once. We don't want to overwrite existing options with default options. So it sort of makes sense.

makes sense.

vtolstov:
@magodo thanks

broker/nats/nats.go
magodo: Asim is right, it is what i mean here. Though I can kind of add some code seperately in `New()` only, instead of using `sync.Once`. While i think it seems more explicit to put them together.

Issue #659, opened 2019-08-09
Error message is helpless when client occur service not found


xpunch:
When client is request a service, but the service is not registered or service name is incorrect, we'll get an error like this:
```
{"id":"go.micro.client","code":500,"detail":"not found","status":"Internal Server Error"}
```
It's really hard to determine which service is not found, especially multiple service is called through the request chain.

Can we change the error message like this:
```
if err != nil && err == ErrNotFound {
	return nil, errors.NotFound(service, "service not found")
} else if err != nil {
	return nil, errors.InternalServerError("go.micro.client", err.Error())
}
``` 
Previous code is:
```
if err != nil && err == ErrNotFound {
	return nil, errors.NotFound("go.micro.client", err.Error())
} else if err != nil {
	return nil, errors.InternalServerError("go.micro.client", err.Error())
}
```

asim:
Sounds good. Feel free to PR the change

asim:
Your issue has been fixed. In future please reference the implementation you're speaking about.

PR #660, opened 2019-08-11
Service not found issue

xpunch:
Change the registry.ErrNotFound error message, and update some other changes.


xpunch:
#659 service not found error message

xpunch:
I was going to pull my implementation, but after a little testing, you closed my issue.
1. selector.ErrNotFound is only used by the router, but registry.ErrNotFound is returned at this time.
2. [gprc service](https://github.com/micro/go-micro/blob/8986b3135f72770b70596871e4ecd8c508c7b354/server/grpc/grpc.go#L273) 
```go
if service == nil {
	return status.New(codes.Unimplemented, fmt.Sprintf("unknown service %v", service)).Err()
}
```
service is nil, so it will just print <nil>, use serviceName is better.

asim:
I can't accept this change. The client uses the selector. The registry doesn't have to be used within the selector. Leaking the registry error and making use of that is wrong. We only need to look for the select error.

xpunch:
OK, I'll change the registry error back, but this issue occurs by below code:
[client/grpc.go](https://github.com/micro/go-micro/blob/e613b0c2058d31d8e22daa3162a1bbcaaf6d6923/client/grpc/grpc.go#L75)
```go
next, err := g.opts.Selector.Select(service, opts.SelectOptions...)
	if err != nil && err == registry.ErrNotFound {}
}
```
[selector/default.go](https://github.com/micro/go-micro/blob/e613b0c2058d31d8e22daa3162a1bbcaaf6d6923/client/selector/default.go#L52)
```go
func (c *registrySelector) Select(service string, opts ...SelectOption) (Next, error) {
	services, err := c.rc.GetService(service)
	if err != nil {
		return nil, err
	}
}
```
[registry/cache.go](https://github.com/micro/go-micro/blob/e613b0c2058d31d8e22daa3162a1bbcaaf6d6923/registry/cache/rcache.go#L363)
```go
func (c *cache) GetService(service string) ([]*registry.Service, error) {
	// get the service
	services, err := c.get(service)
	if err != nil {
		return nil, err
	}
	// if there's nothing return err
	if len(services) == 0 {
		return nil, registry.ErrNotFound
	}
	// return services
	return services, nil
}
```

Do you think this change is ok?
[client/selector/default.go](https://github.com/micro/go-micro/blob/e613b0c2058d31d8e22daa3162a1bbcaaf6d6923/client/selector/default.go#L52)
```go
// get the service
// try the cache first
// if that fails go directly to the registry
services, err := c.rc.GetService(service)
if err != nil {
	if err == registry.ErrNotFound {
		return nil, ErrNotFound
	}
	return nil, err
}
```

asim:
That's fine. Can you squash the commits please.

xpunch:
OK, I'll squash my commits later.

xpunch:
I've squash my commits.

PR #664, opened 2019-08-13
Fix RabbitMQ Transport bug

printfcoder:
I tested the NATS case. it is OK now.
And we also need to fix the rabbit case.

but I really think to fix them in HostPort is not the best choice.


asim:
What is the "rabbitmq case". This PR doesn't explain what it actually fixes.

printfcoder:
When using the RabbitMQ-Transport and Consul registry:
1. The address of the node will be registered as '127.0.0.1:0', which should be a queue name.
2. "rabbitmq case" means that "addr" is a queue name and port is an int "0", so it can't be drop in the "if" of Nats case.

vtolstov:
may be we need to parse url? and switch based on url Schema ? and url must be in
nats://
rabbitmq://
kafka://

or something like this? so we can parse it based on schema?

asim:
Can you squash the commit and also change the comments so its specific to why its doing it without mention of nats or rabbitmq.

asim:
> may be we need to parse url? and switch based on url Schema ? and url must be in
> nats://
> rabbitmq://
> kafka://
> 
> or something like this? so we can parse it based on schema?

Why though? We're not setting the address of the broker or transport. This is the service calling Listen. Should the user not have some instinct about what address to set considering its used as the topic for receiving messages?

vtolstov:
@asim sorry

printfcoder:
close and squash commit

PR #669, opened 2019-08-14
fix registry addr error for mq-rpc

printfcoder:
related [PR664](https://github.com/micro/go-micro/pull/664)

When using the RabbitMQ-Transport and Consul registry:

The address of the node will be registered as '127.0.0.1:0', which should be a queue name.
"rabbitmq case" means that "addr" is a queue name and port is an int "0", so it can't be drop in the "if" of Nats case.

### DIFF
--- a/registry/consul/consul.go
+++ b/registry/consul/consul.go
@@ -13,6 +13,7 @@ import (
 
 	consul "github.com/hashicorp/consul/api"
 	"github.com/micro/go-micro/registry"
+	mnet "github.com/micro/go-micro/util/net"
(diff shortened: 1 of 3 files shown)


printfcoder:
TestLoopbackTunnel fails. never changed it.

milosgajdos:
`TestLoopbackTunnel` is now fixed on `master`. Can you rebase to `master` @printfcoder ? That should fix the build.

printfcoder:
@milosgajdos83 rebased

PR #672, opened 2019-08-15
Monitor outbound links and delete them when disconnected

milosgajdos:
Send keepalive ping messages on outbound links.

### DIFF
--- a/tunnel/default.go
+++ b/tunnel/default.go
@@ -138,8 +138,8 @@ func (t *tun) monitor() {
 			return
 		case <-reconnect.C:
 			for _, node := range t.options.Nodes {
+				t.Lock()
 				if _, ok := t.links[node]; !ok {
-					t.Lock()
 					link, err := t.setupLink(node)
 					if err != nil {
 						log.Debugf("Tunnel failed to setup node link to %s: %v", node, err)
@@ -147,8 +147,8 @@ func (t *tun) monitor() {
 						continue
 					}
 					t.links[node] = link
-					t.Unlock()
 				}
+				t.Unlock()
 			}
 		}
 	}
--- a/tunnel/default.go
+++ b/tunnel/default.go
@@ -14,8 +14,10 @@ import (
 )
 
 var (
-	// DefaultKeepAlive defines interval we check if the links are alive
-	DefaultKeepAlive = 30 * time.Second
+	// KeepAliveTime defines time interval we send keepalive messages to outbound links
+	KeepAliveTime = 30 * time.Second
+	// ReconnectTime defines time interval we periodically attempt to reconnect dead links
+	ReconnectTime = 5 * time.Second
 )
 
 // tun represents a network tunnel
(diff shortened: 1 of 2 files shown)


tunnel/default.go
```diff
@@ -187,14 +203,18 @@ func (t *tun) listen(link *link) {
...
-				loopback = true
 				link.loopback = true
 			}
 			continue
 		case "close":
+			log.Debugf("Tunnel link %s closing connection", link.Remote())
 			// TODO: handle the close message
 			// maybe report io.EOF or kill the link
 			continue
+		case "ping":
```
asim: Change to "keepalive"

tunnel/default.go
```diff
@@ -286,6 +306,36 @@ func (t *tun) listen(link *link) {
...
+	for {
+		select {
+		case <-t.closed:
+			return
+		case <-ping.C:
+			// send ping message
+			log.Debugf("Tunnel sending ping to link %v: %v", link.Remote(), link.id)
+			if err := link.Send(&transport.Message{
+				Header: map[string]string{
+					"Micro-Tunnel":       "ping",
```
asim: Change to "keepalive"

milosgajdos83: 👍

tunnel/default.go
```diff
@@ -286,6 +334,74 @@ func (t *tun) listen(link *link) {
...
+				}
+				// TODO: handle this error
+			}
+		}
+	}
+}
+
+// nodeLink attempts to connect to node and returns link if successful
+// it returns error if the link failed to be established
+func (t *tun) nodeLink(node string) (*link, error) {
```
asim: I was confused by the name nodeLink, I assumed it returned a link and was going to say to change it to getLink but then on further examination of the method it looks like it does the whole connect flow. Maybe it should be called setupLink? or just setup?

milosgajdos: yeah I called it `nodeLink` because it creates links to nodes, but come to think of it, `setupLink` is much better. Will change it.

tunnel/default.go
```diff
@@ -118,6 +128,32 @@ func (t *tun) newSession() string {
...
+		case <-t.closed:
+			return
+		case <-reconnect.C:
+			for _, node := range t.options.Nodes {
+				t.Lock()
+				if _, ok := t.links[node]; !ok {
+					link, err := t.setupLink(node)
+					if err != nil {
+						log.Debugf("Tunnel failed to establish node link to %s: %v", node, err)
+						continue
```
asim, line 147: Must call t.Unlock() here

tunnel/default.go
```diff
@@ -139,16 +138,17 @@ func (t *tun) monitor() {
 			return
 		case <-reconnect.C:
 			for _, node := range t.options.Nodes {
-				t.Lock()
```
asim: This should still be here as its a race condition to read from the map without a lock

milosgajdos: Oops. Beats me how and why I swapped the order here 🤦‍♂️

PR #676, opened 2019-08-17
fix https://github.com/micro/go-micro/issues/625 

### LINKED ISSUE #625: grpc client error  " Unsupported Content-Type: application/grpc+proto", look like bug
I use grpc server as micro server, httpbroker as broker, when I try publish a message, but some error occurs:  {"id":"go.micro.client","code":500,"detail":"Unsupported Content-Type: application/grpc+proto","status":"Internal Server Error"}

mycode:
```
func main() {
	reg := etcdv3.NewRegistry(func(op *registry.Options) {
		op.Addrs = []string{
			"http://127.0.0.1:2379",
		}
	})

	service := micro.NewService(
		micro.Server(grpc.NewServer()),
		micro.Client(grpc.NewClient()),
		micro.Name("HelloGreeter1"),
		micro.Registry(reg),
	)

	service.Init()
	ctx := context.Background()
	cmsg := "message :  Hello World"
	err = service.Client().Publish(ctx, service.Client().NewMessage("testTopic", cmsg))
	if err != nil {
		fmt.Println(err.Error())
	}

	// blocking, cmsg can be sent, because Publish method is async
	select {
	}
}
```

debug find:
/client/grpc/grpc.go 489line
```
func (g *grpcClient) Publish(ctx context.Context, p client.Message, opts ...client.PublishOption) error {
	md, ok := metadata.FromContext(ctx)
	if !ok {
		md = make(map[string]string)
	}
	md["Content-Type"] = p.ContentType()

	cf, err := g.newCodec(p.ContentType()) // here
}
```

/client/grpc/grpc.go 262line
```
func (g *grpcClient) newCodec(contentType string) (codec.NewCodec, error) {
	if c, ok := g.opts.Codecs[contentType]; ok {
		return c, nil
	}
       
       // here,  I think should use defaultGRPCCodecs
	if cf, ok := defaultRPCCodecs[contentType]; ok {
		return cf, nil
	}
	return nil, fmt.Errorf("Unsupported Content-Type: %s", contentType)
}
```

three-zhang:
fix https://github.com/micro/go-micro/issues/625 solve: Unsupported Content-Type: application/grpc+proto

### DIFF
--- a/micro.go
+++ b/micro.go
@@ -67,11 +67,16 @@ func NewFunction(opts ...Option) Function {
 }
 
 // NewPublisher returns a new Publisher
-func NewPublisher(topic string, c client.Client) Publisher {
+func NewPublisher(topic string, c client.Client, contentType... string) Publisher {
 	if c == nil {
 		c = client.NewClient()
 	}
-	return &publisher{c, topic}
+	// compatible old version
+	if len(contentType) > 0 {
+		return &publisher{c, topic, contentType[0]}
+	} else {
+		return &publisher{c, topic, ""}
+	}
 }
 
 // RegisterHandler is syntactic sugar for registering a handler
--- a/publisher.go
+++ b/publisher.go
@@ -7,10 +7,17 @@ import (
 )
 
 type publisher struct {
-	c     client.Client
-	topic string
+	c           client.Client
+	topic       string
+	contentType string
 }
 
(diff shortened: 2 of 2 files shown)


asim:
This is a more appropriate fix https://github.com/micro/go-micro/pull/679

Let me know your thoughts.

asim:
Merged the other PR

PR #677, opened 2019-08-17
 fix file watcher event bug on Linux

printfcoder:
On Linux(ubuntu, centos), file watcher can only get the file change event twice.

can be tested with this [example](https://github.com/micro/examples/tree/master/config/grpc)


asim:
Is this only limited to linux? It feels like the change should be applied more broadly.

printfcoder:
@asim yes, I just added a Linux case comment `// + build Linux`, and then other operating systems use `// +build !Linux` which just like before.

asim:
Can you please keep the same structure of the code, we should have watcher.go and watcher_linux.go

printfcoder:
@asim done. please have a check

asim:
Looks good. Can you now squash the commit please.

PR #685, opened 2019-08-20
Default network implementation

milosgajdos:
This PR is a WIP version of default network implementation.

### DIFF
--- a/network/default.go
+++ b/network/default.go
@@ -1,14 +1,47 @@
 package network
 
 import (
+	"sync"
+	"time"
+
+	"github.com/gogo/protobuf/proto"
 	"github.com/micro/go-micro/client"
+	"github.com/micro/go-micro/proxy"
+	"github.com/micro/go-micro/router"
+	pb "github.com/micro/go-micro/router/proto"
 	"github.com/micro/go-micro/server"
+	"github.com/micro/go-micro/transport"
+	"github.com/micro/go-micro/tunnel"
+	tr "github.com/micro/go-micro/tunnel/transport"
+	"github.com/micro/go-micro/util/log"
+)
+
+var (
+	// ControlChannel is the name of the tunnel channel for passing contron message
+	ControlChannel = "control-msg"
 )
 
 // network implements Network interface
 type network struct {
 	// options configure the network
+	// TODO: we might end up embedding
 	options Options
+	// rtr is network router
+	router.Router
+	// prx is network proxy
+	proxy.Proxy
+	// tun is network tunnel
+	tunnel.Tunnel
+	// srv is network server
+	srv server.Server
+	// client is network client
+	client client.Client
+
+	sync.RWMutex
+	// connected marks the network as connected
+	connected bool
+	// closed closes the network
+	closed chan bool
 }
 
 // newNetwork returns a new network node
@@ -19,8 +52,33 @@ func newNetwork(opts ...Option) Network {
 		o(&options)
 	}
 
+	// init tunnel address to the network bind address
+	options.Tunnel.Init(
+		tunnel.Address(options.Address),
+	)
+
+	// create tunnel client with tunnel transport
+	tunTransport := transport.NewTransport(
+		tr.WithTunnel(options.Tunnel),
+	)
+
+	// srv is network server
+	srv := server.NewServer(
+		server.Transport(tunTransport),
+	)
+
+	// client is network client
+	client := client.NewClient(
+		client.Transport(tunTransport),
+	)
+
 	return &network{
 		options: options,
+		Router:  options.Router,
+		Proxy:   options.Proxy,
+		Tunnel:  options.Tunnel,
+		srv:     srv,
+		client:  client,
 	}
 }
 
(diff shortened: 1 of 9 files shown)


network/default.go
```diff
@@ -0,0 +1,336 @@
...
+	if err != nil {
+		return err
+	}
+
+	// advertise routes
+	go n.advertise(client, advertChan)
+	// process routes
+	go n.process(client)
+
+	// set connected to true
```
asim, line 337: Start the server

network/default.go
```diff
@@ -0,0 +1,336 @@
...
+	n.connected = true
+
+	return nil
+}
+
+func (n *network) close() error {
+	// stop the router
```
asim, line 349: Stop the server

network/default.go
```diff
@@ -0,0 +1,336 @@
...
+	)
+
+	// srv is network server
+	srv := server.NewServer(
+		server.Transport(tunTransport),
+	)
+
+	// client is network client
+	client := client.NewClient(
+		client.Transport(tunTransport),
```
asim, line 73: Set the router selector

```
import "github.com/micro/go-micro/client/selector/router"

client.Selector(
    router.NewSelector(
        router.WithRouter(r),
    ),
)
```

network/network.go
```diff
@@ -1,2 +1,37 @@
...
+	DefaultAddress = ":0"
+	// ResolveTime ddefines the time we periodically resolve network nodes
+	ResolveTime = 1 * time.Minute
+)
+
+// Network is micro network
+type Network interface {
+	// Name of the network
+	Name() string
+	// Connect starts the resolver and tunnel server
```
asim, line 26: Add `Address() string` method to the network interface to expose the `tunnel.Address()`

network/network.go
```diff
@@ -1,2 +1,37 @@
...
+import (
+	"time"
+
+	"github.com/micro/go-micro/client"
+	"github.com/micro/go-micro/server"
+)
+
+var (
+	// DefaultName is default network name
+	DefaultName = "go.micro.network"
```
asim: Set network name to `go.micro`

network/default.go
```diff
@@ -0,0 +1,336 @@
...
+	)
+
+	// create tunnel client with tunnel transport
+	tunTransport := transport.NewTransport(
+		tr.WithTunnel(options.Tunnel),
+	)
+
+	// srv is network server
+	srv := server.NewServer(
+		server.Transport(tunTransport),
```
asim, line 68: Set server name to the same as the network name

network/default.go
```diff
@@ -0,0 +1,336 @@
...
+}
+
+// Name returns network name
+func (n *network) Name() string {
+	return n.options.Name
+}
+
+// Address returns network bind address
+func (n *network) Address() string {
+	return n.options.Address
```
asim: return n.tunnel.Address()

network/default.go
```diff
@@ -0,0 +1,336 @@
...
+		return err
+	}
+
+	// initialize the tunnel to resolved nodes
+	n.Tunnel.Init(
+		tunnel.Nodes(nodes...),
+	)
+
+	// dial into ControlChannel to send route adverts
+	client, err := n.Tunnel.Dial(ControlChannel)
```
asim, line 299: Should we also be listening on the control channel?

network/default.go
```diff
@@ -0,0 +1,336 @@
...
+	"github.com/micro/go-micro/server"
+	"github.com/micro/go-micro/transport"
+	"github.com/micro/go-micro/tunnel"
+	tr "github.com/micro/go-micro/tunnel/transport"
+	"github.com/micro/go-micro/util/log"
+)
+
+var (
+	// ControlChannel is the name of the tunnel channel for passing contron message
+	ControlChannel = "control-msg"
```
asim: Should we name this to `router` or something similar?

network/default.go
milosgajdos: Yes we should! 😜

network/default.go
milosgajdos: Should we consider the router to be started or should we start the router here, too?

network/default.go
milosgajdos: Ditto. Should we stop the router here as well?

network/default.go
milosgajdos: That depends: are we going to be sending **ONLY router** adverts on this or are we going to re-use this particular `tunnel` channel for other "control" messages?

network/default.go
asim: We need to start the router

network/default.go
asim: Yea

network/default.go
asim: Let's just call it control and reuse it for now. Probably keeps things simple.

network/default.go
milosgajdos: I need to make the router start idempotent first 😅

asim: You have the status that will tell you if its running right?

milosgajdos: Yeah, but that code needs to be written. On it now.

asim: Pushed that change

network/default.go
milosgajdos: Actually...should we? 🧐

network/default.go
```diff
@@ -0,0 +1,389 @@
...
+		case advert := <-advertChan:
+			// create a proto advert
+			var events []*pb.Event
+			for _, event := range advert.Events {
+				route := &pb.Route{
+					Service: event.Route.Service,
+					Address: event.Route.Address,
+					Gateway: event.Route.Gateway,
+					Network: event.Route.Network,
+					Link:    event.Route.Link,
```
asim: Should we set the link here to "network" or on receiving?

milosgajdos: Not sure. On send we set it to "network" saying, hey this is a "remote" route. On receive you should just process the routes as you received them, no?

asim: Network field should not be modified. Link field should be set to "Network". This may be why it's not working.

milosgajdos: Yeah I got Network and Link mixed up -- my brain short-circuited the string "network" with `Route.Network` lmao. Nice one!

PR #696, opened 2019-08-23
Make server Start() and Stop() idempotent

milosgajdos:


vtolstov:
```
+	select {
+	case err = <-ch:
+		s.started = false
+	}
+
```

you miss the lock when assign to s.started

milosgajdos:
Great spot @vtolstov 🥇

PR #701, opened 2019-08-26
Check last for the address binded in lo interface (LVS , DR mode)

h-hy:
ignore Loopback Address
In LVS,DR mode, we need to bind the vip in iface "lo";
but i should not be registed.

consul:
https://github.com/hashicorp/consul/blob/master/ipaddr/detect.go

### DIFF
--- a/util/addr/addr.go
+++ b/util/addr/addr.go
@@ -40,6 +40,7 @@ func Extract(addr string) (string, error) {
 	}
 
 	var addrs []net.Addr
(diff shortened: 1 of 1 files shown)


asim:
What if loopback is the only interface?

h-hy:
> What if loopback is the only interface?

yes , I hope fix it now.
Check last for the address binded in lo interface .

asim:
Thanks

PR #705, opened 2019-08-28
stream bug

lpxxn:
in the `stream` function,  need a client connection with flag `Stream = true`
https://github.com/micro/go-micro/blob/a4f5772555f017fdfde44c5a352c5ae42eb207eb/client/rpc_client.go#L208

if use pool get Connection
https://github.com/micro/go-micro/blob/a4f5772555f017fdfde44c5a352c5ae42eb207eb/client/rpc_client.go#L215
there will be a bug, maybe get an  old client connect with  `Stream = false`
and  will get stuck here
https://github.com/micro/go-micro/blob/a4f5772555f017fdfde44c5a352c5ae42eb207eb/transport/http_transport.go#L129

### DIFF
--- a/client/rpc_client.go
+++ b/client/rpc_client.go
@@ -212,7 +212,7 @@ func (r *rpcClient) stream(ctx context.Context, node *registry.Node, req Request
 		dOpts = append(dOpts, transport.WithTimeout(opts.DialTimeout))
 	}
 
-	c, err := r.pool.Get(address, dOpts...)
+	c, err := r.opts.Transport.Dial(address, dOpts...)
 	if err != nil {
 		return nil, errors.InternalServerError("go.micro.client", "connection error: %v", err)
 	}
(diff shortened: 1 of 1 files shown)


asim:
Should we not fix the transport so that it supports streaming?

lpxxn:
I have no other way to fix this bug, if you think this is not correct，I can close this request. but this is really a bug。

asim:
Technically internally the client.Call request is a stream, so surely we could treat it as such and said transport.WithStream on the Call side as well?

lpxxn:
the logic of the current code is that
the  `Stream` of `Transport` is set to true or false only when a client `Transport` is created.
https://github.com/micro/go-micro/blob/a4f5772555f017fdfde44c5a352c5ae42eb207eb/transport/http_transport.go#L509

this is not appropriate,  we should set the `Stream` value to `true` or `false` when using the `Transport`

if we do this, we can use the pool to get the client connection and set the Stream value to true or false 
https://github.com/micro/go-micro/blob/a4f5772555f017fdfde44c5a352c5ae42eb207eb/client/rpc_client.go#L215

asim:
What I'm asking is, if you've tested it with the stream being true for the client. Because I believe this will work.

lpxxn:
yes,i have tested it.if the `stream` is true then it will work

but the `Stream` value of some  connections in the pool is false

asim:
Ok so shouldn't this fix work then? https://github.com/micro/go-micro/pull/717

I'm not sure why you didn't submit this as opposed to removing the use of the connection pool.

PR #706, opened 2019-08-28
Broadcast neighbourhood

milosgajdos:
This commit introduces neighbour(hood) announcements which allow to maintain a map of neighbours of each node.

It also fixes a critical bug when accepting connections on a particular tunnel channel. Previously we would only accept one connection, now we run accept continuously in a dedicated goroutine.


network/default.go
```diff
@@ -39,11 +50,18 @@ type network struct {
 	// client is network client
 	client client.Client
 
+	// ctrlClient is ControlChannel client
+	ctrlClient transport.Client
```
asim: Map of clients in case this grows? map[string]transport.Client with channel name as the key.

milosgajdos: Yeah. Good shout. Can't think of a use case as of now but `ControlChannel` might come in handy to hash over.

network/default.go
```diff
@@ -39,11 +50,18 @@ type network struct {
...
+	// netClient is NetwrokChannel client
+	netClient transport.Client
+
 	sync.RWMutex
 	// connected marks the network as connected
 	connected bool
 	// closed closes the network
 	closed chan bool
+	// neighbours maps the node neighbourhood
+	neighbours map[node]map[node]bool
```
asim: Hmm. Whenever I see map in a map it makes me uncomfortable. Highlights the need for a struct. Maybe store as `map[string]*node` with the below struct and node id as key.

```
type node struct {
  id string
  address string
  neighbours map[string]*node
}
```

milosgajdos: This is normal thing to do in other languages (Python, Ruby and Rust), including [Go](https://blog.golang.org/go-maps-in-action).

Using `node` as a key makes it easy to iterate over and perform ops on the key object, but I'm happy to rework this of course.

network/default.go
asim: Agree, can't think of a better name yet...

PR #707, opened 2019-08-29
set AllowUnknownFields to true

jiyeyuran:
fix: json to proto failed if a field is in json but not in proto.

### DIFF
--- a/codec/json/json.go
+++ b/codec/json/json.go
@@ -25,7 +25,9 @@ func (c *Codec) ReadBody(b interface{}) error {
 		return nil
 	}
 	if pb, ok := b.(proto.Message); ok {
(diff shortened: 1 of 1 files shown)


vtolstov:
Why this needed?why you need unknown fields in request?

jiyeyuran:
To be compatible with the old protocols which will carry some deprecated fields.

vtolstov:
In this case you need to have in you never app this fields. I'm not agree to unconditionally allow unknown fields.

asim:
I agree with Vasiliy. I think you should have a codec that does this for you rather than broadly enabling it for everyone. Either that or provide a way to programatticaly set this.

PR #713, opened 2019-08-29
Avoid setting routes that route back to the node without its being direct GW to dest

milosgajdos:
Should address https://github.com/micro/development/projects/3#card-25690276

### DIFF
--- a/network/default.go
+++ b/network/default.go
@@ -14,7 +14,7 @@ import (
 	"github.com/micro/go-micro/server"
 	"github.com/micro/go-micro/transport"
 	"github.com/micro/go-micro/tunnel"
-	trn "github.com/micro/go-micro/tunnel/transport"
+	tun "github.com/micro/go-micro/tunnel/transport"
 	"github.com/micro/go-micro/util/log"
 )
 
@@ -84,8 +84,8 @@ func newNetwork(opts ...Option) Network {
 	)
 
 	// create tunnel client with tunnel transport
-	tunTransport := trn.NewTransport(
-		trn.WithTunnel(options.Tunnel),
+	tunTransport := tun.NewTransport(
+		tun.WithTunnel(options.Tunnel),
 	)
 
 	// server is network server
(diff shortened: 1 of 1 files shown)


network/default.go
```diff
@@ -439,11 +439,42 @@ func (n *network) processCtrlChan(l tunnel.Listener) {
...
+					if advertNode.address == "" {
+						advertNode.address = event.Route.Gateway
+					}
+					// if advertising node id is not the same as Route.Router
+					// we know the advertising node is not the origin of the route
+					if advertNode.id != event.Route.Router {
+						// if the origin router is not in the advertising node neighbourhood
+						// we can't rule out potential routing loops so we bail here
+						if _, ok := advertNode.neighbours[event.Route.Router]; !ok {
+							continue
```
asim, line 475: A secondary check here might be necessary. 

If event.Route.Router is in your neighbour list, check if the advert node is in that list. If it's not then you can also ignore.

milosgajdos: Yeah, thats a good point (checking bidirectionality). I actually had that in my original code before I changed that back to this version and I can't remember now why >_<

milosgajdos: Oh I know why. I figured if the `event.Route.Router` is in `advertNode` neighbour list that means we can route to `Route.Router` directly from `advertNode`, hence no need to check the other direction. Felt redundant -- as sooner or later they'll announce themselves to each other.

asim: Wait, here's a question. Does that mean we limit routes that we know of to one hop from our neighbour? Meaning metric of 100. Should we be doing this programmatically by defining a hop count so we actually get to that metric of 1000? Or is this a secondary concern? We know we don't want all the routes but how many do we want?

milosgajdos: I think having a hop count is an interesting one. So, we can tell hop count up to 2 from the direct neighbourhood list and neighbours' neighbourhood. That is reflected in calculating the Metric in `setRouteMetric`: anything beyond hop count 2 is assigned Metric 1000

Think the idea of allowing a certain size of neighbourhood is an interesting thought to consider.

asim: I'll make a note rather than adding a card for this and merge the current PR.

PR #717, opened 2019-08-31
use with stream for client connection

asim:


client/rpc_client.go
```diff
@@ -96,7 +96,15 @@ func (r *rpcClient) call(ctx context.Context, node *registry.Node, req Request,
...
-	c, err := r.pool.Get(address, transport.WithTimeout(opts.DialTimeout))
+	dOpts := []transport.DialOption{
+		transport.WithStream(),
+	}
+
+	if opts.DialTimeout >= 0 {
+		dOpts = append(dOpts, transport.WithTimeout(opts.DialTimeout))
+	}
+
+	c, err := r.pool.Get(address, dOpts...)
```
lpxxn, line 107: if you do this，the code will be useless.
https://github.com/micro/go-micro/blob/0d94784e72337b02b3bf163d72483e0b46c37d09/transport/http_transport.go#L128

asim: I'm not worried about that right now. Does this fix work for the client? If so we'll merge it.

lpxxn: fixed

PR #725, opened 2019-09-03
initial support for cuelang

xmlking:
This PR adds support for  [config.cue](https://cuelang.org/) encoding 

Work in progress...


asim:
Can we move this to go-plugins for the time being until Cue matures to a v1.

xmlking:
Sure. will do that.

PR #729, opened 2019-09-04
Tunnel session management and unicast/multicast

asim:
This PR adds session management and discovery via some new message types: discover, announce, open, session. It also adds a multicast flag to signify that a connection is to be sent across many links. Otherwise we will always pin a session to a single link.


tunnel/session.go
```diff
@@ -52,6 +63,10 @@ type message struct {
 	outbound bool
 	// loopback marks the message intended for loopback
 	loopback bool
+	// whether to send as multicast
+	multicast bool
```
milosgajdos, line 67: Should we turn `{multi,broad,uni}cast`s into some enum type rather than keeping them in separate fields?

tunnel/tunnel.go
```diff
@@ -42,6 +45,12 @@ type Session interface {
 	transport.Socket
 }
 
+var (
```
milosgajdos: Would be cool to have this on top of the src file like we talked about a while ago to be consistent with other source files and Go std library. Also, if we could add some comments here that would be dope.

tunnel/options.go
```diff
@@ -29,6 +31,15 @@ type Options struct {
 	Transport transport.Transport
 }
 
+type DialOption func(*DialOptions)
+
+type DialOptions struct {
```
milosgajdos, line 36: If you decide to turn the `cast`* options into some enum, this would have to be changed 🤔

tunnel/link.go
```diff
@@ -27,11 +28,67 @@ type link struct {
...
+
+			l.RLock()
+			for ch, t := range l.channels {
+				if d := time.Since(t); d > killTime {
+					kill = append(kill, ch)
+				}
+			}
+			l.RUnlock()
+
+			// if nothing to kill don't both with a wasted lock
```
milosgajdos: should this be "dont bothER" as opposed to just "dont both". Typo?

tunnel/listener.go
```diff
@@ -17,26 +18,81 @@ type tunListener struct {
...
+	// del func to kill listener
+	delFunc func()
+}
+
+// periodically announce self
+func (t *tunListener) announce() {
+	tick := time.NewTicker(time.Minute)
+	defer tick.Stop()
+
+	announce := func() {
```
milosgajdos: Wouldnt turning this `func` definition into a method be "nicer"? If defined as method it's defined once and possibly testable should we decide to start writing tests at some point?

milosgajdos (commented):
Left some comments. But I'll have to read through the PR again to make sure I understand it properly.

tunnel/link.go
asim: Typo. Fixed.

tunnel/listener.go
asim: Moved into session as an Announce method.

tunnel/tunnel.go
asim: Done.

tunnel/session.go
asim: In the future potentially but I've done it this way because I want the ability to internally broadcast without allowing the caller to do so. IPv6 has removed broadcast for appropriate reasons. Its probably best if we avoid it to considering the scale the network could get to.

tunnel/options.go
asim: Related to the other point. There should probably be a cleaner way but depends on whether we support multiple messaging types.

PR #734, opened 2019-09-05
Update tunnel to send discovery on connect and multicast messages. An…

asim:
This PR includes a number of updates to try help convergence in tunnelling.

1. Send discover message to all links once connected (todo: send on reconnect or new link setup)
2. The announce returns all channels at once when the first discover is sent
3. We sent a discover message on multicast Dial (this however can be blocking startup speed)
4. Reduce announce time for channels to 30 seconds
5. Broadcast the announce for the listener rather than just directly back to the link

TODO
1. Send discover on reconnect or new link setup. We only do this in tunnel.Connect
2. Send channels from the second node more promptly after the connect
3. Ensure we're not blasting information to everyone in a flooded way

### DIFF
--- a/tunnel/default.go
+++ b/tunnel/default.go
@@ -94,6 +94,21 @@ func (t *tun) delSession(channel, session string) {
 	t.Unlock()
 }
 
+// listChannels returns a list of listening channels
+func (t *tun) listChannels() []string {
+	t.RLock()
+	defer t.RUnlock()
+
+	var channels []string
+	for _, session := range t.sessions {
+		if session.session != "listener" {
+			continue
+		}
+		channels = append(channels, session.channel)
+	}
+	return channels
+}
+
 // newSession creates a new session and saves it
 func (t *tun) newSession(channel, sessionId string) (*session, bool) {
 	// new session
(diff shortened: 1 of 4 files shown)


tunnel/default.go
```diff
@@ -463,22 +487,46 @@ func (t *tun) listen(link *link) {
...
-						"Micro-Tunnel-Channel": channel,
-						"Micro-Tunnel-Session": sessionId,
-						"Micro-Tunnel-Link":    link.id,
-						"Micro-Tunnel-Token":   t.token,
-					},
-				})
 			}
+
+			// send back the announcement
+			link.Send(msg)
```
milosgajdos: Can this fail? If yes, shouldnt we log the error?

tunnel/default.go
```diff
@@ -891,17 +949,45 @@ func (t *tun) Dial(channel string, opts ...DialOption) (Session, error) {
...
+			// shorten this
+			dialTimeout = time.Millisecond * 500
+		}
+
 		// wait for announce
 		select {
 		case msg := <-c.recv:
 			if msg.typ != "announce" {
-				return nil, errors.New("failed to discover channel")
+				err = errors.New("failed to discover channel")
```
milosgajdos: Should we turn this into error definition i.e. `ErrDiscoverChan` or something like that?

PR #736, opened 2019-09-05
Solicit routes when new node is discovered

milosgajdos:
This PR introduces the following changes:

* When receiving `connect` message we now send back `neighbour` message and `advert` with all the routes.
* When a new neighbour is detected we send a new `solicit` message to ask for routes. The other side flushes all the routes and sends an `advert` message with all routes included in it
* `router` interface now contains `Solicit()` message which flushes all the routes in the routing table and advertise them


router/router.go
```diff
@@ -28,6 +28,8 @@ type Router interface {
 	Advertise() (<-chan *Advert, error)
 	// Process processes incoming adverts
 	Process(*Advert) error
+	// Solicit advertises the whole routing table ot the network
```
asim: Minor typo "ot" => "to"

PR #750, opened 2019-09-10
[WIP] Neighbour is now Peer. Peer is a node in network topology.

milosgajdos:
This is a chunky PR that brings the following changes:

* `neighbour` is now `peer`
* `proto` files have been changed to accommodate ☝️change
* `node` has now been separated into dedicated source file (`node.go`) as it started bloating `network.go` and it has been getting increasingly harder to navigate the `network.go` file
* we now have tests for `node`
* `Nodes()` has been removed from `network.Network` interface
* `Peers()` has been added to `network.Node`; it returns all the nodes in the node neighbourhood up to `network.MaxDepth`
* `Neighbourhood()` is now `Peers()` ☺️
* Some race conditions have been fixed
* `handler.Neighbourhood()` is now `handler.ListPeers()` ☺️

🚨**DO NOT MERGE THIS JUST YET! TESTING IN PROGRESS!** 🚨


network/proto/network.proto
```diff
@@ -6,32 +6,46 @@ import "github.com/micro/go-micro/router/proto/router.proto";
 
 // Network service is usesd to gain visibility into networks
 service Network {
-        rpc ListRoutes(go.micro.router.Request) returns (go.micro.router.ListResponse) {};
         rpc ListNodes(ListRequest) returns (ListResponse) {};
-        rpc Neighbourhood(NeighbourhoodRequest) returns (NeighbourhoodResponse) {};
+        rpc ListPeers(PeerRequest) returns (PeerResponse) {};
```
asim, line 9: I'm assuming Neighbourhood is being replaced by Topology. And now we also have ListPeers which seems to make ListNodes redundant.

Maybe we should remove ListNodes and simplify as:

Network.{Peers, Routes, Topology}
- Peers - return list of peers
- Routes - return list of routes
- Topology - return the whole known graph

Seems to make it more consistent?

network/proto/network.proto
```diff
@@ -49,16 +63,16 @@ message Close {
         Node node = 1;
 }
 
-// Solicit is sent when requesting route advertisement from the network nodes
+// Solicit is sent when soliciting routes from the network nodes
 message Solicit {
```
asim: Only just noticing this is in the network.proto. Is this because we receive both a peer and advert message on solicit? If its not the case then this should surely be in the router.proto and fired over the control channel?

network/network.go
```diff
@@ -46,6 +46,8 @@ type Network interface {
 	Connect() error
 	// Nodes returns list of network nodes
 	Nodes() []Node
```
asim: If we eliminate the ListNodes endpoint we no longer need Nodes()

network/network.go
```diff
@@ -46,6 +46,8 @@ type Network interface {
...
+	// Topology returns node topology up to MaxDepth hops
+	Topology() Node
```
asim: Does this not seem to basically return Node which is the existing Node embedded in this Network? Should we maybe just return the entire topology in Peers and remove the Topology() method on the interface?

network/proto/network.proto
milosgajdos: Yes, `Topology()` does replace `Neighbourhood()`

We can remove `ListNodes`, of course. Probably makes a lot of sense.

One thing I'm not sure now regarding your comment, though, is: are you suggesting `Topology()` should return the **whole known graph** as opposed to the originally intended graph up to `MaxDepth=3`?

network/network.go
milosgajdos: Yes and No. 

Yes it does return the network `Node`
No, it returns the network `Node` [graph] that contains all the reachable nodes up to 3 hops.

`Peers()` returns **adjacent nodes only**. `Peers()` returns the simplest possible *graph/topology* starting from network node containing only the peers i.e. immediate neighbours and nothing else.

network/proto/network.proto
milosgajdos: Looking at the `processCtrlChan()` I'm thinking it makes more sense to move the `Solicit` message into `router.proto`, actually 🤔

Right now it contains `Node` (so we can track what node sends it) we'd have to change it to something like:
```
message Solicit {
   string id = 1;
}
```

Not a biggie...

network/proto/network.proto
asim: Bad wording. Yes just max depth 3. We're limiting to the depth to 3 on processing peer messages so the "whole known graph" is basically that.

network/network.go
asim: Lets return the known topology (max depth 3) here in Peers() and remove Nodes()/Topology() methods in the interface.

network/proto/network.proto
asim: Yea think it makes sense to move it to router.proto and use in control channel if its related to returning routes

network/network.go
milosgajdos: I see what you mean here, but won't this overload the term `Peers()`?

network/proto/network.proto
milosgajdos: Working on this now.

network/proto/network.proto
milosgajdos: Below you mentioned you want to return topology (known graph up to 3 hops) in `Peers()` in `network.Network` and chuck `Nodes()/Topology()`. Does that mean we eliminate `handler.Topoly`, too? I'm assuming it does?

network/network.go
asim: It depends on the expectation. If Peers() returns a list of nodes and each node has a list of Peers then you probably expect that you can continue to walk that as far as it goes. I just think limiting the surface area of the API early on is better because its going to be harder to claw it back later.

network/proto/network.proto
milosgajdos: That original logic was broken and inefficient as it was working around the inaccessibility of the full graph it was doing this:
* [get the whole graph i.e. all nodes](https://github.com/micro/go-micro/blob/master/network/handler/handler.go#L73)
* [search through all nodes](https://github.com/micro/go-micro/blob/master/network/handler/handler.go#L78)

That's pretty bad and the only reason for at the time was the lack of "access" to the network graph; the only access to the network graph we had was via the `Nodes()` method of the `network.Node` which was returning a list of nodes in the whole network graph which we could do some processing on. Hence, we traversed the whole graph, collected all the nodes and built subgraph by searching.

In other words: the original solution you keep mentioning was really just a "placeholder" workaround until we would figure out proper node/network interface. Now, if we scrap `Nodes()`, too I'm wondering how to build up the graph etc...that's the point I was trying to explain to you yesterday.

network/network.go
milosgajdos: You expect but at the moment `Peers()` returns literary just a list of immediate neighbours like we originally agreed. Not a graph. A list of immediate neighbours. Nothing else. If you want to replace the `Peers()` logic with what the `Topology()` does now and then scrap the `Topology` then we can do that, but then I feel we're overloading term `Peers` or something. Im ok to do this if you want me to.

network/default.go
```diff
@@ -647,41 +539,23 @@ func (n *network) processCtrlChan(client transport.Client, listener tunnel.Liste
...
-					}
-					// if advertising node id is not the same as Route.Router
 					// we know the advertising node is not the origin of the route
-					if advertNode.id != event.Route.Router {
-						// if the origin router is not in the advertising node neighbourhood
+					if pbRtrAdvert.Id != event.Route.Router {
+						// if the origin router is not the advertising node peer
 						// we can't rule out potential routing loops so we bail here
-						if _, ok := advertNode.neighbours[event.Route.Router]; !ok {
+						if !advertNode.HasPeer(event.Route.Router) {
```
asim: This will throw away routes that are more than one hop away.

If the chain is (A - B - C - D) then A will throw away routes from B that were originated at D. It will only accept routes from B-C and nothing else. The peer graph must be walked to see if B and D are connected. While continuing to skip routes for our own ID to stop loops.

`HasPeer` should walk its graph or use `GetPeerNode` instead.

asim:
```diff --git a/network/default.go b/network/default.go
index c981f1d..605fa01 100644
--- a/network/default.go
+++ b/network/default.go
@@ -554,7 +554,7 @@ func (n *network) processCtrlChan(client transport.Client, listener tunnel.Liste
                                        if pbRtrAdvert.Id != event.Route.Router {
                                                // if the origin router is not the advertising node peer
                                                // we can't rule out potential routing loops so we bail here
-                                               if !advertNode.HasPeer(event.Route.Router) {
+                                               if peer := advertNode.GetPeerNode(event.Route.Router); peer == nil {
                                                        log.Debugf("Network skipping advert message from peer: %s", pbRtrAdvert.Id)
                                                        continue
                                                }
```

PR #761, opened 2019-09-16
Delete dead peer [gateway] routes

milosgajdos:
We currently only prune routes of dead nodes which were originated by the dead nodes i.e. the dead nodes were the original routers of the [propagated] routes.

We need to prune the routes which route via the dead nodes, too i.e. the routes which have the dead nodes set as a gateway.

This PR addresses that.

Furthermore we prune the nodes in the whole node topology as opposed to just the direct peers,
otherwise we will keep on advertising dead peer(s) (they might be "buried" inside the peer topology as a neighbours of the alive peers).

Finally some cleanup of `node` and a bit more consistency and tests.

### DIFF
--- a/network/default.go
+++ b/network/default.go
@@ -427,7 +427,7 @@ func (n *network) prunePeerRoutes(peer *node) error {
 
 	// lookup all routes routable via gw
 	q = router.NewQuery(
-		router.QueryGateway(peer.address),
+		router.QueryGateway(peer.id),
 	)
 	if err := n.pruneRoutes(q); err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


network/default.go
```diff
@@ -416,8 +415,29 @@ func (n *network) pruneNode(id string) error {
...
+	q := router.NewQuery(
+		router.QueryRouter(peer.id),
+	)
+	if err := n.pruneRoutes(q); err != nil {
+		return err
+	}
+
+	// lookup all routes routable via gw
+	q = router.NewQuery(
+		router.QueryGateway(peer.address),
```
asim: If we set Gateway to peer.id we should remember to change this

PR #780, opened 2019-09-24
[WIP] Registry event

milosgajdos:
Adds `registry.Event` types (Go types and protobufs)


registry/proto/registry.proto
```diff
@@ -71,3 +71,22 @@ message WatchRequest {
...
+// Event is registry event
+message Event {
+        // Event Id
+        string id = 1;
+        // type of event
+        EventType type = 2;
+        // unix timestamp of event
+        int64 timestamp = 3;
+        // service entry
+	Service service = 4;
```
asim: formatting issue with tabs?

milosgajdos: FML

PR #790, opened 2019-09-25
[WIP] Memory registry TTL expiry

milosgajdos:
**This is a WIP DO NOT MERGE YET**

This PR allows to specify memory registry TTL and expires the registry records that exceed the TTL


registry/service/service.go
```diff
@@ -58,8 +58,13 @@ func (s *serviceRegistry) Register(srv *registry.Service, opts ...registry.Regis
 		o(&options)
 	}
 
+	ctx := context.Background()
+	if options.TTL.Nanoseconds() != 0.0 {
+		ctx = context.WithValue(ctx, "register_ttl", options.TTL)
```
asim: This won't actually do anything, We need to either pass TTL in the proto as part of some sort of new options field or ignore entirely and assume a default TTL for the registry service.

asim: This can probably be removed

asim: Actually. This should be updated to send TTL in the options right?

registry/handler/handler.go
```diff
@@ -26,10 +27,15 @@ func (r *Registry) GetService(ctx context.Context, req *pb.GetRequest, rsp *pb.G
 }
 
 func (r *Registry) Register(ctx context.Context, req *pb.Service, rsp *pb.EmptyResponse) error {
-	err := r.Registry.Register(service.ToService(req))
+	var regOpts []registry.RegisterOption
+	ttl := time.Duration(req.Options.Ttl)
```
asim: Does this imply TTL will be sent as nanoseconds? We're using seconds everywhere else so we may want to process this as seconds.

registry/service/service.go
milosgajdos: Yeah, good spot! Will push the change shortly.

registry/handler/handler.go
milosgajdos: It should not matter in the grand scheme of things. Because 0.0 is 0.0 if either of nano or just seconds is used. But let's be consistent and change it so seconds 👍

asim: This will need to be `time.Duration(req.Options.Ttl) x time.Second` otherwise it'll parse as nanoseconds.

milosgajdos: Truth!

asim:
Closing in favour of the other PR

PR #800, opened 2019-09-29
fix rcache node overwrited issue

kordenlu:


vtolstov:
looks good, but can you drop from pr go.mod and go.sum changes?

asim:
Please change CopyOne to CopyService

milosgajdos:
RE: `go.mod` comment by @vtolstov 

I've seen something similar on my other workstation where I have `go 1.13` installed instead of `go 1.12`. Sometimes `1.13` "pulls in" the `reflect2` dependencies into `go.mod`. 

I have not looked properly why that happens but I did see that happening.

vtolstov:
@milosgajdos83 i know, but i think that we need to bump and change deps only that it really needed.
So each fix and new feature not update deps.
Also will be good to update deps only when feature it really needs or via separate commit.

milosgajdos:
Yeah, man. I agree with you.

I just thought I'd point out what could be the potential source of the `go.mod` change as I noticed the very same thing happening to me

PR #801, opened 2019-09-29
[WIP] Proof of concept for using map in memory registry

milosgajdos:
This is a Proof of Concept (PoC) which attempts to use `map` for storing service records in memory registry. Its functionality is now entirely on par with https://github.com/micro/go-micro/pull/790 but it provides the following advantages:

* **much less boiler plate code** -- we no longer have to do Go slice kung-fu; the code is now arguable more readable and therefore better maintainable; this also means we can get rid of bunch of helper code which is no longer needed
* **more extensive test coverage** -- we now properly check `Register` and `Deregister` results as well as `ListServices`
* **we no longer export internal memory registry fields** -- this has led to updating some tests in other packages which were not taking advantage of `Service()` memory registry option
* we squashed more bugs

if we decide to merge this PR we can close of https://github.com/micro/go-micro/pull/790


registry/memory/memory.go
```diff
@@ -74,91 +118,152 @@ func (m *Registry) Init(opts ...registry.Option) error {
...
+	var options registry.RegisterOptions
+	for _, o := range opts {
+		o(&options)
 	}
 
-	return service, nil
-}
+	// if no TTL has been set, set it to DefaultTTL
+	if options.TTL.Seconds() == 0.0 {
+		options.TTL = DefaultTTL
```
asim, line 157: If a TTL is not set we don't set it here because there's no guarantee the caller will call Register again to update the TTL.

registry/memory/memory.go
```diff
@@ -74,91 +118,152 @@ func (m *Registry) Init(opts ...registry.Option) error {
...
-	if service, ok := m.Services[s.Name]; !ok {
-		m.Services[s.Name] = []*registry.Service{s}
-		go m.sendEvent(&registry.Result{Action: "update", Service: s})
-	} else {
-		svcCount := len(service)
-		svcNodeCounts := make(map[string]map[string]int)
-		for _, s := range service {
-			if _, ok := svcNodeCounts[s.Name]; !ok {
-				svcNodeCounts[s.Name] = make(map[string]int)
+	if _, ok := m.records[s.Name]; ok {
```
asim, line 213: Should we do if !ok to return immediately so the rest of this is not nested under an if statement?

registry/memory/memory.go
```diff
@@ -74,91 +118,152 @@ func (m *Registry) Init(opts ...registry.Option) error {
...
-		} else {
-			m.Services[s.Name] = service
+	records, ok := m.records[name]
+	if !ok {
+		return nil, registry.ErrNotFound
+	}
+
+	services := make([]*registry.Service, len(m.records[name]))
+	i := 0
+	for _, record := range records {
```
asim, line 247: Should you not just be able to do `for i, record` and get the index integer for free rather than incrementing a counter?

registry/memory/memory.go
```diff
@@ -175,11 +280,12 @@ func (m *Registry) Watch(opts ...registry.WatchOption) (registry.Watcher, error)
...
-	m.Watchers[w.id] = w
+	m.watchers[w.id] = w
 	m.Unlock()
+
 	return w, nil
 }
 
 func (m *Registry) String() string {
-	return "memory"
+	return "memorynew"
```
asim: This should probably just stay as "memory"

registry/proto/registry.proto
```diff
@@ -17,6 +17,7 @@ message Service {
 	map<string,string> metadata = 3;
 	repeated Endpoint endpoints = 4;
 	repeated Node nodes = 5;
+        Options options = 6;
```
asim, line 20: Formatting issue?

registry/memory/memory.go
milosgajdos: Nope, because `records` is a map

`m.records` is your favourite data structure: map of maps; `records` here is a particular instance of the map of maps.

registry/memory/memory.go
milosgajdos: Nope, because in the nested `if` we gradually delete records from the map (and ths return memory to heap) if they're no longer needed:
1. node
2. service version record
3. the whole service record

registry/memory/memory.go
milosgajdos: what is the DefaultTTL then? If the client does not set it (unless it sets it be default) then TTL is set to 0 and it will be pruned by `m.registry.ttlPrune()`

milosgajdos:
Closing. This PR uses map of maps to store service records.

registry/memory/memory.go
asim: There is no default TTL. Records can be stored indefinitely. This is the behaviour of all other registries.

asim: If the TTL is not set you'd not try to prune it

milosgajdos:
Let's close this one in favour or https://github.com/micro/go-micro/pull/790 which is more in line with micro codebase.

PR #807, opened 2019-10-02
some spelling fixes in memory and gossip registry

vtolstov:


vtolstov:
also i think that it now too noisy with Debugf, why we need message for refreshing ttl ?

PR #810, opened 2019-10-02
First commit to add etcd registry support

milosgajdos:
This moves the `etcdv3` registry implementation from `go-plugins` to `go-micro` with slight (minimal) code modifications. This PR also updates the `go-micro` dependencies.

This PR has been mostly tested with `etcd 3.3.15`, but seems to be working with `etcd 3.4`, too. It might need few changes here and there.


asim:
Wanted to merge asap. We should make one more change to set prefix as /micro/registry so if we add anything else to etcd it's all scoped under /micro

PR #829, opened 2019-10-08
Limit the number of outbound connections to MaxConnections

milosgajdos:
This commit also fixes control channel shenanigans:
- recover error in control channel `Accept()`

### DIFF
--- a/network/default.go
+++ b/network/default.go
@@ -29,6 +29,8 @@ var (
 	ControlChannel = "control"
 	// DefaultLink is default network link
 	DefaultLink = "network"
+	// MaxCconnections is the max number of network client connections
+	MaxCconnections = 3
 )
 
 var (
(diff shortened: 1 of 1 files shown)


asim:
This is a bit of a hammer. Backbone nodes will basically only connect to local nodes in one cloud provider since they'll resolve via the registry first and retrieve 3 nodes and then after append the peers provided via option but ignore them.

- We want the peer list via option to be used no matter what. So when we specify --nodes it uses all these addresses. 
- We want to limit number of nodes we use via the resolver so clients using the http resolver only make 3 connecfions. They will not specify --nodes so this works well.

So max connections should be applied to nodes resolved via the resolver only.

milosgajdos:
Via the `network.Resolver` you mean?

asim:
Yes you're right on that overloading part. The question is where we optimise that. Google rotates IPs returned by DNS records. We could do the same for the network api used by the http resolver but we could also internally do some level of randomisation. Have a feeling randomness is probably going to yield simila R results to anything we try cleverly do.

Also yes on the network.Resolver

PR #830, opened 2019-10-08
Start abstracting away the ACME provider

jakexks:
... in preparation for global certificates issued with the DNS challenge.

Does this make sense?


api/server/acme/acme.go
```diff
@@ -0,0 +1,50 @@
...
+const (
+	// LibAutoCert is the acme library from golang.org/x/crypto/acme/autocert
+	LibAutoCert = "autocert"
+	// LibCertMagic is the acme library from github.com/mholt/certmagic
+	LibCertMagic = "certmagic"
+)
+
+// Library is an provider interface
+type Library interface {
+	NewListener([]string) (net.Listener, error)
```
asim: NewListener([]string) => NewListener(...string)

This would allow calling new listener without any hosts so it operates on any hostname. Also cleaner with call to NewListener() or NewListener("foo", "bar")

api/server/acme/acme.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// AutoCert is the ACME provider from golang.org/x/crypto/acme/autocert
+type AutoCert struct{}
```
asim: I would put the implementations under a sub directory for each

acme/autocert
acme/certmagic

And provide `type func New() acme.Acme` in each. We'd then accept the Acme interface as an option to a server and flag flip or switch between providers in the micro toolkit API and web packages.

PR #833, opened 2019-10-09
Add dialoptions and calloptions

orbli:
Adding dial options and call options to `go-micro/client/grpc`

Major changes:
1. Adding `go-micro/client.option` converter at `client/grpc/options.go`
2. Adding options behaviour to `go-micro/client/grpc.grpcClient.call` and `go-micro/client/grpc.grpcClient.stream` to implement the dial options and call options

Questions for it to be used:
1. Would it affect behaviour of go-micro dialoptions / calloptions usage?
2. Not yet tested. Does it work?

### DIFF
--- a/client/grpc/options.go
+++ b/client/grpc/options.go
@@ -79,8 +79,8 @@ func MaxSendMsgSize(s int) client.Option {
 //
 // DialOptions to be used to configure gRPC dial options
 //
-func DialOptions(opts ...grpc.DialOption) client.Option {
-	return func(o *client.Options) {
+func DialOptions(opts ...grpc.DialOption) client.CallOption {
+	return func(o *client.CallOptions) {
 		if o.Context == nil {
 			o.Context = context.Background()
 		}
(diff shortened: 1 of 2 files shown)


client/grpc/options.go
```diff
@@ -72,3 +75,27 @@ func MaxSendMsgSize(s int) client.Option {
 		o.Context = context.WithValue(o.Context, maxSendMsgSizeKey{}, s)
 	}
 }
+
+//
+// DialOptions to be used to configure gRPC dial options
+//
+func DialOptions(opts ...grpc.DialOption) client.Option {
```
asim: This should likely return a client.CallOption

client/grpc/options.go
```diff
@@ -72,3 +75,27 @@ func MaxSendMsgSize(s int) client.Option {
...
+			o.Context = context.Background()
+		}
+		o.Context = context.WithValue(o.Context, grpcDialOptions{}, opts)
+	}
+}
+
+//
+// CallOptions to be used to configure gRPC call options
+//
+func CallOptions(opts ...grpc.CallOption) client.Option {
```
asim: This should likely return a client.CallOption

asim:
This will likely continue to work but I'd ask you to test it and also switch the options to return client.CallOption since these are likely things users want to do at call time as grpc specifies.

client/grpc/options.go
orbli: Solved at https://github.com/micro/go-micro/pull/833/commits/77108771db383631df49ef08732eb5159b85157c#diff-4c65aaabc8f3a5b0c00b82ff6d39a485R82

client/grpc/options.go
orbli: Solved at https://github.com/micro/go-micro/pull/833/commits/77108771db383631df49ef08732eb5159b85157c#diff-4c65aaabc8f3a5b0c00b82ff6d39a485R94

lcken: Here don't work, because the `client.Option` apply on the copy of `grpcClient.opts.CallOptions.Context`, and `g.getGrpcDialOptions()` get `grpc.CallOption` from origin `grpcClient.opts.CallOptions.Context`

PR #835, opened 2019-10-09
Router strategy

milosgajdos:
This PR adds a router advertising strategy option:
- `All` advertises all routes
- `Optimal` advertises optimal routes only

Optimal routes are a collapsed list of routes per service:
- prefer routes with lowest metric
- when routes with equal metric are found, prefer to advertise your route i.e. the route that originates at your router.

### DIFF
--- a/router/default.go
+++ b/router/default.go
@@ -682,7 +682,7 @@ func (r *router) flushRouteEvents(evType EventType) ([]*Event, error) {
 		return nil, fmt.Errorf("failed listing routes: %s", err)
 	}
 
-	if r.options.Advertise == All {
+	if r.options.Advertise == AdvertiseAll {
 		// build a list of events to advertise
 		events := make([]*Event, len(routes))
 		for i, route := range routes {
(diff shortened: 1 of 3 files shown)


router/router.go
```diff
@@ -139,6 +139,28 @@ type Advert struct {
 	Events []*Event
 }
 
+// Strategy is route advertisement strategy
+type Strategy int
+
+const (
+	// All advertises all routes to the network
+	All Strategy = iota
```
asim: Might want to call these AdvertiseAll/AdvertiseOptimal so when they're used they look like:

```
router.AdvertiseAll
router.AdvertiseOptimal
```

router/router.go
```diff
@@ -139,6 +139,28 @@ type Advert struct {
...
+	// Optimal advertises optimal routes to the network
+	Optimal
```
asim: Let's go with AdvertiseBest

router/default.go
```diff
@@ -692,15 +682,55 @@ func (r *router) flushRouteEvents(evType EventType) ([]*Event, error) {
...
+	}
+
+	// routeMap stores optimal routes per service
+	optimalRoutes := make(map[string]Route)
+
+	// go through all routes found in the routing table and collapse them to optimal routes
+	for _, route := range routes {
+		optimal, ok := optimalRoutes[route.Service]
+		if !ok {
+			optimalRoutes[route.Service] = route
```
asim: This should probably be service+network as we'll likely provide isolated networks

PR #856, opened 2019-10-15
Cloudflare Store implementation for workers KV

jakexks:
As the cloudflare go library doesn't expose expiring keys, this isn't complete yet.


vtolstov:
I think that pr needs to be for go-plugins repo

jakexks:
Fair comment. It's a PR against go-micro as we are about to use it, and @asim doesn't want to pull go-plugins in the main build. However, it might be good to make everything a plugin and have a selective build.

vtolstov:
This is bad thing. We again bloat go.mod deps for go-micro.

asim:
Can you offer a better solution? We otherwise have to "download the world" by using go-plugins.

vtolstov:
Better split it to logical parts like wrappers store, brokers and so.

PR #876, opened 2019-10-22
Peerlink route metric

milosgajdos:
WIP for reading tunnel link metrics in network.

I'm opening a PR so we have somewhere to discuss the code as we work through this.

### DIFF
--- a/network/default.go
+++ b/network/default.go
@@ -700,7 +700,7 @@ func (n *network) getRouteMetric(router string, gateway string, link string) int
 	defer n.RUnlock()
 
 	if link == "local" && gateway == "" {
-		log.Debugf("Network link: %s, gateway: blank", link, gateway)
+		log.Debugf("Network link: %s, gateway: blank", link)
 		return 1
 	}
 
--- a/network/default.go
+++ b/network/default.go
@@ -859,7 +859,7 @@ func (n *network) advertise(advertChan <-chan *router.Advert) {
 					address = fmt.Sprintf("%d", hasher.Sum64())
 				}
 				// calculate route metric to advertise
-				metric := n.getRouteMetric(event.Route.Router, event.Route.Gateway, DefaultLink)
+				metric := n.getRouteMetric(event.Route.Router, event.Route.Gateway, event.Route.Link)
 				// NOTE: we override Gateway, Link and Address here
 				route := &pbRtr.Route{
 					Service: event.Route.Service,
--- a/network/default.go
+++ b/network/default.go
@@ -369,6 +369,7 @@ func (n *network) processNetChan(listener tunnel.Listener) {
 				if err := n.updatePeerLinks(pbNetConnect.Node.Address, m.session.Link()); err != nil {
 					log.Debugf("Network failed updating peer links: %s", err)
 				}
+				// add peer to the list of node peers
 				if err := n.node.AddPeer(peer); err == ErrPeerExists {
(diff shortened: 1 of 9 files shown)


network/network.go
```diff
@@ -54,6 +56,14 @@ type Network interface {
 	Server() server.Server
 }
 
+// Message is network message
+type Message struct {
+	// msg is transport message
+	msg *transport.Message
```
milosgajdos, line 62: Should export `msg`

network/default.go
```diff
@@ -681,14 +706,20 @@ func (n *network) processCtrlChan(listener tunnel.Listener) {
...
-						continue
+					n.RLock()
+					if link, ok := n.peerLinks[event.Route.Gateway]; ok {
+						// NOTE: should we change router.Route.Metric to int64?
+						if int(link.Length()) < route.Metric {
+							route.Metric = int(link.Length())
+						}
 					}
+					n.RUnlock()
+					// TODO: Are we dropping any routes?
```
milosgajdos: We need to come up with some way of dropping some routes. Otherwise we might end up with more routes than before.

network/network.go
asim: Do not export for now since it's not externally useful

network/default.go
asim: We're discarding routes if they're not in the peer graph right? So basically we might just want to recheck that it's not above the hop count of 3/4. 

Otherwise a route is probably not useful if the roundtrip is more than 5 seconds. So we may want to set the upper limit of the metric to that.

network/default.go
```diff
@@ -681,14 +706,20 @@ func (n *network) processCtrlChan(listener tunnel.Listener) {
...
-					n.node.RUnlock()
-					// throw away metric bigger than 1000
-					if route.Metric > 1000 {
-						log.Debugf("Network route metric %d dropping node: %s", route.Metric, route.Router)
-						continue
+					n.RLock()
+					if link, ok := n.peerLinks[event.Route.Gateway]; ok {
+						// NOTE: should we change router.Route.Metric to int64?
+						if int(link.Length()) < route.Metric {
+							route.Metric = int(link.Length())
```
asim: Should we be doing more here besides just using link length? Maybe we want metric to still be a number between 1 and 1000 so it's readable. Also you want to check if length is 0 because it means it has not been measured yet.

milosgajdos: Yes, we totally should! This is a placeholder for the time being so we have a place to discuss.

asim:
```
if delay == 0 {
  delay = 1
}

if hop == 0 {
  hop = 1
}

if length == 0 {
  length = 10e10
}

metric := (delay * length * hop) / 10e9

route.Metric += metric
```

network/default.go
asim: We have agreed to just leave this for now. Since we discard routes beyond the peer graph at maxdepth 3 that creates a level of pruning. Once we have a better understanding of our new metric value then we'll decide how to proceed. This may require some moving average e.g if all things slow down the boundary is always moving higher.

We must also keep in mind Starlink!

network/network.go
```diff
@@ -54,6 +56,14 @@ type Network interface {
 	Server() server.Server
 }
 
+// Message is network message
+type Message struct {
```
asim: this should not be exported since its not used outside this package

PR #881, opened 2019-10-24
Mark the session as accepted and discovered when receiving messages

milosgajdos:
install [parallel](https://www.gnu.org/software/parallel/) tool. Please build against the latest `micro`.

Run `greeter` service

```
for i in `seq 1 100000`; do echo $i >> list.out; done
cat list.out | parallel -j 4 script.sh {}
```

Script (make sure you modify the service names appropriately):
```shell
echo "RPC request $1"
MICRO_PROXY_ADDRESS=localhost:64522 ./micro call go.micro.srv.greeting Say.Hello '{"name": "john"}'
```

### DIFF
--- a/tunnel/default.go
+++ b/tunnel/default.go
@@ -524,12 +524,17 @@ func (t *tun) listen(link *link) {
 			s, exists := t.getSession(channel, sessionId)
 			// we don't need this
 			if exists && s.mode > Unicast {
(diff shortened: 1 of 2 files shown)


milosgajdos:
Closing this PR as the tunnel and session interaction is more complex than what this PR intended to fix.

PR #883, opened 2019-10-25
PatchContext method added

xmlking:
adding PatchContext - this will create new context with original + patch metadata


asim:
Was this not going to be called Append?

xmlking:
I can change the name, but since `append` name may miss-lead  some users thinking , it only  add given metadata, it actually will overwrite existing metadata if same keys exist. So I thought `patch` will be appropriate name. Let me know if you prefer **append**

asim:
Update then would be better? Patch does not sound right for this context

xmlking:
would you like me add overwrite argument ? 
`AppendContext(ctx context.Context, patchMd Metadata, overwrite bool) ` 

`overwrite defaults to false`

Issue #878, opened 2019-10-23
How to detect Transport timeouts in (RPC) client?


veger:
I am using the (default) HTTP transport on an RPC client (so rpc_codec.go is used).

I want to timeout receiving messages as it locks the client mutex, so I cannot send (I need/use bidirectional stream), which works. After a while an error is returned (instead of indefinitely blocking the `Recv()` call like before).

Unfortunately it is impossible to properly detect there was a timeout: [rpc_coded.go#227](https://github.com/micro/go-micro/blob/master/client/rpc_codec.go#L227) converts the transport (timeout) error into a string and puts it in `Error.Details`!
So I would need to parse the string...?!

I don't think this is very nice, so I wonder if I am using the timeout (detection) wrongly?

veger:
I guess I am circumventing #532, a proper fix would not require to detect the Timeouts (as I do not want/need the timeouts on `Recv()` at all)

asim:
If there's a timeout the context is cancelled. You can check context.Done()

veger:
There is no context in [`Recv()`](https://github.com/micro/go-micro/blob/master/client/rpc_stream.go#L75)
And [HTTPTransport](https://github.com/micro/go-micro/blob/master/transport/http_transport.go#L141) also does not seem to have/use contexts. It just sets deadline and read response

```
	if h.ht.opts.Timeout > time.Duration(0) {
		h.conn.SetDeadline(time.Now().Add(h.ht.opts.Timeout))
	}

	rsp, err := http.ReadResponse(h.buff, r)
	if err != nil {
		return err
	}
```

But the whole point of me trying to set timeout is to be able to send/receive on the same RPC stream, which is not possible due to the Mutex in `Send()` and `Recv()`...

veger:
For us this issue is obsolete, we needed to detect the Timeouts to fake 'true bidirectional streaming', implemented by #884.
So feel free to close (unless you are interested to providing ways to detect these Timeouts as well)

vtolstov:
@asim so now close the issue as we now have grpc by default?

PR #884, opened 2019-10-25
RPC stream client/server mutex fix

veger:
As described in #878 (and #532) it is not possible to send messages while waiting to receive them as well (or vice-versa), due to RPC stream client and server being locked by both `Send()` and `Recv()` functions.

In order to solve this issue, I unlock the Mutex before calling the `codec` to receive data, and locking it again after `codec` returned.
This way we still have the Mutex protection of the client/server structs, but we can send/receive at same time.

:warning: Note that this assumes that `codec` does not need protection while being active.

I did not update `Send()` because I assumed that it won't block. By only fixing `Recv()` the `Send()` worked as welll. Maybe depending on transport/codec it can block as well?

I only considered these 2 components/files, as we are (currently) not having issues in other parts of the go-micro framework regarding to this. If needed I can update other parts as well to stay in sync with this change.

I have tested this in our micro-service environment and it all seems to work fine. I did not trigger race conditions or other panics. But I did not yet perform (high) load or long running tests, which would make race conditions more likely to pop up (if there are any).

### DIFF
--- a/client/rpc_stream.go
+++ b/client/rpc_stream.go
@@ -130,11 +130,15 @@ func (r *rpcStream) Error() error {
 }
 
 func (r *rpcStream) Close() error {
+	r.RLock()
+
 	select {
 	case <-r.closed:
+		r.RUnlock()
 		return nil
 	default:
 		close(r.closed)
+		r.RUnlock()
 
 		// send the end of stream message
 		if r.sendEOS {
--- a/client/rpc_stream.go
+++ b/client/rpc_stream.go
@@ -83,7 +83,10 @@ func (r *rpcStream) Recv(msg interface{}) error {
 
 	var resp codec.Message
 
-	if err := r.codec.ReadHeader(&resp, codec.Response); err != nil {
+	r.Unlock()
+	err := r.codec.ReadHeader(&resp, codec.Response)
+	r.Lock()
+	if err != nil {
 		if err == io.EOF && !r.isClosed() {
 			r.err = io.ErrUnexpectedEOF
 			return io.ErrUnexpectedEOF
@@ -102,11 +105,17 @@ func (r *rpcStream) Recv(msg interface{}) error {
 		} else {
 			r.err = io.EOF
 		}
-		if err := r.codec.ReadBody(nil); err != nil {
+		r.Unlock()
+		err = r.codec.ReadBody(nil)
+		r.Lock()
+		if err != nil {
 			r.err = err
 		}
 	default:
-		if err := r.codec.ReadBody(msg); err != nil {
+		r.Unlock()
+		err = r.codec.ReadBody(msg)
+		r.Lock()
+		if err != nil {
 			r.err = err
 		}
 	}
--- a/server/rpc_stream.go
+++ b/server/rpc_stream.go
@@ -48,13 +48,13 @@ func (r *rpcStream) Send(msg interface{}) error {
 }
 
 func (r *rpcStream) Recv(msg interface{}) error {
-	r.Lock()
-	defer r.Unlock()
-
 	req := new(codec.Message)
 	req.Type = codec.Request
 
-	if err := r.codec.ReadHeader(req, req.Type); err != nil {
+	err := r.codec.ReadHeader(req, req.Type)
+	r.Lock()
+	defer r.Unlock()
+	if err != nil {
 		// discard body
 		r.codec.ReadBody(nil)
 		r.err = err
@@ -67,7 +67,9 @@ func (r *rpcStream) Recv(msg interface{}) error {
 		switch req.Error {
 		case lastStreamResponseError.Error():
 			// discard body
+			r.Unlock()
 			r.codec.ReadBody(nil)
+			r.Lock()
 			r.err = io.EOF
 			return io.EOF
 		default:
@@ -77,7 +79,10 @@ func (r *rpcStream) Recv(msg interface{}) error {
 
 	// we need to stay up to date with sequence numbers
 	r.id = req.Id
-	if err := r.codec.ReadBody(msg); err != nil {
+	r.Unlock()
+	err = r.codec.ReadBody(msg)
+	r.Lock()
+	if err != nil {
 		r.err = err
 		return err
 	}
--- a/server/rpc_stream_test.go
+++ b/server/rpc_stream_test.go
@@ -1,0 +1,133 @@
+package server
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"math/rand"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/micro/go-micro/codec/json"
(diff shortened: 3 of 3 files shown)


asim:
We'll need a test to validate this continues to work as expected without race conditions and continues to send and receive messages in the right sequence. We also want to see this tested in parallel and under load.

veger:
Do you have a proposal how to test this?
Only codec and extractor have some tests, server/client code is fully untested at the moment... (as far as I can see)

veger:
We are using this fix for quite a while already, and it seems to be working nicely in our cloud environments.

In addition we found a race condition in the `Close()` function:
```
	select {
	case <-r.closed:
		return nil
	default:
		close(r.closed)
```

There is a small gap were 2 go routines can call Close() and the channel is not yet closed, resulting in a panic that `r.closed` is closed twice...
The latest commit adds the Mutex to the `Close()` function as well to protect against this.

vtolstov:
if the question about how to test i tink that it will be easy:
write test that spawns two micro services that ping/pong messages and run it with go test -race flag
to check race conditions i thing that we need some concurrency like 100-1000 connections between of them.
I think that it is enough to test this changes.
Also i'm try to switch own stack to use grpc service and transport and check it too.

asim:
We can't accept this without actual tests which account for any race conditions in codec but also sequencing of requests. Removing the locking exposes an issue that can lead to such a problem. Its not enough to send and receive, you have to actually send and receive the messages in order.

vtolstov:
@veger do you have time to write tests and check the code ?

veger:
Not a lot, especially since there are no such tests yet I could base it on (or extend).

But I think I have some time today to figure out how to write them to test the 'race condition in codec' and 'sequence of requests'

asim:
Surely it's just a test use of the server? Start a server and execute an endpoint in parallel?

veger:
**About sequence test**

I wrote a simple Sequence test, but it generally does not make sense, as it write something 1000 and then read it reads it a 1000 times. So of course it is in same order (no concurrency and the whole Mutex is not required)

I could put it in some go routines, but then the order can vary depending on the golang scheduler (on both read and write side).

**About concurrency test**

I spawn 10 go routines to write data 10 times (with some random delay to have better chance to trigger race condition) and 10 go routines to read the data (not checking as I don't care about payload/order now)

No race conditions are detected with current test, but as you can see I switched to proto codec (which we use :wink: ).
The json codec uses a shared encoder/decoder internally which is triggering race conditions (which makes sense).

So the concurrency test shows the change works for the rpc_stream.go, but it trigger race conditions in other place(s).

**Conclusion**

If these kind of tests are what you are looking for, I am willing clean them up, add client/rpc_stream.go tests, include all codecs to the tests, and fix them.
But I rather know before I make these changes, to potentially prevent a lot of work getting thrown away.
If you like to see other/more tests, also let me know.

vtolstov:
thanks!

1) in Send may be good not r.Lock()/defer r.Unlock() from begining?
r.RLock() before r.IsClosed(), r.RUnlock() after setting r.err
r.RUnlock() after 

and so on ?

veger:
You mean for `client/rpc_stream.go#Send()`, as I cannot see the `IsCLosed()` in the `server/rpc_stream.go#Send()`?

veger:
The json codec cannot be fixed properly (to support both read and write as same time), as it has the (shared) `Conn` so it needs to block... Or the (buffered) Decoder/Encoder needs to be replaced with json.(Un)Marshall in order to fully prevent the race condition in this codec

veger:
The grpc codec also has issues: The `ReadHeader()` sets `c.ContentType` and the `ReadBody()` uses it. In between some other call to `ReadHeader()` can be called changing the content type (not very likely in a (proper) application, but it is possible).

To fix this the `ContentType`, needs to be stored in `Message` which needs to be returned to `ReadBody()` so it can (re)use it.

veger:
What is the best course to solve this in your opinions?

I see the following possibilities:
1. Fix all, which will require a lot of internal refactoring/fixing.
2. Create a new/separate 'real async' RPC stream client/server, to be used with limited codecs (the ones without race conditions)
3. Ignore this PR (and maybe add a note that it is not really async in the sense that it is only possible to send _or_ receive and the same time)

For us either option 1 or 2 have our preference (for option 3 it means we need to keep and maintain our fork...), quality-wise I think option 1 would be best... (maybe option 2 can grow into 1 over time?)

Please let me know, so I know how to fix issues for this PR.

vtolstov:
as i understand 2 means that after creating new client/server you add one by one fixed codecs ?

veger:
2 would require codecs that are deemed safe. The ones we fixed and checked for race conditions.
I guess using a (different) `interface` we can enforce this @ compile time?

vtolstov:
@asim what you can say and suggest?

asim:
I think @veger has a specific requirement that no one else has had an issue with yet. I don't really know what the best place is for this considering codecs are not thread safe and were never intended to be. I believe most of this pushes for an internal refactor rather than hacks.

So I'm inclined to say 1. is the best option and for the moment either maintaining your own client/server is probably the best approach until that happens. Remember that go-micro is pluggable and the intention was always for people to adopt the same interfaces but plug and play where necessary.

Ultimately my goal is to have the http transport be the default and defacto transport that operates on http 1.1, http2 and http3 in the future. I'd then like the ability to operate on any protocol in our client and server without having to maintain a separate implementation for grpc and other things. So wherever we can improve I think that makes sense.

Tbh if the codecs did not do any internal mutation it would be simpler but thats hard to get right at the moment.

I can take a look at this issue/code next month during the quiet period over christmas. I have a feeling I'll be able to produced the desired outcome with limited code changes but that might introduce locking in a more fine grained way.

vtolstov:
My own IMHO: We need to say that codecs are not thread safe, so who use it must provide locking.
So in case of go-micro - locking must be set on transport or server/client level. If we have locks inside codecs to i think that we bring up slowdown with mutex contention.

asim:
Well only alternative is to refactor codes to

```
type Codec interface {
    ReadHeader(b []byte) (Header, error)
    ReadBody(b []byte, v interface{}) error
    Write(v interface{}) ([]byte, error)
}
```

Or something like that

veger:
@asim your refactoring will help to be able to make the codecs 'lockable' form the outside yes. As they are now, it is not possible as reading and writing to the connection (and thus blocking) is done inside the codecs. All of them that share their state (buffer) are 'unlockable' from outside.

The protobuf codec is an example that does not share state (that is why I could use it for my test)

I'll wait until I hear from you again, hopefully with some good new, but otherwise I/we can give it another try (with some additional instructions or plans you guys have).

asim:
I am going to merge this code. We are moving to grpc by default which will radically improve performance but having this fix may also be useful where we're using it for transport agnostic services. Obviously there's still a refactor required but I don't have time to deal with that at the moment.

PR #888, opened 2019-10-27
[PoC] MemDB routing table

milosgajdos:
**DO NOT MERGE THIS IN**

This PR is just to demonstrate how to use [go-memdb](https://github.com/hashicorp/go-memdb) to store routes.

This might provide both huge performance benefits as well as rich and fast queries.

In order to make this happen, `route.Options` had `Table` options added making `router.Table` configurable. Due to this change other things had to be changed as well.

In order to test this you'll have to check out the `memdb` branch in [my fork of micro](https://github.com/milosgajdos83/micro/tree/memdb).


milosgajdos:
I'm closing this for now, but won't be deleting the branch so if the need comes up we can dig it out.

PR #895, opened 2019-10-30
[WIP] K8s update and runtime package changes

milosgajdos:
This PR contains a bunch of needed by https://github.com/micro/micro/pull/402

It's been open to start the discussion before we can merge it in.

It will also need to be properly tested, which has not quite been done yet.


asim:
As per comments from discussion

- Make Poller => Notifier interface 
```
type Notifier interface {
    Notify() (<-chan Event, error)
    Close() error
}
type Event struct {
    Type // create, update, etc
    Service // string?
}
```
- Move micro.mu/update polling code to micro/internal/update
  * which returns `update.NewNotifier() runtime.Notifier`
- Move flag flipping processing of `runtime` into go-micro/config/cmd or in micro/cmd/cmd.go
- Do not delete services when runtime.Stop() is called

config/cmd/cmd.go
```diff
@@ -221,13 +230,19 @@ var (
 		"quic":   quic.NewTransport,
 	}
 
+	DefaultRuntimes = map[string]func(...runtime.Option) runtime.Runtime{
+		"local": runtime.NewLocalRuntime,
+		"k8s":   runtime.NewK8sRuntime,
```
asim: - local => runtime.NewRuntime
- rename "k8s" to "kubernetes"
- have kubernetes runtime in its own runtime/kubernetes package

runtime/default.go
```diff
@@ -25,161 +22,38 @@ type runtime struct {
 	services map[string]*service
 }
 
-type service struct {
-	sync.RWMutex
+// NewLocalRuntime creates new local runtime and returns it
+func NewLocalRuntime(opts ...Option) Runtime {
```
asim: NewRuntime

runtime/default.go
```diff
@@ -204,19 +78,53 @@ func (r *runtime) run() {
...
-				log.Debugf("Runtime error starting %s: %v", service.Name, err)
+				log.Debugf("Runtime error starting service %s: %v", service.Name, err)
+			}
+		case event := <-events:
+			log.Debugf("Runtime received notification event: %v", event)
+			// NOTE: we only handle Update events for now
+			switch event.Type {
+			case Update:
+				// parse returned response to timestamp
+				buildTime, err := time.Parse(time.RFC3339, event.Version)
```
asim: Should we be parsing out the version or just compare to the old verson and update if its not the same? This assumes a certain version format which we can't guarantee here.

runtime/default.go
```diff
@@ -204,19 +78,53 @@ func (r *runtime) run() {
...
+				if err != nil {
+					log.Debugf("Runtime error parsing build time: %v", err)
+					continue
+				}
+				r.Lock()
+				for _, service := range r.services {
```
asim: Event now has a Service string, so we should probably only update the specific service the event is for. Unless the Service name is blank in which case I guess you could assume a blanket update across all things.

runtime/default.go
```diff
@@ -204,19 +78,53 @@ func (r *runtime) run() {
...
+						service.Version = event.Version
+					}
+				}
+				r.Unlock()
+			}
+		case <-r.closed:
+			log.Debugf("Runtime stopped. Attempting to stop all services.")
+			for name, service := range r.services {
+				// TODO: handle this error
+				if err := r.Delete(service.Service); err != nil {
```
asim: Let's not stop the services right now

runtime/kubernetes.go
```diff
@@ -0,0 +1,252 @@
+// package kubernetes implements kubernetes micro runtime
+package runtime
```
asim: Let's move the kubernetes implementation to a runtime/kubernetes package

runtime/kubernetes.go
```diff
@@ -0,0 +1,252 @@
...
+			// * create a deployment
+			// * expose a service
+			log.Debugf("Runtime starting service: %s", service.Name)
+		case event := <-events:
+			// NOTE: we only handle Update events for now
+			log.Debugf("Runtime received notification event: %v", event)
+			switch event.Type {
+			case Update:
+				// parse returned response to timestamp
+				buildTime, err := time.Parse(time.RFC3339, event.Version)
```
asim: Same comments as for local runtime

config/cmd/cmd.go
milosgajdos: `kubernetes` is effing long. Remember, these are cli flags. But, ok. Will change, boss.

milosgajdos: Can't put `kubernetes` in its own package because that will create import cycle:
Runtime requires `Option` to pass in `runtime.Notifier`, and `runtime.Service`.
Need to think about how to do this, but it's not an easy file move -- I would have done that already if it were.

runtime/default.go
milosgajdos: I'm not sure how this would work, actually. Right now we assume *newer* build = a build that was released after we had started the service.

If that's changed we need to figure out some version "math" handling version comparisons. Otherwise simple string equal might lead to downgrades or whatnot.

config/cmd/cmd.go
asim: Not sure why it would create an import cycle. Kubernetes package importing the go-micro/runtime should cause zero issues

asim: All our other things follow this pattern with no issues

runtime/default.go
milosgajdos: This is a local runtime.  Delete stops them. I understand why we dont want to do this in K8s runtime right now, but being able to stop multiple services when stopping local runtime seems reasonable. Otherwise we're gonna end up chasing child process PIDs manually?

config/cmd/cmd.go
milosgajdos: Ah yeah, I forgot the `NewRuntime` function in `runtime` package no longer does the switch on type of runtime -- that's moved to `config/cmd` now, so should be ok, actually.

config/cmd/cmd.go
```diff
@@ -231,8 +232,8 @@ var (
 	}
 
 	DefaultRuntimes = map[string]func(...runtime.Option) runtime.Runtime{
-		"local": runtime.NewLocalRuntime,
-		"k8s":   runtime.NewK8sRuntime,
+		"local":       runtime.NewRuntime,
+		"kubernetess": kubernetes.NewRuntime,
```
asim: kubernetess => kubernetes

milosgajdos: OH fml. Great catch!

runtime/default.go
asim: The processes are children of this runtime and since Go can't fork they'll all die. But also micro/cmd called runtime.Delete explicitly. I think the caller who creates should also delete.

runtime/default.go
asim: I guess we can revisit later. My thinking was just that versions may not be solely built timestamp formatted like docker.

runtime/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,271 @@
...
+}
+
+// Registers a service
+func (k *kubernetes) Create(s *runtime.Service, opts ...runtime.CreateOption) error {
+	k.Lock()
+	defer k.Unlock()
+
+	// TODO:
+	// * create service
+	// * create dpeloyment
```
asim: typo

runtime/service.go
```diff
@@ -0,0 +1,157 @@
...
+		Process: new(proc.Process),
+		Exec: &process.Executable{
+			Binary: &packager.Binary{
+				Name: s.Name,
+				Path: exec,
+			},
+			Env:  c.Env,
+			Args: args,
+		},
+		output: c.Output,
```
asim, line 67: May need to add `closed: make(chan bool)` here as it panicked in the old version ala what Jake said

milosgajdos: This will be resolved by rebase I just did..

runtime/kubernetes/kubernetes.go
milosgajdos: Meh, this is a comment. Happy to fix.

milosgajdos:
This should now annotate Deployments as follows:
```yaml
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  annotations:
    build: "2019-11-01T23:08:13Z"
    deployment.kubernetes.io/revision: "1"
...
```

PR #896, opened 2019-10-31
Preallocation of slices when known at the time of its creation

tegk:
The size of a slice can be specified at the time of creation when its known. This can have significant impact on performance when a lot of appending to slices is happening.

Consider this benchmark:

```
package main
import "testing"

func BenchmarkNoPreallocate(b *testing.B) {
  b.ReportAllocs()
	e := make([]int64, 10, 10)
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		// Don't preallocate initial slice
		var i []int64
		for _, el := range e {
			i = append(i, el)
		}
	}
}


func BenchmarkPreallocate(b *testing.B) {
  b.ReportAllocs()
	e := make([]int64, 10, 10)
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		// Preallocate initial slice
		i := make([]int64, 0, len(e))
		for _, el := range e {
			i = append(i, el)
		}
	}
}

```

```
BenchmarkNoPreallocate-12    	 6262358	       316 ns/op	     248 B/op	       5 allocs/op
BenchmarkPreallocate-12      	39661172	        56.7 ns/op	      80 B/op	       1 allocs/op
```
go version go1.13.3 linux/amd64

### DIFF
--- a/api/handler/api/util_test.go
+++ b/api/handler/api/util_test.go
@@ -8,7 +8,7 @@ import (
 
 func TestRequestToProto(t *testing.T) {
 	testData := []*http.Request{
-		&http.Request{
+		{
 			Method: "GET",
 			Header: http.Header{
 				"Header": []string{"test"},
--- a/api/handler/http/http_test.go
+++ b/api/handler/http/http_test.go
@@ -27,7 +27,7 @@ func testHttp(t *testing.T, path, service, ns string) {
 	s := &registry.Service{
 		Name: service,
 		Nodes: []*registry.Node{
-			&registry.Node{
+			{
 				Id:      service + "-1",
 				Address: l.Addr().String(),
 			},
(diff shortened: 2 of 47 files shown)


vtolstov:
sounds good, but can you split this big pr to smaller prs?
For example firstly sent gofmt stuff, and the second slice optimizations?

tegk:
So is it better for you to have a PR for each package or something?

milosgajdos:
Well, we need something that does not touch the whole of framework in one PR.

The risk of merging  a PR that touches 47 files must surely be understandable by you :-)

vtolstov:
gofmt commit needs to be in separate pr, also how you do it? my goimports not reformat like this

tegk:
@milosgajdos83 To be honest the risk is negligible in my opinion as there is no change in logic and the tests pass. I will nonetheless send a PR this weekend with changes per package or something like that.

@vtolstov `gofmt -w -s .`

milosgajdos:
The project is formatted by `gofmt` but maybe we do not use the extra switches you do.

I suspect `-s` changes the output a bit.

PR #908, opened 2019-11-02
fix-907

hb-chen:
web service need modify registry service nodes while register interval


asim:
Can you check we're doing the same for services?

hb-chen:
> Can you check we're doing the same for services?

You mean we need to check new service(`genSrv()`) and current service(`s.srv`) have the same service name and node id?

PR #909, opened 2019-11-03
avoid deadlock in syncMap.Iterate

spacewander:
Previously, when syncMap iterates a list of records which have the same
content in different order, a deadlock might happen. By enforcing a certain
order, the deadlock can be avoided.

### DIFF
--- a/sync/map.go
+++ b/sync/map.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"sort"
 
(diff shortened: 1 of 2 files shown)


spacewander:
The test case doesn't 100% reproduce the deadlock, because the output of `memory.Store.List()` is not determinate (sometimes it returns `[A, B]` and sometimes returns `[B, A]`).

It would be better if we can use https://github.com/golang/mock to mock the `Store.List` to return a certain result. Maybe in a separate PR?

PR #915, opened 2019-11-05
Flap detection vol2

milosgajdos:
This PR improves on the work started in https://github.com/micro/go-micro/pull/914

It makes the flap detection code more readable and arguably also more correct.

The core algorithm has not changed. There are two changes:

* advert processing has been moved to dedicated method (`process()`) of a dedicated type (`adverts` which used to be called `advertMap`)
*we now suppress events as soon as they've been received from the table

**NOTE:** previously we only suppressed events when `AdvertiseEventsTick` time has been reached, which is set to `5s` and by which time the events may have decayed to the point of recovery -- we want to suppress them as soon as we can and potentially even remove them should they have reached `MaxSuppressTime` threshold. (**This threshold should probably be lowered to something like 1 minute from the current 5 minutes**)

There is a small program I wrote to play with this implementation available in [this gist](https://gist.github.com/milosgajdos83/4b20724acc124e0db78860c98ba8d5c7) which lets you play with the code by artificially simulating table events (these are stored in the `test.txt` file which is also present in the gist)

### DIFF
--- a/router/default.go
+++ b/router/default.go
@@ -339,8 +339,8 @@ func (r *router) advertiseTable() error {
 type advert struct {
 	// event received from routing table
 	event *Event
-	// lastUpdate records the time of the last advert update
-	lastUpdate time.Time
+	// lastSeen records the time of the last advert update
+	lastSeen time.Time
 	// penalty is current advert penalty
 	penalty float64
 	// isSuppressed flags the advert suppression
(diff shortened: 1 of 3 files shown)


asim:
Can you add the gist or whatever else as a test so we can validate this code is correctly doing what we want

milosgajdos:
I was thinking about it, but could not think of a good way to do that as the code in `router` carries in a lot of cruft on its own.

asim:
We should be able to pass a memory registry to the router and start it?

PR #921, opened 2019-11-06
K8s list deployments

milosgajdos:
Adds `ListDeployments` method to list `micro` deployments in Kubernetes runtime.

**NOTE:** For now this method is used on every tick to query K8s API and log the found deployments. We should probably not hammer K8s API every 10s, though, so let's discuss what the right time/tick should be.


asim:
Rather than polling should we be watching the events?

PR #926, opened 2019-11-08
fix# Change the Log level and add WarnLevel

huangxingx:
the log level should be
LevelFatal < LevelError < LevelInfo < LevelWarn < LevelDebug < LevelTrace


asim:
You have defined warn and it uses error. You have made error use info. This doesn't make any sense.

PR #930, opened 2019-11-09
store/cloudflare: fix dropped error

alrs:
This PR picks up a dropped error in the `store/cloudflare` package.


asim:
Thanks for the fix

PR #931, opened 2019-11-09
fix data race for server Wait option

vtolstov:


server/rpc_server.go
```diff
@@ -744,10 +746,12 @@ func (s *rpcServer) Start() error {
 			log.Logf("Server %s-%s deregister error: %s", config.Name, config.Id, err)
 		}
 
+		s.Lock()
 		// wait for requests to finish
 		if s.wg != nil {
 			s.wg.Wait()
```
asim: Wait blocks, so is locking here not going to block everything else in the entire server?

vtolstov: yeah, i'm recheck the server logic

Issue #927, opened 2019-11-08
Can we provide service isolation options between networks.


hb-chen:
`DC2`->`DC1`, `DC3`->`DC1`, now `DC2` and `DC3` can share service through `DC1`, and their communication is through `DC1` network.
But some times I don't want `DC2` and `DC3` know each other.

asim:
We're working on something for network rules but would be interested to hear your ideas. How would this be configured?

hb-chen:
Router's QueryOptions
```go
// QueryOptions are routing table query options
type QueryOptions struct {
	// Service is destination service name
	Service string
	// Address of the service
	Address string
	// Gateway is route gateway
	Gateway string
	// Network is network address
	Network string
	// Router is router id
	Router string
}
```
I think we can add a `Link` option.If `Link="local"` now only local services will be shared with other networks.

hb-chen:
I have a test.
`router/default.go` function `flushRouteEvents` and `advertiseEvents` add `route.Link!="local"` filter.

```go
// advertiseEvents advertises routing table events
// It suppresses unhealthy flapping events and advertises healthy events upstream.
func (r *router) advertiseEvents() error {
	// ticker to periodically scan event for advertising
	ticker := time.NewTicker(AdvertiseEventsTick)
	defer ticker.Stop()

	// advertMap is a map of advert events
	advertMap := make(map[uint64]*routeAdvert)

	// routing table watcher
	tableWatcher, err := r.Watch()
	if err != nil {
		return fmt.Errorf("failed creating routing table watcher: %v", err)
	}

	r.wg.Add(1)
	go func() {
		defer r.wg.Done()
		select {
		case r.errChan <- r.watchTable(tableWatcher):
		case <-r.exit:
		}
	}()

	for {
		select {
		case <-ticker.C:
			var events []*Event
			// collect all events which are not flapping
			for key, advert := range advertMap {
				// decay the event penalty
				delta := time.Since(advert.lastUpdate).Seconds()
				advert.penalty = advert.penalty * math.Exp(-delta*PenaltyDecay)

				// suppress/recover the event based on its penalty level
				switch {
				case advert.penalty > AdvertSuppress && !advert.isSuppressed:
					log.Debugf("Router supressing advert %d for route %s", key, advert.event.Route.Address)
					advert.isSuppressed = true
					advert.suppressTime = time.Now()
				case advert.penalty < AdvertRecover && advert.isSuppressed:
[...]

PR #932, opened 2019-11-10
Add the ability to only advertise local routes or don't advertise at all

asim:
This PR adds two new router advertise strategies; Local and None.

- AdvertiseLocal will only advertise the local routes. Related to https://github.com/micro/go-micro/issues/927
- AdvertiseNone will avoid advertising routes completely.

We now need to add the ability to set this as a flag/env var

### DIFF
--- a/router/default.go
+++ b/router/default.go
@@ -413,8 +413,10 @@ func (r *router) advertiseEvents() error {
 	for {
 		select {
 		case <-ticker.C:
-			// We may only be advertising local routes
-			local := r.options.Advertise == AdvertiseLocal
+			// If we're not advertising any events then sip processing them entirely
+			if r.options.Advertise == AdvertiseNone {
+				continue
+			}
 
 			var events []*Event
 
@@ -430,9 +432,8 @@ func (r *router) advertiseEvents() error {
 					continue
 				}
 
-				// if we only advertise local routes delete and bail out
-				if local && advert.event.Route.Link != "local" {
-					delete(adverts, key)
+				// if we only advertise local routes skip processing anything not link local
+				if r.options.Advertise == AdvertiseLocal && advert.event.Route.Link != "local" {
 					continue
 				}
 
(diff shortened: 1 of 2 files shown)


router/default.go
```diff
@@ -728,32 +750,42 @@ func (r *router) flushRouteEvents(evType EventType) ([]*Event, error) {
...
+		if current.Metric == route.Metric {
 			if route.Router == r.options.Id {
 				bestRoutes[routeKey] = route
 				continue
 			}
 		}
 	}
 
-	log.Debugf("Router advertising %d best routes out of %d", len(bestRoutes), len(routes))
+	log.Debugf("Router advertising %d %s routes out of %d", len(bestRoutes), r.options.Advertise.String(), len(routes))
```
milosgajdos: `.String()` is not required here as `%s` will explicitly call `.String()`

router/default.go
```diff
@@ -426,6 +430,12 @@ func (r *router) advertiseEvents() error {
 					continue
 				}
 
+				// if we only advertise local routes delete and bail out
+				if local && advert.event.Route.Link != "local" {
```
milosgajdos: Makes me wonder whether we should skip adding the routes to the adverts map in the first place as soon as they're received.

router/default.go
```diff
@@ -714,6 +730,12 @@ func (r *router) flushRouteEvents(evType EventType) ([]*Event, error) {
 		return nil, fmt.Errorf("failed listing routes: %s", err)
 	}
 
+	// Do not advertise anything
+	if r.options.Advertise == AdvertiseNone {
```
milosgajdos: Shouldnt this be called right the first thing i.e. before we even call `List()` on router?

router/default.go
```diff
@@ -436,6 +446,12 @@ func (r *router) advertiseEvents() error {
 				delete(adverts, key)
 			}
 
+			// If we're not advertising any events then throw them away.
+			// We do this here because someone may change the strategy at any time.
+			if r.options.Advertise == AdvertiseNone {
```
milosgajdos: Shouldnt we call this before we start iterating over the adverts?

router/default.go
asim: Good spot

router/default.go
asim: Done

router/default.go
asim: Yea good idea, done.

PR #934, opened 2019-11-10
preallocation of slices

tegk:
"While the Go does attempt to avoid reallocation by growing the capacity in advance, this sometimes isn't enough for longer slices. If the size of a slice is known at the time of its creation, it should be specified."


tegk:
This should be the last PR for this...

asim:
Thanks. Curious to know why the changes if you're not using this at work.

tegk:
I am using this for personal non commercial projects and I like to contribute to open source projects.

PR #935, opened 2019-11-11
fix# Change the Log level and add WarnLevel

huangxingx:
Careless last time #926 . Sorry. I fixed it.


vtolstov:
Thanks, now looks good.

PR #942, opened 2019-11-13
Write a test to prove memory registry TTLs are busted

erikh:
memory registries currently suffer a minimum reaping limit of a single minute; this makes it incredibly hard to use it in a testing scenario. This behavior is iconified in the tests in the first commit.

The second commit was mostly just a demonstration of the timing being the issue. If the system needed to wait a whole minute to clear any TTLs, not only would it make testing a single service restart very painful but also it doesn't make sense for many production scenarios where TTLs should be much shorter.

Overhead of the goroutine is a problem. @milosgajdos83 has tried to approach this in other PRs... I want to use this PR as a way to address the problem one way or another.

### DIFF
--- a/registry/memory/memory_context.go
+++ b/registry/memory/memory_context.go
@@ -1,0 +1,330 @@
+package memory
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/micro/go-micro/registry"
+	"github.com/micro/go-micro/util/log"
+)
+
+type contextNode struct {
+	ctx    context.Context
+	cancel context.CancelFunc
+}
+
+type ContextRegistry struct {
+	options registry.Options
+
+	sync.RWMutex
+	Services map[string][]*registry.Service
+	nodes    map[string]map[string]contextNode
+	nodeLock sync.Mutex
+	Watchers map[string]*Watcher
+}
+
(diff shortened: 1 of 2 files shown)


erikh:
cc @milosgajdos83

milosgajdos:
I feel like `nanosecond` is super low. Otherwise it looks ok.

cc @asim

vtolstov:
what about cpu usage in case of 30 services and nanoseconds?

erikh:
ignore the second commit, please. it's largely to demonstrate the code does work if the timings are correct.

erikh:
(which it doesn't appear to be doing)

erikh:
I've updated the description of the ticket with more details; sorry, was in a rush.

erikh:
Discussions in slack lead me to think that we can resolve this with contexts that work like promises more or less; contexts are registered with `context.WithTimeout()` and could be simply iterated over for a given service and `Done()` could be checked to see if the record is expired or not, at which point it can be pruned *at query time* if done safely.

WDYT?

erikh:
I've included an alternative implementation called the "ContextRegistry". This registry uses golang contexts and `context.WithTimeout` to implement TTLs; these TTLs are then resolved at querying time instead of the background. Seems to be efficient enough and I doubt it will break due to a timing issue.

asim:
Was merged into the other pr

PR #943, opened 2019-11-14
SignalChan provides a way to override micro's signal behavior

erikh:
If provided, it will use this provided channel and not install a signal
handler of its own.

This is the simplest approach I could think of.

Signed-off-by: Erik Hollensbe <github@hollensbe.org>

### DIFF
--- a/options.go
+++ b/options.go
@@ -2,6 +2,7 @@ package micro
 
 import (
 	"context"
+	"os"
 	"time"
 
 	"github.com/micro/cli"
@@ -31,6 +32,8 @@ type Options struct {
 	// Other options for implementations of the interface
 	// can be stored in a context
 	Context context.Context
+
+	SignalChan chan os.Signal
 }
 
 func newOptions(opts ...Option) Options {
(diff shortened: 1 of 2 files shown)


vtolstov:
@asim i'm prefer to have ability to disable signal processing inside server at all.
what you think?
@erikh how can i disable signal processing with such option? I need to create new non nil signal channel and pass to service?

erikh:
pass nil to the option

erikh:
I documented the flag, fwiw.

vtolstov:
@erikh if i pass nil to option, this code https://github.com/micro/go-micro/pull/943/files#diff-4aae454123f50b906f9aaef44f10011eR171 precreate chan automatic and all work like before, but i want to say micro Server don't handle any signals, i want to handle it by my own

erikh:
No notifications are registered on the channel; it's basically a no-op if you pass nil, the channel will never produce any data. Then you can handle signals however you'd like. If you want to register a specific set of signals to the channel, then call `signal.Notify` yourself on a channel you pass.

There is a small bug in this, the default value of `SignalChan` on the options struct is not nil, which means the signal handler will never get installed properly right now. I need to fix that, so don't merge this until after that.

erikh:
https://github.com/micro/go-micro/pull/943/files#diff-4aae454123f50b906f9aaef44f10011eR169-R173 for more information.

erikh:
that should be fixed now

vtolstov:
@erikh thanks, now i'm understand. I'm misread pr code.

vtolstov:
@asim can we merge this?

asim:
No we can't really merge this right now. I'm not sure we should leak something like a signal channel out. It might be more appropriate to just operate on Context in the event it's passed in and ignore use of signal in that specific case. Otherwise we're always going to be tied to some option that allows someone to pass in a signal channel for interrupt. I just have a bad feeling it's going to be abused.

erikh:
this is exactly what I suggested in slack before. Would you like me to do
that patch instead?

asim:
Yea that would be good thanks but don't close this so we can compare

asim:
closing in favour of the other approach

Issue #889, opened 2019-10-28
Request's Body without Content-type becomes empty


mirwaisx:
Hi
First of all let me thank you for this framework, it's really well documented and with examples we can learn faster the usage.

I am using for now micro with api handler and I came across a strange behavior: with there is no Content-Type present in the request, micro adds "application/x-www-form-urlencoded" to the header and later in the code (line 38), it does not read the Body from request:

https://github.com/micro/go-micro/blob/d89256d8d57b7fcf523b769d9d2e62dea3c0e8ed/api/handler/api/util.go#L32

This behavior changes the original http.request and I think it should not be the case, or am I missing something else?

NB: to get around this problem I wanted to add a plugin which acts like a middleware and adds a Content-type (but different than application/x-www-form-urlencoded) if it's missing, somehow I could not make it work, but this is another question in case I have to go with this solution.

mirwaisx:
I see this as a Bug. Anyone else?

PR #946, opened 2019-11-14
bugfix #889, setting request body correctly

mirwaisx:
This is a fix for this: #889 
I think we need to set request Body anyways, independent of the `content-type`, otherwise this information is lost in `go-micro` api.

### DIFF
--- a/api/handler/api/util.go
+++ b/api/handler/api/util.go
@@ -27,19 +27,17 @@ func requestToProto(r *http.Request) (*api.Request, error) {
 		Url:    r.URL.String(),
 	}
 
-	ct, _, err := mime.ParseMediaType(r.Header.Get("Content-Type"))
-	if err != nil {
-		ct = "application/x-www-form-urlencoded"
-		r.Header.Set("Content-Type", ct)
+	if _, ok := r.Header.Get("Content-Type"); ok {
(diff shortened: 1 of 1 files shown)


api/handler/api/util.go
```diff
@@ -27,19 +27,17 @@ func requestToProto(r *http.Request) (*api.Request, error) {
 		Url:    r.URL.String(),
 	}
 
-	ct, _, err := mime.ParseMediaType(r.Header.Get("Content-Type"))
-	if err != nil {
-		ct = "application/x-www-form-urlencoded"
-		r.Header.Set("Content-Type", ct)
+	if _, ok := r.Header.Get("Content-Type"); ok {
```
milosgajdos: `Header.Get` returns one variable, not two
https://golang.org/pkg/net/http/#Header.Get

mirwaisx: strange otherwise VS Code shows me that error, thanks I did not want to build it all again as I have go version 1.12 and did not want to mess up my environment. I'll update that. What about the body part? do you also think to pass it around independent of the Content Type.

mirwaisx:
I do not know why Travis is failing: that TestRequestToProto is passing fine locally on my environment.

asim:
Please resubmit when you have run tests.

PR #947, opened 2019-11-14
Micro Runtime 

milosgajdos:
This PR adds the first draft implementation of `go-micro` kubernetes runtime.

A short summary of the [many] changes:
* Kuberentes API client implementation changes to enable to create K8s Service and Deployment objects
* A bunch of necessary K8s types have been added to make K8s API communication work
* `DefaultService` and `DefaultDeployment` functions are provided -- they generate a default `micro` service and deployment K8s objects which are then overridden based on `CreateOptions`
* runtime service interface has been updated to add `CreateOptions` and `Get` method to query a service
* few tests have been added to make sure the K8s API requests use the correct parameters
* Many other changes have been added


runtime/kubernetes/client/kubernetes.go
```diff
@@ -1,43 +1,149 @@
...
+
+var (
+	// DefaultImage is default micro image
+	DefaultImage = "micro/micro"
+	// ServiceRegexp is used to validate service name
+	ServiceRegexp = regexp.MustCompile("^" + dns1123LabelFmt + "$")
+)
+
 // Kubernetes client
 type Kubernetes interface {
```
asim, line 26: This would probably be good as 

```
type Kubernetes interface {
    Deployment() Deployment
    Service() Service
}

type Deployment interface {
    Create() ...
    Delete() ...
    ...
}

type Service interface {
    Create() ...
    Delete() ...
    ...
}
```

runtime/kubernetes/client/kubernetes.go
```diff
@@ -1,43 +1,149 @@
...
+	if err == nil {
+		buildUnixTimeUTC := time.Unix(buildTime, 0)
+		Metadata.Annotations = map[string]string{
+			"build": buildUnixTimeUTC.Format(time.RFC3339),
+		}
+	} else {
+		log.Debugf("Runtime could not parse build: %v", err)
+	}
+
+	Env := []EnvVar{
```
asim: We should leave this empty and then pass in the env configuration in the `micro runtime` service as something like

```
MICRO_REGISTRY=service
MICRO_BROKER=service
MICRO_REGISTRY_ADDRESS="registry-service"
MICRO_BROKER_ADDRESS="broker-service"
```

runtime/default.go
```diff
@@ -219,6 +226,30 @@ func (r *runtime) Update(s *Service) error {
...
 
+// Delete removes the service from the runtime and stops it
+func (r *runtime) Delete(s *Service) error {
+	r.Lock()
+	defer r.Unlock()
+
+	log.Debugf("Runtime deleting service %s", s.Name)
+	if s, ok := r.services[s.Name]; ok {
+		// check if running
+		if !s.Running() {
```
jakexks, line 237: This is weird

```go
if s.Running() {
	if err := s.Stop(); err != nil {
		return err
	}
}
delete(r.services, s.Name)
return nil

File: runtime/default.go [hunk 4]
```diff
@@ -175,28 +176,34 @@ func (r *runtime) Create(s *Service, opts ...CreateOption) error {
...
+	var services []*Service
+	// if we track the service check if the version is provided
+	if s, ok := r.services[name]; ok {
+		if len(gopts.Version) > 0 {
+			if s.Version == gopts.Version {
+				services = append(services, s.Service)
+			}
+			return services, nil
+		}
+		// no version ha sbeen requested, just append the service
```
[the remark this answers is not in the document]
milosgajdos: Damn those typos!

runtime/kubernetes/client/kubernetes.go
milosgajdos: Need to come up with some renaming as there are already types called `Service` and `Deployment` in `types.go` -- these are the core K8s object structs

asim: ServiceAPI interface?

milosgajdos: Wait, I'm testing a different idea now. Will push shortly. Might be better 🤞

PR #948, opened 2019-11-15
Signal handling try #2: through the context

erikh:
This approach expects the user to control their context to manage the
service. A signal handler is installed by default by you can specify the
option `NoSignalHandler()` to turn it off.

The choice of keeping the signal handler in was intended to preserve
backwards compatibility so that users of go-micro today were not
surprised by the signal handling changes.

Signed-off-by: Erik Hollensbe <github@hollensbe.org>

### DIFF
--- a/options.go
+++ b/options.go
@@ -31,6 +31,8 @@ type Options struct {
 	// Other options for implementations of the interface
 	// can be stored in a context
 	Context context.Context
+
+	Signal bool
 }
 
 func newOptions(opts ...Option) Options {
@@ -42,6 +44,7 @@ func newOptions(opts ...Option) Options {
 		Registry:  registry.DefaultRegistry,
 		Transport: transport.DefaultTransport,
 		Context:   context.Background(),
+		Signal:    true,
 	}
 
 	for _, o := range opts {
@@ -78,6 +81,15 @@ func Client(c client.Client) Option {
 func Context(ctx context.Context) Option {
 	return func(o *Options) {
 		o.Context = ctx
+	}
+}
+
+// HandleSignal toggles automatic installation of the signal handler that
+// traps TERM, INT, and QUIT.  Users of this feature to disable the signal
+// handler, should control liveness of the service through the context.
+func HandleSignal(b bool) Option {
+	return func(o *Options) {
+		o.Signal = b
 	}
 }
 
--- a/service.go
+++ b/service.go
@@ -169,7 +169,9 @@ func (s *service) Run() error {
 	}
 
 	ch := make(chan os.Signal, 1)
-	signal.Notify(ch, syscall.SIGTERM, syscall.SIGINT, syscall.SIGQUIT)
+	if s.opts.Signal {
+		signal.Notify(ch, syscall.SIGTERM, syscall.SIGINT, syscall.SIGQUIT)
+	}
 
 	select {
 	// wait on kill signal


erikh:
cc @asim

vtolstov:
I'm prefer inverted option.
So like DisableSigHandler.
By default it false and all works like before, with true signal handler not created.

asim:
I agree with Vasiliy. Inversion is better but I'm also a fan of brevity. Just see the current codebase. Look at the options. If I could mention a style guide I would but we don't really have one because the code is that.

```
// Signal enables the ability to enable/disable signal handling for interrupt
type Signal(b bool) micro.Option {
    return func(o *Options) {
        o.Signal = b
    }
}
```

That's effectively what it should be

erikh:
So we don't care about backwards compat at all?

erikh:
nvm, I think I understand. I'll get to it tonight

options.go
```diff
@@ -31,17 +31,20 @@ type Options struct {
 	// Other options for implementations of the interface
 	// can be stored in a context
 	Context context.Context
+
+	InstallSignalHandler bool
```
vtolstov: // Don't use internal os.Signal handler
NoSignal bool

options.go
```diff
@@ -81,6 +84,15 @@ func Context(ctx context.Context) Option {
 	}
 }
 
+// NoSignalHandler turns off automatic installation of the signal handler.
+// Users of this feature should control liveness of the service through the
+// context.
+func NoSignalHandler() Option {
```
vtolstov: use asim provided example please

vtolstov (changes requested):

options.go
asim: `Signal bool` which should default to true

vtolstov: ah ok

erikh:
I called it signalhandler -- I felt signal was confusing. I hope that's ok. Otherwise I think it's what you want.

options.go
```diff
@@ -81,6 +83,15 @@ func Context(ctx context.Context) Option {
 	}
 }
 
+// SignalHandler toggles automatic installation of the signal handler.
+// Users of this feature to disable the signal handler, should control liveness
+// of the service through the context.
+func SignalHandler(b bool) Option {
```
asim: HandleSignal might make more sense since the option will be called as

```
micro.NewService(
    micro.HandleSignal(bool),
)
```

options.go
```diff
@@ -31,6 +31,8 @@ type Options struct {
 	// Other options for implementations of the interface
 	// can be stored in a context
 	Context context.Context
+
+	DisableSignalHandler bool
```
asim: Too verbose. See the other options. Call it `Signal bool` and add a comment to indicate what its for.

options.go
```diff
@@ -31,6 +31,8 @@ type Options struct {
...
 }
 
 func newOptions(opts ...Option) Options {
```
asim, line 38: You may need to set `Signal: true` here so that its enabled by default.

service.go
```diff
@@ -168,14 +168,17 @@ func (s *service) Run() error {
...
-	ch := make(chan os.Signal, 1)
-	signal.Notify(ch, syscall.SIGTERM, syscall.SIGINT, syscall.SIGQUIT)
-
-	select {
-	// wait on kill signal
-	case <-ch:
-	// wait on context cancel
-	case <-s.opts.Context.Done():
+	if s.opts.DisableSignalHandler {
+		<-s.opts.Context.Done()
```
asim: Surely if `if s.opts.Signal {` then we add `signal.Notify(ch, syscall.SIGTERM, syscall.SIGINT, syscall.SIGQUIT)`. In this way the code will remain clean as before, but we'll just never receive anything on the `ch` channel.

options.go
erikh: sg

service.go
erikh: this is literally what I was doing in #943. Which one do you prefer?

erikh: Your comments: https://github.com/micro/go-micro/pull/943#issuecomment-554111145 am I confused?

asim: Misunderstanding

```
ch := make(chan os.Signal)

if s.opts.Signal {
    signal.Notify(ch, syscall.SIGTERM, syscall.SIGINT, syscall.SIGQUIT)
}

// rest remains the same
select {
case <-ch:

case <-s.opts.Context.Done():

}
```

erikh:
PTAL

PR #970, opened 2019-11-21
improve the syncMap.Iterate test to make it 100% reproducible

spacewander:


store/mocks/Store.go
```diff
@@ -0,0 +1,103 @@
+// Code generated by mockery v1.0.0. DO NOT EDIT.
```
asim: Can you lower case this filename

vtolstov: pass to mockery -case underscore so all output files will be lowercase
```
-case string
        name the mocked file using casing convention [camel, snake, underscore] (default "camel")
```

spacewander: @asim @vtolstov 
Addressed in the new commit.

store/mocks/store.go
```diff
@@ -0,0 +1,103 @@
+// Code generated by mockery v1.0.0. DO NOT EDIT.
+
+package mocks
```
asim: Can this package be called "mock"?

vtolstov: hm, does some other projects uses mock and not mocks?
by the way:
```
 -outpkg string
       name of generated package (default "mocks")
 -output string
       directory to write mocks to (default "./mocks")
```
so this is easy to be setted to mock

spacewander: @asim @vtolstov 
Updated.

PR #976, opened 2019-11-25
First interface for auth

asim:
The start of an auth interface. We have deferred authentication in Micro for a long time, mostly because auth was not believed to be a core networking concern. As we move to a fully integrated framework for all things microservices its clear that auth has to be part of the core system. Every service should transparently have auth built in, without the requirement for configuration. It does not have to be mandatory but it should be there.

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -1,0 +1,26 @@
+// Package auth provides authentication and authorization capability
+package auth
+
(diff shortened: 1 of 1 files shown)


xmlking:
client cert based auth would  be nice for  use case: micro 2 micro authentication

PR #978, opened 2019-11-25
Update the runtime for k8s name formatting and move Get to Read endpoint

asim:
- Move Get to Read endpoint to just end up with CRUD for runtime interface
- Add a client.Format method to make service names compliant with k8s names
- Set name/version as part of deployment annotations and extract back out
- Delete name/version/source from metadata on extraction 
- Move some values out of kubernetes/client to potential set the path for reusability or extraction


runtime/kubernetes/kubernetes.go
```diff
@@ -86,19 +85,24 @@ func (k *kubernetes) Create(s *runtime.Service, opts ...runtime.CreateOption) er
...
+		name = name + "-" + s.Version
 	}
 
-	if !client.ServiceRegexp.MatchString(svcName) {
-		return fmt.Errorf("invalid service name: %s", svcName)
+	// format as we'll format in the deployment
+	name = client.Format(name)
+
+	// check the name to ensure what we pass into metadata will be validated
+	if !client.ServiceRegexp.MatchString(name) {
```
milosgajdos: I really do wonder whether we should be doing the `regexp` validation ourselves or whether we should just rely on K8s and return error if K8s validation fails 🤷‍♂️I feel there is a rabbit hole looming over here, but I might be wrong.

asim: Yea I agree. maybe we should just try to shovel cruft at k8s and expect that it does the right thing. Will strip it out.

PR #979, opened 2019-11-25
[WIP] Tunnel encryption

milosgajdos:
First stab at `network.Tunnel` encryption. 

We use [AES](https://golang.org/pkg/crypto/aes/) symmetric encryption to encrypt the body of the transport messages sent over tunnel sessions.

The key aka passphrase to encrypt the payload is set to `token+session_name+channel_name` at the moment.

Open questions:
* what should we do if the `token` is empty/not found?
* should we also encrypt headers?

### DIFF
--- a/tunnel/session.go
+++ b/tunnel/session.go
@@ -301,13 +301,27 @@ func (s *session) Send(m *transport.Message) error {
 		// no op
 	}
 
+	// get the token
+	token, ok := m.Header["Micro-Tunnel-Token"]
+	if !ok {
+		// TODO: should we continue or return error
+		log.Debugf("no token found, insecure channel")
+	}
+
+	// encrypt the transport message payload
+	body, err := Encrypt(m.Body, token+s.channel+s.session)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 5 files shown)


asim:
- what should we do if the token is empty/not found?

We should just keep encrypting with token + session + channel even where token is blank

- should we also encrypt headers?

Yes let's encrypt all the headers inside the session. I think Micro-Tunnel- headers will be injected before and after so the routing will continue to work but user data will be encrypted

- As per discussion

We should no longer share token across the connection, instead we should just use it to encrypt/decrypt and throwaway connections/packets where we can't decrypt

tunnel/session.go
```diff
@@ -301,13 +301,27 @@ func (s *session) Send(m *transport.Message) error {
 		// no op
 	}
 
+	// get the token
+	token, ok := m.Header["Micro-Tunnel-Token"]
```
asim: Token should probably be added to `t.newSession` in default so that its accessible internally as I don't think this exists in the header. It's injected after session.Send and stripped before session.Recv.

Issue #832, opened 2019-10-09
Recover subscriber?


three-zhang:
Subscriber crash will interrupt the program without recover

milosgajdos:
Can you provide some code sample that can help us reproduce the issue?

We also welcome PRs

vtolstov:
close as resolved

PR #981, opened 2019-11-25
subscriber recovery

vtolstov:


vtolstov:
may be fixes https://github.com/micro/go-micro/issues/832

asim:
You do not need a channel for this. An error value can be set in defer before returning.

vtolstov:
@asim does the never commit looks good?

asim:
Yea makes sense

PR #993, opened 2019-11-26
Add fixes for label selector and skipping things that don't match

asim:


runtime/kubernetes/client/client.go
```diff
@@ -126,9 +126,9 @@ func (c *client) Update(r *Resource) error {
 
 	switch r.Kind {
 	case "service":
-		req.Body(r.Value.(*Service).Spec)
+		req.Body(r.Value.(*Service))
```
milosgajdos, line 129: This is a good catch!

PR #996, opened 2019-11-28
fix http_broker bug port "0"

printfcoder:
fix the bug 'missing service port'. this bug occurs on we don't set the broker port.

### DIFF
--- a/broker/http_broker.go
+++ b/broker/http_broker.go
@@ -643,6 +643,12 @@ func (h *httpBroker) Subscribe(topic string, handler Handler, opts ...SubscribeO
 		return nil, err
 	}
 
+	// if zero port then randomly assign one
+	if len(port) > 0 && port == "0" {
+		i := rand.Intn(20000)
+		port = fmt.Sprintf("%d", 10000+i)
+	}
+
 	addr, err := maddr.Extract(host)
 	if err != nil {
 		return nil, err


asim:
This is not the correct fix. If port is 0 it means Connect() was not called and the server has not yet assigned a port. This will basically subscribe the http broker using a completely random port which means unless the server is going to listen on the identical port or the expectation is that this is indirectly subscribing for something else it will not work. 

Can you explain what you're trying to fix and what the existing usage is and why Connect() is not being called before subscribe? If this is a bug related to some other part of the code we should fix that.

printfcoder:
just want to use Broker before calling service.Run():

```
	service := micro.NewService(
		micro.Name("go.micro.config"),
	)
	service.Init(func(options *micro.Options) {
		// ...
		domain.Init(service.Options().Broker)
		// ...
	})

	if err := service.Run(); err != nil {
		panic(err)
	}
```

```
ERRO[0000] [sub] broker sub err: missing service port    source="broker.go:50"
panic: [sub] broker sub err: missing service port
```

printfcoder:
I know I can use  like this:

```
	service.Init(func(options *micro.Options) {
		config.Init()
		options.Broker.Init()
		options.Broker.Connect()
		domain.Init(options.Broker)
		facade.Init(service.Server())
	})
```

but this is not a graceful way.

vtolstov:
If you need to use broker before run service call Connect and use it, what the problem?

printfcoder:
OK. trying another way.

PR #1004, opened 2019-11-30
Debug logs

milosgajdos:
This is the WIP for the logs support in `go-micro`.

Defines `log.Log` interface.

What I dont like too much + questions:
* should `log.Log` have `Stream` method?
* once we stop streaming logs using `micro logs` (this is RPC request to `debug.Handler`) the stream does not seem to stop immediately (note there is a `stop` channel that allows the handler to clean up its list of subscribers) and the map might potentially bloat the handler.

In summary, things seem to work as they are, but we need to put some proper thinking into this.


debug/handler/debug.go
```diff
@@ -39,3 +45,66 @@ func (d *Debug) Stats(ctx context.Context, req *proto.StatsRequest, rsp *proto.S
 	rsp.Threads = uint64(runtime.NumGoroutine())
 	return nil
 }
+
+func (d *Debug) Logs(ctx context.Context, req *proto.LogRequest, stream proto.Debug_LogsStream) error {
+	var options []log.ReadOption
+
+	since := time.Unix(0, req.Since)
```
asim: I think it's supposed to be time.Unix(req.Since, 0) right?

debug/log/log.go
```diff
@@ -0,0 +1,179 @@
...
+	// prefix for all messages
+	prefix string
+)
+
+// Log is event log
+type Log interface {
+	// Read reads log entries from the logger
+	Read(...ReadOption) []Record
+	// Write writes logs to logger
+	Write(...interface{})
```
asim: This should probably be ...Record so that metadata can be passed through but I know that's going to be an annoying change below so can always be done later.

debug/proto/debug.proto
```diff
@@ -1,20 +1,19 @@
 syntax = "proto3";
 
 service Debug {
- 	rpc Health(HealthRequest) returns (HealthResponse) {}
- 	rpc Stats(StatsRequest) returns (StatsResponse) {}
+ 	rpc Health(HealthRequest) returns (HealthResponse) {};
+ 	rpc Stats(StatsRequest) returns (StatsResponse) {};
+        rpc Logs(LogRequest) returns (stream Log) {};
```
asim: Just a formatting issue

debug/proto/debug.proto
```diff
@@ -28,3 +27,25 @@ message StatsResponse {
...
+        // relative time in seconds
+        // before the current time
+        // from which to show logs
+        int64 since = 2;
+        // stream logs continuously
+        bool stream = 3;
+}
+
+// Log is service log record
+message Log {
```
asim: Should this be Record? Or does it just look better as Log

asim:
Made a few comments. Nothing major. I think merging and refactoring after we've actually actively used it might be a good idea. Then we can learn what's working or isn't. 

Addressing the other comments. I guess if log.Log does not have a stream interface it might be difficult to make this work efficiently from kubernetes or say a message broker so probably makes sense to have it there. Although wondering whether it should take a close channel or return a Stream interface that can be stopped. 

The rpc issue I guess needs to be chased but not a deal breaker right now.

Otherwise all looks good 👍

debug/service/service.go
```diff
@@ -0,0 +1,86 @@
...
+			break
+		}
+
+		metadata := make(map[string]string)
+		for k, v := range resp.Metadata {
+			metadata[k] = v
+		}
+
+		record := log.Record{
+			Timestamp: time.Unix(0, resp.Timestamp),
```
asim: Also time.Since(resp.Timestamp, 0)

debug/handler/debug.go
milosgajdos: Depends on the resolution (https://play.golang.org/p/aVkhnRgurCS); I think we use `nanoseconds` in RPC requests, no? I feel like the `router` does...

asim: The proto says we use a unix seconds timestamp

milosgajdos: Damn. Good catch. Will change this tomorrow. (and the below)

asim:
One interesting question is whether there are methods for accepting context as the first arg, peeling out our metadata and then shoving that in there. Saves the user from specifying it.

e.g 

```
# user does
log.Print(ctx, "some stuff happened")
```

```
# we do
func Print(ctx context.Context, format string, v ...interface{}) {
    md, ok := metadata.FromContext(ctx)
    rec := Record{
       metadata: md,
    }
}
```

Thoughts?

vtolstov:
What you think about adding some structure logging, not fully like zerolog or zap.
But something like providing additional metadata from go-micro core? So inside go-micro broker,server and so we not only log something like log.Log("[gossip]: xxx bad") but provide metadata to this logs, like kind of component, and maybe some other info?
Today i'm wrap std logger with uber/zap and lose ability to proper understand what happened. I'm prefer to have ability to get LogWithMeta/LogfWithMeta that receives map[string]interface{} or map[string]string metadata... or maybe that receives ctx context.Context that contains needed metadata that can be extracted by metadata.FromContext ...

asim:
@vtolstov see my above comment and read the PR. We're adding metadata as part of the logging.

milosgajdos:
The "issue" I have with the `context` call is that if I dont want to pass any metadata downstream, I still need to pass context in. Other than that I think it's a good idea

debug/proto/debug.proto
milosgajdos: I think it makes sense to "map" the `proto` messages to `Go` code, so let's rename.

asim:
I have merged so that we have a working solution. Now we can iterate from here. I think there will be minor superficial interface and cli changes but it is better to have something to build on than cruft into one PR, I want to be able to tangibly kick the tyres on the working solution because experience is making realise its easier to comment based on hands on experience than just looking at the code.

milosgajdos:
Agree! I shall open a PR in micro with the basic cli interface cruft I tested this on so we have something to iterate on.

PR #1009, opened 2019-12-02
recover panics

vtolstov:


server/rpc_router.go
```diff
@@ -603,16 +616,16 @@ func (router *router) ProcessMessage(ctx context.Context, msg Message) error {
...
+			if err = fn(ctx, rpcMsg); err != nil {
 				results = append(results, err.Error())
 			}
 		}
 	}
 
 	// if no errors just return
-	if len(results) == 0 {
-		return nil
+	if len(results) != 0 {
```
asim: results > 0

vtolstov: fixed

server/rpc_router.go
```diff
@@ -505,6 +507,17 @@ func (router *router) Subscribe(s Subscriber) error {
 }
 
 func (router *router) ProcessMessage(ctx context.Context, msg Message) error {
+	var err error
```
acynothia, line 510: should define in named return value. it will doesn't work when defer recover.
```
func (router *router) ProcessMessage(ctx context.Context, msg Message) (err error) {
```

transport/grpc/handler.go
```diff
@@ -16,6 +17,8 @@ type microTransport struct {
 }
 
 func (m *microTransport) Stream(ts pb.Transport_StreamServer) error {
+	var err error
```
acynothia, line 20: same as above.

```
        fn := func() error {
            var err error
            defer func() {
                if r := recover(); r != nil {
                    err = fmt.Errorf("%v", r)
                }   
            }() 
            panic("micro")
            return err
        }   

        fmt.Println(fn()) // print nil
```

PR #1012, opened 2019-12-03
improve code quality

acynothia:
1. the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak (govet)
2. assigning the result of this type assertion to a variable (switch b := b.(type)) could eliminate the following type assertions (gosimple)
3. using time.Tick leaks the underlying ticker, consider using it only in endless functions, tests and the main package, and use time.NewTicker here (staticcheck)
4. should use a simple channel send/receive instead of `select` with a single case (gosimple)

### DIFF
--- a/network/default.go
+++ b/network/default.go
@@ -823,7 +823,7 @@ func (n *network) processCtrlChan(listener tunnel.Listener) {
 					log.Tracef("Network metric for router %s and gateway %s: %v", event.Route.Router, event.Route.Gateway, metric)
 
 					// check we don't overflow max int 64
-					if d := route.Metric + metric; d > math.MaxInt64 || d <= 0 {
+					if d := route.Metric + metric; d <= 0 {
 						// set to max int64 if we overflow
 						route.Metric = math.MaxInt64
 					} else {
--- a/network/service/handler/handler.go
+++ b/network/service/handler/handler.go
@@ -94,11 +94,6 @@ func (n *Network) Connect(ctx context.Context, req *pbNet.ConnectRequest, resp *
 
 // Nodes returns the list of nodes
 func (n *Network) Nodes(ctx context.Context, req *pbNet.NodesRequest, resp *pbNet.NodesResponse) error {
-	depth := uint(req.Depth)
-	if depth <= 0 || depth > network.MaxDepth {
-		depth = network.MaxDepth
-	}
-
 	// root node
 	nodes := map[string]network.Node{}
 
--- a/server/server.go
+++ b/server/server.go
@@ -201,7 +201,7 @@ func Run() error {
 	}
 
 	ch := make(chan os.Signal, 1)
-	signal.Notify(ch, syscall.SIGTERM, syscall.SIGINT, syscall.SIGKILL)
+	signal.Notify(ch, syscall.SIGTERM, syscall.SIGINT)
 	log.Logf("Received signal %s", <-ch)
 
 	return Stop()
--- a/web/service.go
+++ b/web/service.go
@@ -376,7 +376,7 @@ func (s *service) Run() error {
 	go s.run(ex)
 
 	ch := make(chan os.Signal, 1)
-	signal.Notify(ch, syscall.SIGTERM, syscall.SIGINT, syscall.SIGKILL)
+	signal.Notify(ch, syscall.SIGTERM, syscall.SIGINT)
 
 	select {
 	// wait on kill signal
--- a/agent/input/telegram/conn.go
+++ b/agent/input/telegram/conn.go
@@ -44,11 +44,9 @@ func (tc *telegramConn) run() {
 	tc.recv = updates
 	tc.syncCond.Signal()
 
-	for {
-		select {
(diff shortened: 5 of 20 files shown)


network/service/handler/handler.go
```diff
@@ -94,11 +94,6 @@ func (n *Network) Connect(ctx context.Context, req *pbNet.ConnectRequest, resp *
 
 // Nodes returns the list of nodes
 func (n *Network) Nodes(ctx context.Context, req *pbNet.NodesRequest, resp *pbNet.NodesResponse) error {
-	depth := uint(req.Depth)
-	if depth <= 0 || depth > network.MaxDepth {
```
vtolstov, line 98: why this removed?
first check <= 0 is redundant because of uint before, but why removed MaxDepth checking?

vtolstov (commented):
minor questions, but overall looks good

network/service/handler/handler.go
acynothia: but depth seems doesn't be used.

milosgajdos: Yeah, this seems to be a copy-paste leftover and should be removed completely.

tegk: indeed, depth is not used

vtolstov: nice

vtolstov:
@asim i think that this can be safety merged

network/default.go
```diff
@@ -823,7 +823,7 @@ func (n *network) processCtrlChan(listener tunnel.Listener) {
 					log.Tracef("Network metric for router %s and gateway %s: %v", event.Route.Router, event.Route.Gateway, metric)
 
 					// check we don't overflow max int 64
-					if d := route.Metric + metric; d > math.MaxInt64 || d <= 0 {
```
milosgajdos, line 826: Can you please make sure that we check that `d` is not `0` i.e. `d <= 0` ? If `route.Metric` and `metric` are `0` then `d` is `0` and we want to handle that properly.

server/server.go
```diff
@@ -201,7 +201,7 @@ func Run() error {
 	}
 
 	ch := make(chan os.Signal, 1)
-	signal.Notify(ch, syscall.SIGTERM, syscall.SIGINT, syscall.SIGKILL)
```
milosgajdos, line 204: Why did you remove `SIGKILL` from the list of signals to handle?

web/service.go
```diff
@@ -376,7 +376,7 @@ func (s *service) Run() error {
 	go s.run(ex)
 
 	ch := make(chan os.Signal, 1)
-	signal.Notify(ch, syscall.SIGTERM, syscall.SIGINT, syscall.SIGKILL)
```
milosgajdos, line 379: Ditto, my earlier comment.

milosgajdos:
There is one thing that needs to be addressed (around handling network route metrics). And I'm curious about the `SIGKILL` handler removal. Otherwise I agree with @vtolstov this looks great.

server/server.go
acynothia: because syscall.SIGKILL cannot be trapped. [http://linuxcommand.org/lc3_man_pages/kill1.html](http://linuxcommand.org/lc3_man_pages/kill1.html).

web/service.go
acynothia: try trap SIGKILL will never work.

network/default.go
asim: My mistake for merging without checking this. Can someone fix.

acynothia: first no value of type int64 is greater than math.MaxInt64. 
second, if we can make sure route.Metric >= 0 and metric >=0.
```
v1 := math.MaxInt64
v2 := math.MaxInt64
fmt.Println(v1+v2) // print -2  
```

milosgajdos: Yes I get the first part; but as I said I think `route.Metric` and `metric` can be `0` which will yield `d` to be `0` and we want to handle when that happens.

milosgajdos: Just to elaborate:
If the metric is not set by link (i.e. it's `0`), we can't make any assumption/guarantees about the "quality" of the route hence we set it to the MAX VALUE, because that link might be broken or for whatever reason we can't measure its properly, so we want to make sure it's considered to be bad hence we are setting it to the highest cost/metric possible

acynothia: you are right. thanks for review. i will remove it. sorry for my pool english.  :laughing: 

but it's not intuitive,  if d is zero set to MaxInt64.

milosgajdos: Yes, it's not very intuitive. In `micro` toolkit we check for it and render infinity character https://github.com/micro/micro/blob/0becdfac18d514a84ce67472d464edcff864a374/network/web/web.go#L232-L233

milosgajdos:
I think this is ok to be merged now @asim ; Awesome work @Astone-Chou on this!

network/default.go
acynothia: I got it, thanks for clarify.

PR #1016, opened 2019-12-03
lint fixes mostly for prealloc also remove deadcode from grpc client

vtolstov:


vtolstov:
needs some fixes

vtolstov:
now i think that this is ready to merge

PR #1019, opened 2019-12-05
No newlines at end of log messages

tboerger:
Log messages should not end with a new line, this should be entirely
handled by the underlying log library.

Signed-off-by: Thomas Boerger <thomas@webhippie.de>


vtolstov:
lgtm =)

PR #1021, opened 2019-12-05
add RegisterCheck web server option for internal health checks

vtolstov:


vtolstov:
@asim may be you have some questions or .. ?

asim:
I would rather not add things here. This thing needs to change.

vtolstov:
@asim
> I would rather not add things here. This thing needs to change.

i don't worry if after changing this gone and need something never.

PR #1024, opened 2019-12-05
add server Context option to pass own context

vtolstov:


asim:
What is the context for here? Last I recall we don't do any shutdown signalling in the server

vtolstov:
I'm plan to use it in RegisterCheck func and may be in some other places.

vtolstov:
@asim gentle ping

PR #1028, opened 2019-12-09
named return value for error defer modify.

acynothia:
```
        fn := func() error {
            var err error
            defer func() {
                if r := recover(); r != nil {
                    err = fmt.Errorf("%v", r)
                }   
            }() 
            panic("micro")
            return err
        }   

        fmt.Println(fn()) // print nil
```

```
        fn := func() (err error) {
            defer func() {
                if r := recover(); r != nil {
                    err = fmt.Errorf("%v", r)
                }   
            }() 
            panic("micro")
            return err
        }   

        fmt.Println(fn()) // print micro
```

### DIFF
--- a/server/grpc/subscriber.go
+++ b/server/grpc/subscriber.go
@@ -167,8 +167,7 @@ func validateSubscriber(sub server.Subscriber) error {
 }
 
 func (g *grpcServer) createSubHandler(sb *subscriber, opts server.Options) broker.Handler {
-	return func(p broker.Event) error {
-		var err error
+	return func(p broker.Event) (err error) {
 
 		defer func() {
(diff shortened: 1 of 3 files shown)


asim:
We tend to avoid named return values, please just var err error at the top

acynothia:
https://blog.golang.org/defer-panic-and-recover

If you want to modify variable in defer func u can only use named return value or pass the value reference into defer func.

```
       fn := func() error {
            var err error
            defer func() {
                if r := recover(); r != nil {
                    err = fmt.Errorf("%v", r)
                }   
            }() 
            panic("micro")
            return err
        }   

        fmt.Println(fn()) // print nil
```

It will print `<nil>`  in this example.

PR #1032, opened 2019-12-09
nlopes/slack dep update to fix broken unmarshal

jankremlacek:
About a month ago Slack introduced the updated structure of RTM messages which resulted in an inability to unmarshal received msg (original issue: https://github.com/nlopes/slack/issues/630). It's not an issue of micro itself, but of the github.com/nlopes/slack lib. The fix was already merged into master (https://github.com/nlopes/slack/pull/618), but the lib has not been released in any new version.

Thus so I propose to update directly to the commit:

go get github.com/nlopes/slack@d06c2a2b3249b44a9c5dee8485f5a87497beb9ea

The MicroBot does now work now with any Slack client newer than around a month old.

Pardon me please, if using @commit in go.mod is anyhow against your rules or methods, I didn't find anywhere a "code of conduct" or so. Thank you.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,7 +34,7 @@ require (
 	github.com/miekg/dns v1.1.22
 	github.com/mitchellh/hashstructure v1.0.0
 	github.com/nats-io/nats.go v1.9.1
-	github.com/nlopes/slack@d06c2a2b3249b44a9c5dee8485f5a87497beb9ea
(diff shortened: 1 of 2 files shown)


asim:
You need to actually run go get using the changed commit so that it properly version it and changes go sum, otherwise this will not work

jankremlacek:
I know, sorry for that, I have no idea why I didn't run `go mod tidy` 🤷🏻‍♂️ , will make it immediately.

PR #1046, opened 2019-12-16
Fix cockroachdb store implementation

jakexks:


store/cockroach/cockroach.go
```diff
@@ -47,6 +43,7 @@ func (s *sqlStore) List() ([]*store.Record, error) {
 	}
 	defer rows.Close()
 	for rows.Next() {
+		println("next!")
```
asim: Might want to remove this

PR #1049, opened 2019-12-17
Kubernetes logging

jakexks:
This extends the inbuilt Kubernetes client to have logging, and implements a skeleton for the Kubernetes logger


debug/log/kubernetes/log.go
```diff
@@ -0,0 +1,34 @@
...
+
+func (k *klogStreamer) Chan() <-chan log.Record {
+	if k.streamChan == nil {
+		k.streamChan = make(chan log.Record)
+	}
+	return k.streamChan
+}
+
+func (k *klogStreamer) Stop() error {
+	close(k.streamChan)
```
asim, line 32: Stop() may be called multiple times so you want to do this sort of select statement. But also because something might be sending to streamChan it may panic if you close it so you want to close a stop channel instead.

```
select {
case <-k.stop:
  return nil
default:
  close(k.stop)
}
```

jakexks: Yep, I'll worry about that while implementing

PR #1053, opened 2019-12-20
fix etcd LogConfig bug & add etcd grpc log setter

shikbupt:
fix etcd LogConfig bug & add etcd grpc log setter
LogSet is different from LogConfig. LogSet set the grpc communicate log between etcd client and server.
LogConfig only set the log in etcd client


asim:
Just call clientv3.SetLogger(grpclog) yourself. We do not need to bake in a global call.

PR #1054, opened 2019-12-20
Kubernetes logging

jakexks:
This is an implementation of the log interface that reads logs from the Kubernetes API


debug/log/kubernetes/kubernetes.go
```diff
@@ -2,29 +2,165 @@
...
+		defer logs.Close()
+		s := bufio.NewScanner(logs)
+		for s.Scan() {
+			record := k.parse(s.Text())
+			record.Metadata["pod"] = l
+			records = append(records, record)
+		}
+	}
+
+	sort.Sort(byTimestamp(records))
```
asim, line 59: You can use `sort.Slice` which accepts an inline function. Will stop from needing to define a whole sort type

debug/log/kubernetes/kubernetes.go
```diff
@@ -2,29 +2,165 @@
...
+	return klog
+}
+
+func (k *klog) getMatchingPods() ([]string, error) {
+	r := &client.Resource{
+		Kind:  "pod",
+		Value: new(client.PodList),
+	}
+	l := make(map[string]string)
+	l["micro"] = "runtime"
```
asim, line 143: Should probably just remove this for now since services created with k8s runtime will have "service" rather than runtime.

debug/log/kubernetes/kubernetes.go
```diff
@@ -2,29 +2,165 @@
...
+	if err := k.client.Get(r, l); err != nil {
+		return nil, err
+	}
+
+	var matches []string
+	for _, p := range r.Value.(*client.PodList).Items {
+		if p.Metadata.Labels["name"] == k.Options.Name {
```
asim, line 150: name will probably need to use util/kubernetes/client.Format since our dotted names will be switched

Issue #1060, opened 2019-12-25
support grpc client mux


jamsonzan:
grpc client supports mux, but we now only get a connection from pool. why not use the pool as a streams pool? we can use maxStreams to control the number of streams on a connection. while maxStreams = 1, it degenerate into a connection pool. [this is the code](https://github.com/jamsonzan/go-micro/blob/master/client/grpc/grpc_pool.go).
and we use maxIdle to control the number of idle conn. if need, i can create a pr.

vtolstov:
can you share some info for example speedup, or minimization of resource allocations, or some other benefits ?

jamsonzan:
yes, i had finished a [test](https://github.com/jamsonzan/testpool)

vtolstov:
nice, can you provide pr?

PR #1062, opened 2019-12-26
support streams pool for grpc

jamsonzan:
see [#1060](https://github.com/micro/go-micro/issues/1060)


vtolstov:
@asim with my minimal knowledge of connection pooling in go-micro, looks good. Can you review?

jamsonzan:
i had [test](https://github.com/jamsonzan/testpool) again, add concurrency = 200, 300, 400 and count the total number of dail. 
mux on tcp means that we can have a bigger connection pool without more tcp connection.

asim:
LGTM. Thank you. I may do some refactor/cleanup of naming but otherwise makes sense.

PR #1063, opened 2019-12-27
Fix missing recover while occur panic in handler

wudi:
Hi there,

Process exit.
By default rcpServer can't recover handler panic. 
Because of the recover code at top of ServeConn, but handler process in different coroutine and no recover.

Useless for process handler and outbound messages.
https://github.com/micro/go-micro/blob/master/server/rpc_server.go#L160

Missing recover
https://github.com/micro/go-micro/blob/master/server/rpc_server.go#L398-L403

Reference:
https://github.com/golang/go/wiki/PanicAndRecover

### DIFF
--- a/server/rpc_server.go
+++ b/server/rpc_server.go
@@ -374,6 +374,12 @@ func (s *rpcServer) ServeConn(sock transport.Socket) {
 				pool.Release(psock)
 				// signal we're done
 				wg.Done()
+
+				// recover any panics for outbound process
+				if r := recover(); r != nil {
+					log.Log("panic recovered: ", r)
+					log.Log(string(debug.Stack()))
+				}
 			}()
 
 			for {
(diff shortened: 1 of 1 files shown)


wudi:
Some other way to do it by handler wrapper. But must keep the wrapper first.

I think should do it in core framework.

example
```go
s := micro.NewService(
	micro.Name("demo.srv.test"),
	micro.Version("latest"),

        // Must keep recover wrapper first
	micro.WrapHandler(func(handlerFunc server.HandlerFunc) server.HandlerFunc {
		return func(ctx context.Context, req server.Request, rsp interface{}) error {
			defer func() {
				if r := recover(); r != nil {
					log.Log("panic recovered: ", r)
					log.Log(string(debug.Stack()))
				}
			}()

			return handlerFunc(ctx, req, rsp)
		}
	}),
)
```

asim:
Thanks very much. Is @tencent using micro?

wudi:
@asim  TKS for reply.
We try to use micro in small team. Internal promotion and use it.
Personally, I'm very interested in micro. Awesome design 👍

asim:
Thank you, very exciting to see micro at tencent. Please let me know how things progress. It would be amazing to see tencent adopt it at a bigger scale.

PR #1069, opened 2019-12-28
Moving to gRPC by default

asim:
Taking the steps to move to gRPC by default.

Was working on a way to do it cleanly but for now we'll replace the default implementations of the client and server. The full implementation will come soon.


wudi:
@asim  Why make the gRPC default directly?
Is there have same roadmap ?

asim:
Because everyone wants grpc by default. This PR did not yet enable it.

Felyne:
good.

weisd:
good

xmlking:
Thanks

PR #1070, opened 2019-12-30
set grpc by default

asim:
Enable grpc service by default.

Currently there is painful rewriting of code to make something a default so I'm opting for a high level switch. I think ideally we should move the current defaults into their own packages and then initialise the pkg.DefaultX values from some external method.

### DIFF
--- a/defaults.go
+++ b/defaults.go
@@ -1,0 +1,17 @@
+package micro
+
+import (
+	"github.com/micro/go-micro/client"
+	"github.com/micro/go-micro/server"
+
+	// set defaults
+	gcli "github.com/micro/go-micro/client/grpc"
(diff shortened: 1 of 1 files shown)


vtolstov:
yes, i'm do something like this in my code.
i have small parts and include it in higher level to some bigger block, and this bigger blocks compound to product.

asim:
I think my goal would be to create something like "profiles" or "environments" that let you define these as a grouping maybe under the config package. 

So then we could do something like 

```
profile := local.NewProfile()
// profile := kubernetes.NewProfile()

config.LoadProfile(profile)
```

xmlking:
Will micro cli also move to grpc?

PR #1075, opened 2020-01-02
remove ignore error in method publish

wudi:
Export registry error in method publish.

Show error while service name not found or registry occur some exceptions.


asim:
The reason we don't do this is because pubsub is mostly an async thing. Subscribers may not exist before a publisher starts sending messages. Whats the reason for returning the error?

crazybber:
need an explain @wudi

wudi:
I think any ready task occur error should export before async.
Just like Marshal error. https://github.com/micro/go-micro/blob/master/broker/http_broker.go#L525

asim:
Considering we're going to move away from http broker, I'm going to merge and lets see what happens.

PR #1083, opened 2020-01-05
Correct config name to resolve error on install

lujstn:
When using `go get`, an error appears with quic that this config property does not exist. It's actually called "MaxIdleTimeout", so just a quick PR to fix it.

Attaching to #1082 (although the core content of the post is different, this error is described in it).


asim:
We're using go modules so this change does not reflect the version we're using.

PR #1091, opened 2020-01-08
Add web service signal handler option

shikbupt:
Add web service signal handler option


vtolstov:
Sorry, but please rewrite pr to use consistent names like in HandleSignal(true/false)
By default HandleSignal is true

shikbupt:
> Sorry, but please rewrite pr to use consistent names like in HandleSignal(true/false)
> By default HandleSignal is true

sorry，but the pr is similar to the option in root(https://github.com/micro/go-micro/blob/master/options.go#L35). Should I rewrite the Signal to HandleSignal?

vtolstov:
@shikbupt in Options struct this is Signal variable, so you can use the same name. But Option arg for NewService receives option with name HandleSignal

shikbupt:
@vtolstov May I misunderstand you? In the code，I write a Signal variable and set a HandleSignal option function for it. Then NewService can reveice the HandleSignal function to set the Signal variable just as you say. Or, you said is that I should rewrite the pr title?

vtolstov:
strange, i saw that option have another name.. but now looks fine. sorry.
@asim  can you merge it?

PR #1092, opened 2020-01-08
support ability to set store, address and namespace via flags and env…

asim:
… vars

### DIFF
--- a/store/service/service.go
+++ b/store/service/service.go
@@ -28,6 +28,17 @@ type serviceStore struct {
 	Client pb.StoreService
 }
 
+func (s *serviceStore) Init(opts ...store.Option) error {
+	for _, o := range opts {
+		o(&s.options)
+	}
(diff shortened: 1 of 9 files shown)


store/service/service.go
```diff
@@ -28,6 +28,17 @@ type serviceStore struct {
 	Client pb.StoreService
 }
 
+func (s *serviceStore) Init(opts ...store.Option) error {
+	for _, o := range opts {
+		o(&s.options)
+	}
+	s.Namespace = s.options.Namespace
```
milosgajdos, line 35: Something irks me about the repetition/duplication of `serviceStore` struct fields and `store.Option` fields 🤷‍♂️

asim: I go back and forth between in. Being able to explictly see the things you use is useful but agree that there's duplication which is actually error prone. Not just limited to this implementation. Probably one for a wider cleanup.

PR #1099, opened 2020-01-09
config/reader.Values add Set for specific path merge

printfcoder:
jsonReader still has this method. it can help us conveniently merge data with a specific path.

eg.
```
{"a":{"b":{"c":{"d":{"name":"im d"}}}}}
```
we can merge `{d:{"name":"i am not  d"}}` into it via `Set({d:{"name":"i am not  d"}}, "a", "b", "c")

### DIFF
--- a/config/default.go
+++ b/config/default.go
@@ -172,6 +172,17 @@ func (c *config) Get(path ...string) reader.Value {
 	return newValue()
 }
 
+func (c *config) Set(val interface{}, path ...string) {
+	c.Lock()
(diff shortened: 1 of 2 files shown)


asim:
Build failed for some reason

printfcoder:
Ah! forgot to implement Set for config. fixed

asim:
Please test these changes and provide tests for them. It feels like you're just adding to the API rather than seeing if it works.

PR #1104, opened 2020-01-10
[WIP] Network hackery

milosgajdos:
This is a WIP to start the conversation about potential improvements of the network operation.

This PR adds the following changes:
*   Send `Solicit` message where we previously sent `Peer` message by accident
*   Added `Sync` message (**NOTE:** the name and/or semantics of this message will most likely change)
 `Sync` message is sent in response to `Connect` message. It contains all the routes and the peer graph of the receiving node. This will speed up the convergence of the local node and of the network in general. Upon receiving `Sync` message we respond by sending a peer message back to the (re)connecting peer origin.
*    We now gossip `Advert`s to up to 3 randomly selected peers instead of sending a multicast `Advert` message to the network.
*    Equally, `Solicit` messages i.e. full table adverts are also gossiped to a randomly selected peer from the node's `Peers`. Only if sending a "unicast" `Advert` message fails we send a multicast `Solicit` message to the network.
* Updated comments and documentation


vtolstov:
description looks good =)

asim:
Code looks good. Let me know how the testing formulates. I think there are probably two things I think about here. 

1. Yea the naming might be a quirk but maybe this is less of a concern and actually these things can be easily changed in a backwards compatible way considering we type switch and the content of the proto would not change. So lets move forward with Sync.

2. I wonder about what the behaviour of the network ends up looking like when you do gossip adverts versus full multicast, especially with those intermittent events. If a client connects to the network and is not receiving adverts because they are gossiped to random peers when will it next know about those changes? So e.g 100 clients connect to US, US and Asia regions, some event occurs and clients randomly receive new routes, now asia is preferred for some new service rather than a local region for a handful of the clients. How do we ensure when services change or new ones come online that you're getting routes to local services.

milosgajdos:
> I wonder about what the behaviour of the network ends up looking like when you do gossip adverts versus full multicast, especially with those intermittent events

When the client connects, it receives `Sync` message back from the node it responds to its `Connect` message. So the connecting node should be ready to actively participate in the network pretty fast.

But you are right, once the node is connected making sure it receives [at least] some adverts from wide spectrum of "geo" locations is important. I think there are few possible solutions to this:

* PR randomly selects a node in the network for the `Solicit` (a.k.a. full table sync) response, but we can also get each node to request full Sync periodically from the network (random node would be selected from the node's peer graph). This will basically be some sort of `push-pull` model (push = each node flushes its routes periodically, pull = each table actively requests sync from the network)
* another thing and this is something we have talked about for a while on an off is, we could pick N random nodes from each core region i.e. each core region would be gossiped to -- for this though we would need the network to be more "intelligent" i.e. geo-aware -- right now I don't know what's the best way to do that; I'd have to give it a bit of thought
* finally we could pick a random node for gossiping the `Advert` messages from each hop distance i.e. considering we keep `MaxDepth` peer graph we can pick N nodes from each hop to gossip the changes to

The convergence is a good question. I think with a lot of the latest changes we are getting closer to the SWIM protocol with some improvements, so the network should be converging eventually. 

SWIM: 
> ..complete state exchanges with a random node are done periodically over TCP, but much less often than gossip messages. The interval between full state exchanges is configurable or can be disabled entirely`

As for intermittent changes; I've some ideas but havent fleshed them out into code yet.

asim:
I think the push pull thing might be the first step.

Here's a real scenario to think about:

I start micro, connect to the network and type `micro run github.com/my/service`. My service is pushed to the platform, now I try to query it. Will I have a route to my service? If not, how am I going to query that? Or how am I going to know what's going on. 

If I am the origin of an event/route, maybe I should send to all? And if the route is not mine, I should gossip it? Meaning events are observed by directly connected peers, but the hop beyond is gossip based.

The workaround here is that a default route would outbound anything not already known so that you would probably be able to query it before having a route but its a hack in some ways.

network/node.go
```diff
@@ -36,6 +65,10 @@ type node struct {
 	network Network
 	// lastSeen keeps track of node lifetime and updates
 	lastSeen time.Time
+	// lastSync keeps track of node last sync request
+	lastSync time.Time
+	// err tracks node errors
+	err nodeError
```
asim: Should be a pointer

network/default.go
```diff
@@ -897,6 +963,80 @@ func (n *network) processNetChan(listener tunnel.Listener) {
...
+
+				log.Debugf("Network received sync message from: %s", pbNetSync.Peer.Node.Id)
+
+				peer := &node{
+					id:      pbNetSync.Peer.Node.Id,
+					address: pbNetSync.Peer.Node.Address,
+					link:    m.msg.Header["Micro-Link"],
+					peers:   make(map[string]*node),
+					status: &status{
+						err: &nerr{
```
asim: probably want a method to clean up creation of this status like `makeStatus(peer)`

network/default.go
```diff
@@ -1129,6 +1275,42 @@ func (n *network) manage() {
 					log.Debugf("Network failed deleting routes by %s: %v", route.Router, err)
 				}
 			}
+		case <-netsync.C:
+			// get a list of node peers
+			peers := n.Peers()
+			// pick a random peer from the list of peers and request full sync
+			if peer := n.node.GetPeerNode(peers[rnd.Intn(len(peers))].Id()); peer != nil {
```
asim: Is it cleaner to do `if peer == nil { continue }`

network/service/proto/network.proto
```diff
@@ -16,6 +16,8 @@ service Network {
         rpc Routes(RoutesRequest) returns (RoutesResponse) {};
 	// Returns a list of known services based on routes
 	rpc Services(ServicesRequest) returns (ServicesResponse) {};
+        // Status returns network status
+        rpc Status(StatusRequest) returns (StatusResponse) {};
```
asim, line 20: indentation issue

network/service/proto/network.proto
```diff
@@ -16,6 +16,8 @@ service Network {
         rpc Routes(RoutesRequest) returns (RoutesResponse) {};
 	// Returns a list of known services based on routes
 	rpc Services(ServicesRequest) returns (ServicesResponse) {};
```
asim: or indentation issue here maybe

network/default.go
milosgajdos: Yeah -- lots of this stuff needs to be cleaned up, not just this part 😅

PR #1107, opened 2020-01-12
Replace service prefix with FQDN style prefix

ucosty:
According to the k8s documentation, the label and annotation prefixes should be in the format of a FQDN, with dot separated labels of no more than 63 characters. The current label and annotation paramteres are rejected by the k8s api, most likely because they have two forward slashes in them.

### DIFF
--- a/registry/kubernetes/kubernetes.go
+++ b/registry/kubernetes/kubernetes.go
@@ -23,7 +23,7 @@ type kregistry struct {
 var (
 	// used on pods as labels & services to select
 	// eg: svcSelectorPrefix+"svc.name"
-	servicePrefix = "service.micro.mu/"
+	servicePrefix = "go.micro/"
 	serviceValue  = "service"
 
 	labelTypeKey   = "micro"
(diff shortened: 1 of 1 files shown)


crashdump:
Just to add the @ucosty, the error we are seeing is the following:

`Server register error: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Pod \"micro-api-7fcc79495-9rlwt\" is invalid: [metadata.labels: Invalid value: \"micro/service/go.micro.api\": a qualified name must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]') with an optional DNS subdomain prefix and '/' (e.g. 'example.com/MyName').`

Using the path above solves the issue.

asim:
Can you please change the prefix to go.micro?

ucosty:
@asim done

PR #1116, opened 2020-01-16
QueryStrategy to allow querying routes based on Advertising Strategy

milosgajdos:


router/query.go
```diff
@@ -52,15 +54,23 @@ func QueryRouter(r string) QueryOption {
...
-		Address: "*",
-		Gateway: "*",
-		Network: "*",
-		Router:  "*",
+		Service:  "*",
+		Address:  "*",
+		Gateway:  "*",
+		Network:  "*",
+		Router:   "*",
+		Strategy: AdvertiseAll,
```
asim, line 73: This should probably default to whatever the routes own strategy is so that in the event you do not specify you actually get back default e.g core nodes advertise best and a query returns best, edge nodes advertise local and this returns local.

milosgajdos: But how do we know that at this point? Router's strategy is set in Router's options. 

This function picks default values before it applies the value the user passes in.

asim: Guess you're right

PR #1119, opened 2020-01-16
Send only best routes via Sync. Only apply best routes.

milosgajdos:
We now send only the best routes over `Sync`.
When applying routes received in the `Sync` message we:
* prefer our own routes if the metric for a service we receive is the same as our best route for the same service
* we only add new routes if the received route metric is better than the metric of our best route

### DIFF
--- a/network/default.go
+++ b/network/default.go
@@ -854,7 +854,7 @@ func (n *network) processNetChan(listener tunnel.Listener) {
 					q := []router.QueryOption{
 						router.QueryStrategy(n.router.Options().Advertise),
 					}
-					routes, err := n.options.Router.Table().Query(q...)
+					routes, err := n.router.Table().Query(q...)
 					switch err {
 					case nil:
 						// encode the routes to protobuf
@@ -1051,7 +1051,7 @@ func (n *network) processNetChan(listener tunnel.Listener) {
 					// unmarshal the routes received from remote peer
 					route := pbUtil.ProtoToRoute(pbRoute)
 					// continue if we are the originator of the route
-					if route.Router == n.options.Router.Options().Id {
+					if route.Router == n.router.Options().Id {
 						log.Debugf("Network node %s skipping route addition: route already present", n.id)
 						continue
 					}
@@ -1061,7 +1061,7 @@ func (n *network) processNetChan(listener tunnel.Listener) {
 						router.QueryStrategy(n.router.Options().Advertise),
 					}
 
-					routes, err := n.options.Router.Table().Query(q...)
+					routes, err := n.router.Table().Query(q...)
 					if err != nil {
 						log.Debugf("Network node %s failed listing best routes for %s: %v", n.id, route.Service, err)
 						continue
@@ -1075,6 +1075,7 @@ func (n *network) processNetChan(listener tunnel.Listener) {
 							log.Debugf("Network node %s failed to add route: %v", n.id, err)
 							continue
 						}
+						continue
 					}
 
 					// find the best route for the given service
@@ -1381,7 +1382,7 @@ func (n *network) manage() {
 				q := []router.QueryOption{
 					router.QueryStrategy(n.router.Options().Advertise),
 				}
-				routes, err := n.options.Router.Table().Query(q...)
+				routes, err := n.router.Table().Query(q...)
 				switch err {
 				case nil:
 					// encode the routes to protobuf
--- a/network/default.go
+++ b/network/default.go
@@ -851,7 +851,9 @@ func (n *network) processNetChan(listener tunnel.Listener) {
 					}
 
 					// get a list of the best routes for each service in our routing table
-					q := []router.QueryOption{router.QueryStrategy(router.AdvertiseBest)}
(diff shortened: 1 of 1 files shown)


network/default.go
```diff
@@ -1045,9 +1046,51 @@ func (n *network) processNetChan(listener tunnel.Listener) {
 
 				// add all the routes we have received in the sync message
 				for _, pbRoute := range pbNetSync.Routes {
+					// unmarshal the routes received from remote peer
```
asim, line 1051: Skip route if its your own id

network/default.go
```diff
@@ -1045,9 +1046,51 @@ func (n *network) processNetChan(listener tunnel.Listener) {
...
 					route := pbUtil.ProtoToRoute(pbRoute)
-					if err := n.router.Table().Create(route); err != nil && err != router.ErrDuplicateRoute {
-						log.Debugf("Network node %s failed to add route: %v", n.id, err)
+					// lookup the best route for the given service in our routing table
+					q := []router.QueryOption{
+						router.QueryService(route.Service),
+						router.QueryStrategy(router.AdvertiseBest),
```
asim: Use advertising strategy already set in the router

network/default.go
```diff
@@ -1045,9 +1046,51 @@ func (n *network) processNetChan(listener tunnel.Listener) {
...
+						if err := n.router.Table().Create(route); err != nil && err != router.ErrDuplicateRoute {
+							log.Debugf("Network node %s failed to add route: %v", n.id, err)
+							continue
+						}
+					}
+
+					// Take the best route to given service and:
+					// * prefer our own routes if metric is the same
+					// * only add new routes if the metric is better than the metric of our best route
+					bestRoute := bestRoutes[0]
```
asim: if bestRoute.Metric <= route.Metric { continue }

network/default.go
```diff
@@ -1045,9 +1046,51 @@ func (n *network) processNetChan(listener tunnel.Listener) {
...
+							continue
+						}
+						// NOTE: we might want to skip here, too as we already have equally good route
+						// add routes to the routing table
+						if err := n.router.Table().Create(route); err != nil && err != router.ErrDuplicateRoute {
+							log.Debugf("Network node %s failed to add route: %v", n.id, err)
+							continue
+						}
+					}
+					if bestRoute.Metric > route.Metric {
```
asim: if condition not needed if above change is made

network/default.go
```diff
@@ -1328,8 +1371,9 @@ func (n *network) manage() {
 					Peer: node,
 				}
 
-				// get a list of all of our routes
-				routes, err := n.options.Router.Table().List()
+				// get a list of the best routes for each service in our routing table
+				q := []router.QueryOption{router.QueryStrategy(router.AdvertiseBest)}
```
asim: use router options strategy

network/default.go
milosgajdos: Fixing this now.

network/default.go
```diff
@@ -1058,28 +1058,37 @@ func (n *network) processNetChan(listener tunnel.Listener) {
 					// lookup best routes for the services in the just received route
 					q := []router.QueryOption{
 						router.QueryService(route.Service),
-						router.QueryStrategy(router.AdvertiseBest),
+						router.QueryStrategy(n.router.Options().Advertise),
 					}
-					bestRoutes, err := n.options.Router.Table().Query(q...)
+
+					routes, err := n.options.Router.Table().Query(q...)
```
asim: using n.router and n.options.Router in another place, should be consistent

network/default.go
```diff
@@ -1058,28 +1058,37 @@ func (n *network) processNetChan(listener tunnel.Listener) {
...
 					}
 
-					// we found no route for the given service
-					// create new route we have just received
-					if len(bestRoutes) == 0 {
+					// we found no routes for the given service
+					// create the new route we have just received
+					if len(routes) == 0 {
 						// add routes to the routing table
 						if err := n.router.Table().Create(route); err != nil && err != router.ErrDuplicateRoute {
```
asim, line 1073: If you're creating the route here should you just continue?

milosgajdos: That's the change I was thinking of when I messaged you in slack hah. Let me think...

network/default.go
milosgajdos: We should actually be more consistent with this in general not just in this piece of code 😅

network/default.go
```diff
@@ -1045,7 +1048,54 @@ func (n *network) processNetChan(listener tunnel.Listener) {
...
+						continue
+					}
+
+					// we found no routes for the given service
+					// create the new route we have just received
+					if len(routes) == 0 {
+						// add routes to the routing table
+						if err := n.router.Table().Create(route); err != nil && err != router.ErrDuplicateRoute {
+							log.Debugf("Network node %s failed to add route: %v", n.id, err)
+							continue
```
asim: This continue statement not needed if the below was added

milosgajdos: literally removed it right after you left

PR #1127, opened 2020-01-19
handle Loader.Load  return value error

Allenxuxu:


Allenxuxu:
It seems we have only just started using golangci-lint recently.
Do we already have a plan to fix static checking errors?

vtolstov:
Yes we plan to continue to fix linting errors

asim:
Currently we do not return error when initialising other interfaces. We should be consistent. Ideally we should change everything to return an error but it has broader impact on the surrounding tools.

Issue #1145, opened 2020-01-30
exponentialBackoff is not exponential function.


tpam28:
Hello everyone.
I experimented with retryfunc and I saw what exponentialBackoff function is not exponential function it’s power function when res = 10^n. And this func returns:

1. 10ms
2. 100ms
3. 1s
4. 10s
5. 1m40s
6. 16m40s
7. 2h46m40s
8. 27h46m40s
9. 277h46m40s

 
I can use custom function but I bewilder this fact. after 5 or 6 times we’ve never wait result. 
I think we need normal exponentialBackoff. Maybe we must use 0.1 * e^x  * time.Second for best result.
it retruns:

1. 100ms
2. 600ms
3. 1.9s
4. 4.3s
5. 7.9s
6. 13s
7. 19.8s
8. 28.5s
9. 39.2s

I'll be able to create PR if it is a good solution.

asim:
We welcome a PR!

tpam28:
Hello.
I was reading list of TODO in source code when I saw backoff.Do in util. This function contains same code as how exponentialBackoff had contained before I changed it.
Usage this function looks as kludge for me. 
for example:
sleep := backoff.Do(i)
time.Sleep(sleep)
if i > 5 {
  // reset attempts 5 == ~2mins
  i = 0
}

Maybe need we change this function as well as exponentialBackoff and use it inside  exponentialBackoff?
we’ll able to limit max return value to 2min if it need.
I’m sending PR tonight.

PR #1146, opened 2020-01-30
exponentialBackoff was changed from power function to exponential 

tpam28:
https://github.com/micro/go-micro/issues/1145


asim:
Thanks

PR #1147, opened 2020-01-30
Auth

ben-toogood:
Implement the Auth interface, with jwt and service implementations.


auth/auth.go
```diff
@@ -7,34 +7,42 @@ import (
...
-	Id string `json: "id"`
-	// Time of token creation
+// Role a service account has
+type Role struct {
+	Name     string
+	Resource *Resource
+}
+
+// ServiceAccount providers by an auth provider
+type ServiceAccount struct {
```
asim: Would just call this Account

auth/auth.go
```diff
@@ -7,34 +7,42 @@ import (
...
+	// The parent of the service account, e.g. a user
+	Parent *Resource `json: "parent"`
```
asim: Would maybe just call this Id? as in Account Id?

auth/auth.go
```diff
@@ -7,34 +7,42 @@ import (
...
-// Service is some thing to provide access to
-type Service struct {
-	// Name of the resource
-	Name string
-	// Endpoint is the specific endpoint
-	Endpoint string
+// Resource is an entity such as a user or service
+type Resource struct {
+	// Id of the resource
+	Id string
```
asim: Resource Name?

auth/auth.go
ben-toogood: We aren't storing the parent anywhere yet so what would the ID be?

ben-toogood: Actually, we could do type (user) and name (asim@micro.mu). What would we use the ID for?

asim: Every account should have an Id, whether thats defined as email or service name or whatever. And then maybe a "Root" or something in metadata to define an association to a different account.

auth/auth.go
```diff
@@ -9,39 +9,39 @@ import (
 type Auth interface {
 	// Init the auth package
 	Init(opts ...Option) error
-	// Generate a new auth ServiceAccount
-	Generate(*ServiceAccount) (*ServiceAccount, error)
-	// Revoke an authorization ServiceAccount
+	// Generate a new auth Account
+	Generate(*Account) (*Account, error)
```
asim: I would also change this to `Generate(id string, ...GenerateOption) (*Account, error)` and pass in Roles, Metadata, or whatever so that we prevent the Account Account stutter and have ability to extend.

asim: Example https://gist.github.com/asim/c149a22c1c433a9972cb72289368160c#file-auth-go-L3

ben-toogood: I agree, I'll make this change. Thanks

auth/auth.go
ben-toogood: Agree, i'll action this now

asim:
LGTM. You will just want to add `String()` method to the interface as discussed so you can check implementation then rebase and self merge!

PR #1148, opened 2020-01-30
fix test and description

tpam28:
I’m so sorry. I lied you because function in my PR is x^e instead of e^x which I wrote in issue. But I wrote in issue time interval for x^e. I think it works better than those function.
I fixed test and description  in function. test was wrong too. I don't know what happened to me.

### DIFF
--- a/client/backoff_test.go
+++ b/client/backoff_test.go
@@ -2,13 +2,12 @@ package client
 
 import (
 	"context"
-	"math"
 	"testing"
 	"time"
 )
 
 func TestBackoff(t *testing.T) {
-	delta := time.Duration(0)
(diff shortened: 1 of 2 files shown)


client/backoff_test.go
```diff
@@ -2,13 +2,12 @@ package client
...
 import (
 	"context"
-	"math"
 	"testing"
 	"time"
 )
 
 func TestBackoff(t *testing.T) {
-	delta := time.Duration(0)
+	results := []time.Duration{0 * time.Second, 100 * time.Millisecond, 600 * time.Millisecond, 1900 * time.Millisecond, 4300 * time.Millisecond, 7900 * time.Millisecond}
```
asim: Can you put this on multiple lines so its readable?

tpam28: Of course. A few minutes, please.

tpam28: 87753ad28953cc1635f734edaf61006f18442fe2

PR #1151, opened 2020-02-01
Fix race condition for debug logger

k1ng440:
Race conditions were found in debug logger through the Datadog trace wrapper test.

this pull request is for fixing the issue


vtolstov:
please, strip os std logger stuff that not fix logger races, commits for metadata copy is good and will be merged after that

PR #1153, opened 2020-02-02
New backoff

tpam28:
https://github.com/micro/go-micro/issues/1145#issuecomment-581017838

### DIFF
--- a/client/backoff.go
+++ b/client/backoff.go
@@ -2,7 +2,7 @@ package client
 
 import (
 	"context"
-	"math"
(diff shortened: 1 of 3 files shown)


client/backoff.go
```diff
@@ -2,13 +2,12 @@ package client
 
 import (
 	"context"
-	"math"
+	"github.com/micro/go-micro/v2/util/backoff"
```
asim: Please separate stdlib imports from external. See other files for example.

PR #1161, opened 2020-02-05
[WIP] logger first

printfcoder:


printfcoder:
there is a logrus test

```
// Package logrus allows the creation of Loggers based on logrus.StandardLogger().
package logrus

import (
	"github.com/go-log/log"
	"github.com/sirupsen/logrus"
)

type logger struct {
	*logrus.Logger
}

func (l *logger) Init(option ...log.Option) error {
	return nil
}

func (l *logger) Logfi(level log.Level, msg string, fields ...log.Field) {
	// todo: convert fields
	l.Logger.WithField(fields...).Log(convertToLogrusLevel(level), msg)
}

func (l *logger) SetLevel(level log.Level) {
	l.Logger.SetLevel(convertToLogrusLevel(level))
}

func (l *logger) Level() log.Level {
	return convertLevel(l.Logger.Level)
}

func (l *logger) String() string {
	return "logrus"
}

func (l *logger) Log(level log.Level, args ...interface{}) {
	l.Logger.Log(convertToLogrusLevel(level), args...)
}

func (l *logger) Logf(level log.Level, format string, args ...interface{}) {
	l.Logger.Logf(convertToLogrusLevel(level), format, args...)
}

// WithFields creates a new logger based on logrus.WithFields().
func WithFields(f logrus.Fields) log.Logger {
	l := &logger{logrus.WithFields(f).Logger}
	return l
}

// WithFields creates a new logger based on logrus.StandardLogger().
func New( /*options will be here for the custom features*/) log.Logger {
	// todo parse options

	l := logrus.StandardLogger()
	l.SetFormatter(&logrus.JSONFormatter{})
	l2 := &logger{l}
	l2.Init()
	return l2
}

func convertToLogrusLevel(level log.Level) logrus.Level {
	// todo: fix correct level map
	return logrus.Level(level)
}

func convertLevel(level logrus.Level) log.Level {
	// todo: fix correct level map
	return log.Level(level)
}

```

vtolstov:
Looks fine, but for Options i'm prefer Context option. So you can encapsulate any needed stuff into it and get inside own logger implementation

vtolstov:
p.s. for don't merge github allow to use WIP in pr name and protect to merge it as i know

printfcoder:
@vtolstov I think that context is used for communication among threads. The interface using option seems better. 
we need explicit options to tell people what features we have.

vtolstov:
Context is fine for optional stuff that needed for some logger and not need for another.
go-micro have context inside options for such stuff and its realy useful.

asim:
Point to existing example so he knows what you mean.

asim:
Otherwise I think it's a good start. I think Logfi is a bit weird and we might need to rename

asim:
Potentially have logger.WithFields method that then returns the Logger so that you can then continue to use Log/Logf

vtolstov:
+1 for asim suggestions

vtolstov:
@printfcoder as example https://pkg.go.dev/github.com/micro/go-micro/v2/server?tab=doc#Context and  https://pkg.go.dev/github.com/micro/go-micro/v2/server?tab=doc#Options

printfcoder:
@vtolstov yea, I know we have used Context for holding request info and sth else.  we can also use context to do so in Logger. but what things are necessary to be cached in Context for Logger? 
maybe we can just add context field in Logger for the future.

printfcoder:
@asim  also thought Logfi is weird. a better one?

asim:
Sorry was on my phone when looking at this stuff, here's your interface updated to do what I meant.

```go
// Logger is a generic logging interface
type Logger interface {
	Init(option ...Option) error
	// Level returns the logging level
	Level() Level
	// Log inserts a log entry.  Arguments may be handled in the manner
	Log(level Level, v ...interface{})
	// Logf insets a log entry.  Arguments are handled in the manner of
	Logf(level Level, format string, v ...interface{})
	// SetLevel updates the logging level.
	SetLevel(Level)
        // Set fields to always be logged
        WithFields(...Field) Logger
	// String return the name of logger
	String() string
}
```

You'd return a copy of the Logger with fields set when WithFields is called so someoen could do

```go
log.WithFields("some", "thing").Log("my stuff")
```

Either that or just `Fields(...Field) Logger`

printfcoder:
Why not just use the option to do so? so we can make Logger interface clearer.

```
type fieldsKey struct{}

func WithFields(fields ...Field) Option {
	return func(o *Options) {
		if o.Context == nil {
			o.Context = context.Background()
		}
		o.Context = context.WithValue(o.Context, fieldsKey{}, fields)
	}
}
```
Logger focuses on the top features. `Fields()` is just a sub-feature which can be assembled by params.

printfcoder:
@vtolstov updated to use Context.

vtolstov:
@printfcoder fields for different messages may be different, so then you pass it like option to logger, you must have many loggers. in case of @asim suggestion, you can use one logger, use Fields method to get another logger and use it method. Like logger.WithFields("aaa":"ttt").Log("message")

vtolstov:
about field - why not use map[string]interface{} ? it more powerful then create dedicated types. And underline logger can switch on passed type and provide needed internal type for that

printfcoder:
@vtolstov using `map` is enough for the current scenario. but potentially there would be some specific and custom field encoders like  `RFC3339-time` or ISOxxxx and any biz request for encoding any style what people want to do, eg. encrypting account info.

so if we use the map, it will be difficult to extend Field. the named field is easily converted to map, but conversely maybe not easily.

printfcoder:
think about the news Facebook logged the account plaintext password in their logFile.

logger/logger.go
```diff
@@ -0,0 +1,50 @@
...
+	Level() Level
+	// Log inserts a log entry.  Arguments may be handled in the manner
+	// of fmt.Print, but the underlying logger may also decide to handle
+	// them differently.
+	Log(level Level, v ...interface{})
+	// Logf insets a log entry.  Arguments are handled in the manner of
+	// fmt.Printf.
+	Logf(level Level, format string, v ...interface{})
+	// Logfi prints fields using JSON or KV.
+	Logfi(level Level, msg string, fields ...Field)
```
asim: I would remove Logfi since we have Fields. The idea being that when you want to log with fields you do log.Fields("foo", "bar").Log("my log message")

printfcoder: OK. thats better

vtolstov:
> @vtolstov using `map` is enough for the current scenario. but potentially there would be some specific and custom field encoders like `RFC3339-time` or ISOxxxx and any biz request for encoding any style what people want to do, eg. encrypting account info.
> 
> so if we use the map, it will be difficult to extend Field. the named field is easily converted to map, but conversely maybe not easily.

sorry, don't understand why map[string]interface{} is not suitable for such case
as i understand field name is always something like string in structure loggers

printfcoder:
@vtolstov did you mean 

```
type Field map[string]interface{}
```

1.  we can't extend it for its key easily, how can we encode every Field with custom encoders?
2. keys order in Map is not certain. we need the order logged  as people input

like:

```
type Encoder func(interface{}) string

type Field struct {
	Key     string
	Type    FieldType
	Value   interface{}
	Encoder Encoder
}

func (f *Field) GetValue() interface{} {
	if f.Encoder != nil {
		return f.Encoder(f.Value)
	}
	return f.Value
}

func Pwd(key string, val string) Field {
	return Field{Key: key, Type: Pwd, Value: val, Encoder: func(i interface{}) string {
		return encode(i)
	}}
}
```

so we can use like this

```
log.Fields(log.Pwd("someKey", "anyValue"),).log("anyMsg")
```

but now I am not sure where the Encoder should be in, Field or FieldType.

vtolstov:
I think that we can merge it as-is now. And try to implement zap/logrus/zerolog implementations in go plugins. After  that we can easy modify interface to our demands. After that modify go-micro codebase to use this logger

PR #1163, opened 2020-02-05
Runtime Retries Limit

ben-toogood:
Implement max 3 retries for local and k8s runtimes.


asim:
Any reason its implemented as a field in the service rather than passed as an option and maybe returned back in metadata like status and error?

asim:
Actually you either want retries as a CreateOption or in metadata. The Service struct should remain clean with only the service details itself.

ben-toogood:
Thanks @asim. I'll amend this.

ben-toogood:
@asim I've actioned this, just need to implement the micro side. Can you pls review.

runtime/options.go
```diff
@@ -37,6 +39,13 @@ func WithType(t string) Option {
 	}
 }
 
+// WithMaxRetries sets the arg MaxRetries
+func WithMaxRetries(max int) Option {
```
asim: Just call with WithRetries and the option as Retries

ben-toogood:
@asim updated based on feedback, I agree this is better

runtime/options.go
```diff
@@ -87,6 +80,13 @@ func WithCommand(args ...string) CreateOption {
 	}
 }
 
+// WithRetires sets the max retries attemps
+func WithRetires(retries int) CreateOption {
```
asim: Retires => Retries

runtime/service.go
```diff
@@ -58,9 +58,10 @@ func newService(s *Service, c CreateOptions) *service {
 			Env:  c.Env,
 			Args: args,
 		},
-		closed:  make(chan bool),
-		output:  c.Output,
-		updated: time.Now(),
+		closed:     make(chan bool),
+		output:     c.Output,
+		updated:    time.Now(),
+		maxRetries: c.Retries,
```
asim, line 64: retries

PR #1167, opened 2020-02-05
pass micro errors from grpc server to grpc client

vtolstov:


vtolstov:
server/grpc and client/grpc must be fixed to use proto.Message based error

asim:
This is quite clever

asim:
Should we merge?

vtolstov:
I think so. If we have another issues with this, we can fix later

PR #1169, opened 2020-02-06
Runtime Custom Source (Part 2)

ben-toogood:


runtime/options.go
```diff
@@ -51,8 +51,8 @@ type CreateOptions struct {
 	Output io.Writer
 	// Type of service to create
 	Type string
-	// Base image for docker
-	BaseImage string
+	// Source of the code
+	Source string
```
asim: Source is already in the Service when its passed in

ben-toogood:
@asim let me know if you're happy with this and i'll squash and merge. It's working now in k8s.

asim:
LGTM. We may come across issues that require further changes later but if it works go ahead and squash merge

PR #1170, opened 2020-02-06
Tracing: traces now correctly form a tree

crufter:
To support https://github.com/micro/platform/pull/14

### DIFF
--- a/debug/trace/memory/memory.go
+++ b/debug/trace/memory/memory.go
@@ -84,7 +84,7 @@ func NewTracer(opts ...trace.Option) trace.Tracer {
 
 	return &Tracer{
 		opts: options,
-		// the last 64 requests
-		buffer: ring.New(640),
+		// the last 256 requests
+		buffer: ring.New(256),
 	}
 }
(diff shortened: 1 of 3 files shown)


asim:
Self merge when you are ready!

debug/trace/memory/memory.go
```diff
@@ -85,6 +85,6 @@ func NewTracer(opts ...trace.Option) trace.Tracer {
 	return &Tracer{
 		opts: options,
 		// the last 64 requests
-		buffer: ring.New(64),
+		buffer: ring.New(640),
```
crufter: This might be a bit controversial. Also need to fix comments.

asim: That is definitely controversial. Max I'd do is 256

debug/trace/trace.go
```diff
@@ -34,17 +35,32 @@ type Span struct {
...
+	parentSpanID, ok := metadata.Get(ctx, spanIDKey)
+	return traceID, parentSpanID, ok
 }
 
-// NewContext creates a new context with the span
-func NewContext(ctx context.Context, s *Span) context.Context {
-	return context.WithValue(ctx, spanKey{}, s)
+// ToContext saves the trace and span ids in the context
+func ToContext(ctx context.Context, traceID, parentSpanID string) context.Context {
+	return metadata.MergeContext(ctx, map[string]string{
```
jakexks, line 60: Does this work if ctx is nil?

crufter: Good catch.

PR #1173, opened 2020-02-07
Fix Local Runtime Default Command

ben-toogood:


ben-toogood:
@asim I implemented Auth.Validate instead of Auth.Login because the function isn't yet doing "login", but instead just checking the token which will be provided by the web platform. We can extend the functionality to include login at a later stage. 

Let me know if you would approach this in a different way

PR #1174, opened 2020-02-07
Auth Wrapper

ben-toogood:


service.go
```diff
@@ -38,10 +39,14 @@ func newService(opts ...Option) Service {
 	options.Client = wrapper.FromService(serviceName, options.Client)
 	options.Client = wrapper.TraceCall(serviceName, trace.DefaultTracer, options.Client)
 
+	// parse command line flags
+	cmd.Init()
```
ben-toogood: @asim I needed to do cmd.Init so that the auth package would be initialised with the necessary flags, e.g. excluded endpoints, public key etc. I've tested this and it's working well, but it feels like this shouldn't be here. Can you think of any potential problems I may have missed?

auth/options.go
```diff
@@ -7,10 +7,18 @@ import (
 type Options struct {
 	PublicKey  []byte
 	PrivateKey []byte
+	Excludes   []string
 }
 
 type Option func(o *Options)
 
+// Excludes endpoints from auth
+func Excludes(excludes []string) Option {
```
asim:
```
func Exclude(eps ...string) Option {
```

Can then pass `auth.Exclude("Foo.Bar", "Baz.Bar")`

config/cmd/cmd.go
```diff
@@ -254,6 +254,11 @@ var (
 			EnvVars: []string{"MICRO_AUTH_PRIVATE_KEY"},
 			Usage:   "Private key for JWT auth (base64 encoded PEM)",
 		},
+		&cli.StringSliceFlag{
+			Name:    "auth_excludes",
```
asim: auth_exclude

util/wrapper/wrapper.go
```diff
@@ -126,3 +130,51 @@ func TraceHandler(t trace.Tracer) server.HandlerWrapper {
...
+					return h(ctx, req, rsp)
+				}
+			}
+
+			// Extract the Authorization header
+			md, ok := metadata.FromContext(ctx)
+			if !ok {
+				md = metadata.Metadata{}
+			}
+			authHeader := md["Authorization"]
```
asim: call metadata.Get(ctx, "Authorization") as if this passes through http and http2 it will go through various case formats. Either that or metadata.FromContext needs to camel case keys.

service.go
asim: You cannot do this here. Flags are parsed when calling cmd.Init. This will violate the users expectation. When we do NewService our expectation is that this is just a clean simple initialisation which anyone can use without the flag parsing. If you do `service.Init` there is the knowledge that this will parse flags. It will also barf in the runtime anywhere we've used NewService. Since that execution occurs inside an existing cli parser.

It doesn't look like the auth wrapper needs those things outside of the interface so the assumption being when service.Init is called these things will be initialisation by config/cmd in the default auth interface. In the micro runtime this would end up being preinitialised since its already done that cmd.Init call.

ben-toogood: I thought this also, but when I tested it the auth was noop, despite adding the auth flag. I checked this again when the wrapper was called upon request, and the same was true then.

config/cmd/cmd.go
ben-toogood: Other slice CLI flags used plurals for var names, but I'm happy to amend

service.go
asim: Fixed. Unfortunately I can now remember why this problem occurs. We've moved from setting defaults to setting the values in the service itself. cmd.Init sets defaults and service.Init sets the values in the service itself.

I've fixed by creating a lazy loading function that returns the auth interface from service.options. It probably pushes us to clearly define what the expected behaviour is. Ideally service.Init should only set its only values rather than defaults in different packages and we should make use of our own internal values set in the options.

ben-toogood:
@asim thanks for fixing the `cmd.Init` bug. I've tested this morning and all is looking good now, can you please let me know if you're happy to merge.

asim:
LGTM will merge

PR #1187, opened 2020-02-11
various fixes for broker and messaging in server

vtolstov:


vtolstov:
* finally fix issues with eats broker disconnect/connect
* logs about broker disconnect in server

PR #1193, opened 2020-02-13
add common logger field type

printfcoder:


asim:
Is there a reason why we need this? It seems incredibly verbose and not inline with most of how go-micro interfaces behave in being lean and straight forward. It feels like we need to strip that out as part of the move to map[string]interface{} for field metadata.

printfcoder:
map[string]interface{} is fine, but we will probably lose expansibility of Field like `Encode`. 
so we should take the balance between easy for use and fast for decode and logging.

zap uses explicit Field for logging fast. because Field needs less reflection. I think we need this. 

if just for log things in a simple way, map[string]interface is better.

asim:
We are offloading the actual implementation details to zap and other things. We don't really need to think about performance here. Its never the primary concern of micro to be thinking about performance. The first concern is always simplicity. And extensibility should then be done in a way that does not create this level of verbosity.

asim:
Need to get @vtolstov thoughts. Think it would be better to strip it down.

vtolstov:
I think that we need to have map[string]interface{} , I'm check zap.Any it have simple switch over v.(type) that in go is fast.
So simplicity is better.

PR #1198, opened 2020-02-15
[W.I.P] refactor(logger): logger fields changed to map[string]interface{}

xmlking:
Logger fields changed to map[string]interface{}


asim:
This change includes a grpc listener change. Please remove as the PR is related to the logger.

PR #1210, opened 2020-02-17
Provide a callback to an RPC service

TriAnMan:
This callback allows to fix an infinite loop if NATS broker goes out.
Other handling routines could be also provided by a user of an RPC service.

### DIFF
--- a/server/options.go
+++ b/server/options.go
@@ -46,6 +46,8 @@ type Options struct {
 	// Other options for implementations of the interface
 	// can be stored in a context
 	Context context.Context
+
+	RPCErrorCB func(Server, error)
 }
 
 func newOptions(opt ...Option) Options {
--- a/server/rpc_server.go
+++ b/server/rpc_server.go
@@ -804,6 +804,11 @@ func (s *rpcServer) Start() error {
 			// check the error and backoff
 			default:
 				if err != nil {
+					if s.opts.RPCErrorCB != nil {
+						s.opts.RPCErrorCB(s, err)
+						continue
+					}
+
 					log.Logf("Accept error: %v", err)
 					time.Sleep(time.Second)
 					continue


vtolstov:
where we can find issue about problem that you try to solve?

TriAnMan:
> where we can find issue about problem that you try to solve?

I'm a newcomer in your community. I can post an issue, I guess. Or maybe I can provide a unit test for issue I'm solving, thus it is possible to use NATS in tests programmatically.

TriAnMan:
I also found no contribution guidelines.

vtolstov:
If you write description of problems and to reproduce it, we can solve it.
Mostly we dont want to expose additional options.

TriAnMan:
Have such command:
```
package main

import (
	"fmt"
	"github.com/micro/go-micro/server"
	"github.com/micro/go-micro/transport"
	"github.com/micro/go-plugins/transport/nats"
	natsIO "github.com/nats-io/nats.go"
	"os"
)

func main() {
	natsOptions := natsIO.GetDefaultOptions()
	natsOptions.AllowReconnect = false // or wait for 2 minutes

	natsTransport := nats.NewTransport(
		transport.Addrs("192.168.56.101:4222"),
		nats.Options(natsOptions),
	)

	serverOptions := []server.Option{
		server.Address(":12345"),
		server.Transport(natsTransport),
	}
	rpcServer := server.NewServer(serverOptions...)

	if err := rpcServer.Start(); err != nil {
		fmt.Printf("%T, %s", err, err)
		os.Exit(1)
	}

	<-make(chan os.Signal, 1)
}
```

Launch NATS on 192.168.56.101:4222
Launch command.
Terminate NATS.
You will get an infinite loop like:
```
2020-02-17 20:21:25.745606 I | Transport [nats] Listening on :12345
2020-02-17 20:21:25.748529 I | Broker [http] Connected to [::]:53471
2020-02-17 20:21:25.885129 I | Registry [mdns] Registering node: go.micro.server-27d46f63-834b-40be-a0f9-1c9348c72593
2020-02-17 20:21:36.347707 I | Accept error: nats: connection closed
2020-02-17 20:21:37.347832 I | Accept error: nats: connection closed
2020-02-17 20:21:38.347956 I | Accept error: nats: connection closed
2020-02-17 20:21:39.348053 I | Accept error: nats: connection closed
2020-02-17 20:21:40.348130 I | Accept error: nats: connection closed
.......
.......
```

TriAnMan:
And there is no way to escape this loop by default.
I mean no way to know that NATS broker is down, thus you can't call `rpcServer.Stop()`.

TriAnMan:
Or even throw a `panic()`

vtolstov:
If you bring up nats after that, does it reconnects?

пн, 17 февр. 2020 г., 20:32 Anton Andersen <notifications@github.com>:

TriAnMan:
> If you bring up nats after that, does it reconnects?

No. NATS transport gives up reconnection after `natsOptions.MaxReconnect` attempts.

TriAnMan:
So when you see `Accept error: nats: connection closed` then nothing you can do.

TriAnMan:
To be clear. NATS tries to reconnect to the broker before you get `Accept error: nats: connection closed`. And if you get this message that means that NATS gives up to reconnect.

vtolstov:
But if max reconnects is negative, i think it must try infinitive?

TriAnMan:
But you will also get an infinite loop and you aren't able to know if anything is wrong.

TriAnMan:
Thus you can't make a fail-fast service with go-micro.

vtolstov:
ah, i see only now - why you try to use nats as transport? This is not cool - for each message you need to send data and receive confirmation and all via nats. Why not use more traditional transport and use nats as pub/sub?

vtolstov:
P.S. For pub/sub broker you can use own nats config with all needed callbacks.

TriAnMan:
> ah, i see only now - why you try to use nats as transport? This is not cool - for each message you need to send data and receive confirmation and all via nats.

We have already a microservice constellation (actually lots of them) and we are fine with NATS. Also I guess that it isn't very good solution to change whole infrastructure instead of a couple lines of code. Especially if that code makes go-micro more safe to use and provides more control to users of rpc_server.

TriAnMan:
Anyway this PR is backward compatible and does not change the behavior of rpc_server if RPCErrorCB isn't set.

asim:
For every change we make to the API or options we leak things and increase the surface area. The problem should likely be solved by the NATS transport rather than the level up. Either that or we should completely bail out. See http transport for its logic.

Otherwise the addition has to be carefully reviewed and likely refactored to be more broadly applicable. If you look at the other options this doesn't really fit. It's tacked on. 

I agree with Vasiliy that it should likely be a NATS callback and in the NATS plugin itself.

vtolstov:
I think that this can be transport option for nats transport, for example pass nats config (preferred way)
So if you want smart way - use own config and get full control.

asim:
I agree with Vasiliy. Please open a PR against go-plugins NATS transport to add a callback

PR #1212, opened 2020-02-18
add missing option to client.NewMessage

vtolstov:


asim:
I think reason I didn't add it was specifically because of that long name lol

PR #1213, opened 2020-02-18
fixes for safe conversation and avoid panics

vtolstov:

### DIFF
--- a/api/handler/broker/broker.go
+++ b/api/handler/broker/broker.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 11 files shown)


vtolstov:
may be incomplete, but mostly do all needed checks

vtolstov:
@asim what you think about usage of sync/atomic incase of sync.Once ?
if no objection - i can merge it?

asim:
My only issue with atomic is code simplicity and clarity. It adds complexity because it's not something developers use on a day to day basis. So wherever you're making these changes know that you're becoming the maintainer.

vtolstov:
No problem. for me atomic is simpler then create channel to return error.

PR #1216, opened 2020-02-19
logger with helper methods

xmlking:


logger/logger.go
```diff
@@ -3,54 +3,27 @@ package logger
...
-	// fmt.Printf.
-	Logf(level Level, format string, v ...interface{})
-	// Fields set fields to always be logged
-	Fields(fields map[string]interface{}) Logger
-	// Error set `error` field to be logged
-	Error(err error) Logger
 	// SetLevel updates the logging level.
 	SetLevel(Level)
 	// String returns the name of logger
 	String() string
```
asim, line 24: Please leave String as the last method in the interface

logger/logger.go
```diff
@@ -3,54 +3,27 @@ package logger
...
+	// log at given level with message, fmtArgs and context fields
+	Log(level Level, template string, fmtArgs []interface{}, fields Fields)
```
asim: Is there a particular reason we moved away from the old interface? This is really not as clean as the previous implementation was.

logger/basic.go
```diff
@@ -0,0 +1,120 @@
...
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"log"
+	"os"
+)
+
+type basicLogger struct {
```
asim: Please rename this to defaultLogger and the file to default.go

logger/logger.go
```diff
@@ -71,3 +44,84 @@ func GetLevel(levelStr string) (Level, error) {
...
+	globalLogger.SetLevel(lvl)
+}
+
+func Info(args ...interface{}) {
+	globalLogger.Log(InfoLevel, "", args, nil)
+}
+func Infof(template string, args ...interface{}) {
+	globalLogger.Log(InfoLevel, template, args, nil)
+}
+func Infow(msg string, fields Fields) {
```
asim: Do we need the "*w" methods?

logger/logger.go
```diff
@@ -71,3 +44,84 @@ func GetLevel(levelStr string) (Level, error) {
...
+}
+
+func SetGlobalLevel(lvl Level) {
+	globalLogger.SetLevel(lvl)
+}
+
+func Info(args ...interface{}) {
+	globalLogger.Log(InfoLevel, "", args, nil)
+}
+func Infof(template string, args ...interface{}) {
```
asim: Is template the right word? Should this nor be format?

logger/logger.go
```diff
@@ -71,3 +44,84 @@ func GetLevel(levelStr string) (Level, error) {
...
+	globalLogger.Log(ErrorLevel, "", args, nil)
+}
+func Errorf(template string, args ...interface{}) {
+	globalLogger.Log(ErrorLevel, template, args, nil)
+}
+func Errorw(msg string, err error) {
+	globalLogger.Error(ErrorLevel, msg, nil, err)
+}
+
+func Panic(args ...interface{}) {
```
asim: We don't need Panic methods if we have Fatal.

logger/options.go
```diff
@@ -2,20 +2,38 @@ package logger
...
-//       xxx:
 type Option func(*Options)
 
 type Options struct {
 	Context context.Context
 }
+
+type fieldsKey struct{}
+
+func WithFields(fields Fields) Option {
```
asim: Please add comments to all functions

logger/logger.go
xmlking: will do

logger/logger.go
xmlking: well, the `Fields` has to be added per log event.
if you set `Fields` in logger , Then it is  considered as `sub logger` which is reused for all subsequent logging events. that may not what developer want. 
there is also need for creating sub loggers (example developer want add `category=database` field for all logs form `data access layer`)

logger/logger.go
xmlking: what would you propose? `format` ?

logger/logger.go
xmlking: we offer `PanicLevel` .  how dev can use it without this method?

PR #1218, opened 2020-02-19
implement flow interface

vtolstov:


asim:
Still finding time to review. Likely tomorrow. First use of flow might end up being an end-to-end test service where we can register flows that are executed for product/service lifecycle testing

flow/flow.go
```diff
@@ -0,0 +1,303 @@
...
+	// Concurrency specify count of workers create for nodes in flow
+	Concurrency int
+	// Retries specify count of retries in case of node execution failed
+	Retries int
+	// Context is used for storing non default options
+	Context context.Context
+	// Client for communication
+	Client client.Client
+	// Broker for communication
+	Broker broker.Broker
```
asim: If you have the Client you don't need the broker because the Client contains the broker.

flow/flow.go
```diff
@@ -0,0 +1,303 @@
...
+	// Number of worker goroutines
+	Concurrency int
+	// Preallocate worker goroutines
+	Prealloc bool
+	// If no workers available no blocking
+	Nonblock bool
+	// Wait completiong before stop
+	Wait bool
+	// StateStore is used for flow state marking
+	StateStore DataStore
```
asim: State/Data/Flow stores seem redundant. We should only have to pass in one store.

flow/flow.go
```diff
@@ -0,0 +1,303 @@
...
+	// DataStore is used for intermediate data passed between flow nodes
+	DataStore DataStore
+	// FlowStore is used for storing flows
+	FlowStore FlowStore
+	// EventHandler is used to notification about flow progress
+	EventHandler EventHandler
+	// PanicHandler is used for recovery panics
+	PanicHandler func(interface{})
```
asim: ErrorHandler?

flow/flow.go
```diff
@@ -0,0 +1,303 @@
...
+	Output string
+	// Steps that are required for this step
+	After []string
+	// Steps for which this step required
+	Before []string
+	// Step operation to execute in case of error
+	Fallback Operation
+}
+
+func (s *Step) Name() string {
```
asim: Are Name/Id/String redundant methods?

flow/flow.go
```diff
@@ -0,0 +1,303 @@
...
+	// Timeout for step
+	Timeout int
+	// Step operation to execute
+	Operation Operation
+	// Which step use as input
+	Input string
+	// Where to place output
+	Output string
+	// Steps that are required for this step
+	After []string
```
asim: Should a Step know steps before and after or is this the point of the flow? Or is this provided by the flow once its registered a step? I would probably put []*Step if its going to stay so its a double linked list in some ways.

flow/flow.go
```diff
@@ -0,0 +1,303 @@
...
+	// name of step
+	ID string
+	// Retry count for step
+	Retry int
+	// Timeout for step
+	Timeout int
+	// Step operation to execute
+	Operation Operation
+	// Which step use as input
+	Input string
```
asim: What is an input and output?

flow/flow.go
```diff
@@ -0,0 +1,303 @@
...
+	Result(flow string, reqID string, step *Step) ([]byte, error)
+	// Stop executor and drain active workers
+	Stop() error
+}
+
+type Step struct {
+	// name of step
+	ID string
+	// Retry count for step
+	Retry int
```
asim: Retry/Timeout feel like options?

flow/flow.go
```diff
@@ -0,0 +1,303 @@
...
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/micro/go-micro/v2/broker"
+	"github.com/micro/go-micro/v2/client"
+	"github.com/micro/go-micro/v2/logger"
+)
+
+type Flow interface {
```
asim, line 16: I'm still looking at this interface. Spending time reading about workflow and trying to understand how it should work.

I had previously imagined a "flow" as one thing which would represent a state machine that is executed. In this case it is something of a system that manages many workflows much like a registry, database or broker.

I think you could probably split out Execute/Resume/Pause/Abort/State/Result/Stop into an Executor interface that's then embedded

```
type Flow interface {
    // embedded executor
    Executor
    Init()
    Options()
    String()
}

type Executor interface {
   Execute()
   ...
}
```

asim: You may want to define State as a type which is then returned when checking State so you can provide the sequence of events.

asim: Then I think

```
type Flow interface {
   // add a new step
   Add(flow string, step *Step) error
   // lookup a flow
   Lookup(flow string) (*Step, error)
   // update applies against the same step id
   Update(flow string, step *Step) error
   // remove a step
   Remove(flow string, step *Step) error
}
```

asim: Alternatively

```
type Flow interface {
   // creates a new flow
   Create(id string) (Task, error)
   // register a task
   Register(id string, Task) error
   // execute a flow
   Execute(id string ...) error
}

task.Add(Step)
task.Execute()

File: flow/flow.go (hunk 1 above)
Comment (vtolstov) [reply flow/flow.go]: removed from options, thanks

File: flow/flow.go (hunk 2 above)
Comment (vtolstov) [reply flow/flow.go]: no, i'm not add docs but this is must be different stores. 
Data - is []byte blob, and can be large, so it useful to store sometimes in s3 capable storage.
State - is []byte small blob, must be consistent and in-sync so its preferable distributed db or etcd or something.
Flow - is variable sized []byte blob that can be stored in db or in k/v or something. Mostly fast read store with ability to expose Timestamp for effective caching and transaction update support.

File: flow/flow.go (hunk 3 above)
Comment (vtolstov) [reply flow/flow.go]: yes, makes sense

File: flow/flow.go (hunk 4 above)
Comment (vtolstov) [reply flow/flow.go]: Mostly yes, but it needs for DAG, all dag implementations wants Name or Id do vertex naming

File: flow/flow.go (hunk 5 above)
Comment (vtolstov) [reply flow/flow.go]: Yes, each step may known after which step it must run. If you provide full flow you already have linked list of steps. But if you need to have ability to insert flow, you need to says before/after it must run.

File: flow/flow.go (hunk 6 above)
Comment (vtolstov) [reply flow/flow.go]: Its step names where put/get data in data store.
So if steps not specify input, we thinks that all steps get the same input that passed via execute to flow. If steps have different input - it can get other steps output as input.

File: flow/flow.go (hunk 7 above)
Comment (vtolstov) [reply flow/flow.go]: For some steps you want to have multiple retries, for some - only one.

File: flow/flow.go (hunk 8 above)
Comment (vtolstov) [reply flow/flow.go:16]: About Executor i'm fine to exclude it. But for flow interface i need to think more...

File: flow/flow.go (hunk 2 above)
Comment (asim) [reply flow/flow.go]: Doesn't feel like the right way to impement it. Rather these are concerns taken care of by the implementation of the flow. The default flow being "in memory" perhaps and so its a non-concern. Then you might have an implementation that uses "aws" and so on.

File: flow/flow.go (hunk 4 above)
Comment (asim) [reply flow/flow.go]: But if its all the same then surely you can just call step.ID?

File: flow/flow.go (hunk 6 above)
Comment (asim) [reply flow/flow.go]: Its unclear even still. I have zero idea what to put in the Input/Output. The string values are not useful unless I know how to use them, especially for something so opaque.

File: flow/flow.go (hunk 7 above)
Comment (asim) [reply flow/flow.go]: They should be options to the Create/Add method

File: flow/flow.go (hunk 8 above)
Comment (vtolstov) [reply flow/flow.go:16]: in case of embeding executor, how to deap that executor can have init and options methods ?

**Comment by vtolstov** [comment]
May be Flow does not need to Inherit Executor But provide Executor() as method, so after that Executor can be used to control execution?

File: flow/flow.go (hunk 5 above)
Comment (vtolstov) [reply flow/flow.go]: I can convert After/Before from []string to []*Step, but internally flow uses only ID of steps to create vertices and connect it

File: flow/flow.go (hunk 6 above)
Comment (vtolstov) [reply flow/flow.go]: So i can change it to *Step , but from it flow uses only id to determine store key to retrieve data

**Comment by asim** [comment]
Try it, see if it works

**Comment by asim** [comment]
Otherwise Executor could just be passed as an option and Flow only has a Run() method

File: flow/flow.go (hunk 2 above)
Comment (vtolstov) [reply flow/flow.go]: i think mostly for implementation not like change store, but about all logic inside - like pool of workers, parsing steps definition from store and so. For production mostly in memory not suitable - as you can be interrupted in the middle of execution, and you need continue/rollback laer (or via HA flow executor)

File: flow/flow.go (hunk 4 above)
Comment (vtolstov) [reply flow/flow.go]: main problem that terraform dag uses to name vertex via ID(), but more powerful in case of speed another dag implementation uses Name(). So this code exists for ability to easy switch between implementations. I need cleanup unneeded now, i think

File: flow/flow.go (hunk 7 above)
Comment (vtolstov) [reply flow/flow.go]: in this case when i'm execute step i need to know this variables and if i don't put it into struct, i need to use context. I don't this that adding context to Step struct is really needed

File: flow/flow.go (hunk 8 above)
Comment (vtolstov) [reply flow/flow.go:16]: > Otherwise Executor could just be passed as an option and Flow only has a Run() method

interesting case

**Comment by vtolstov** [comment]
Update with fixed FlowExecuteOperation to provide ability to jump to another flow on failure, or rollback.

**Comment by asim** [comment]
The Flow interface definitely needs to shrink or be consolidated. Its doing too much in the one thing. Its confusing as a user. I need an example of usage.

**Comment by vtolstov** [comment]
Example of usage in tests. Does it enough or i need to add more?

**Comment by vtolstov** [comment]
> Otherwise Executor could just be passed as an option and Flow only has a Run() method

Looks good, but what about Pause/Resume/Stop/Restart - is that Flow thing or Executor? I think that executor.

looks like executor must be like this
```
type Executor interface {                                                                                  
  // Init executor with options                                                                                
  Init(...ExecutorOption) error                                                                                    
  // Get executor options                                                                                      
  Options() ExecutorOptions                                                                                        
  Execute(flow string, req interface{}, rsp interface{}, opts ...ExecuteOption) (string, error)                
  // Resume specific paused flow execution by request id                                                   
  Resume(flow string, reqID string) error                                                                  
  // Pause specific flow execution by request id                                                           
  Pause(flow string, reqID string) error                                                                   
  // Abort specific flow execution by request id                                                           
  Abort(flow string, reqID string) error                                                                   
  // Status show status specific flow execution by request id                                              
  Status(flow string, reqID string) (Status, error)                                                        
  // Result get result of the flow step                                                                    
  Result(flow string, reqID string, step *Step) ([]byte, error)                                            
  // Stop executor and drain active workers                                                                
  Stop() error                                                                                             
}                                                                                                          
```

vtolstov:
and we strip this methods from flow

vtolstov:
@asim i'm split logically to files, next is fix executor. and i think that for first step its fine

asim:
Reopen later with just the interface to start with.

PR #1219, opened 2020-02-19
Runtime Events (Deploy after build completes)

ben-toogood:


asim:
Let's close this for now in favour of pushing an event to the web UI after the build is complete and allowing it to progress the user flow. We will leave the CLI as is. John indicated that the micro run command will execute and once the image is updated it will eventually run so this not really an issue. for MVP we just want to provde feedback in the UI when creating a new service. Separately we just need a way of doing `micro update service` so that the same deployment is kicked or is perhaps updated based on the runtime scheduler seeing new builds.

PR #1227, opened 2020-02-20
encode metadata to safe transmit across grpc

vtolstov:

### DIFF
--- a/metadata/metadata.go
+++ b/metadata/metadata.go
@@ -56,7 +56,6 @@ func FromContext(ctx context.Context) (Metadata, bool) {
 	for k, v := range md {
 		key, err := base32.StdEncoding.DecodeString(k)
 		if err != nil {
-			panic(err)
 			return nil, false
 		}
 		newMD[string(key)] = v
--- a/metadata/metadata.go
+++ b/metadata/metadata.go
@@ -57,7 +57,8 @@ func FromContext(ctx context.Context) (Metadata, bool) {
 	for k, v := range md {
 		key, err := base32.StdEncoding.DecodeString(strings.ToUpper(k))
 		if err != nil {
-			return nil, false
+			// try to be compatible with older micro versions
+			newMD[strings.Title(k)] = v
 		}
 		newMD[string(key)] = v
 	}
--- a/metadata/metadata.go
+++ b/metadata/metadata.go
@@ -4,8 +4,6 @@ package metadata
 import (
 	"context"
 	"encoding/base32"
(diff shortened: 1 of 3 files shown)


metadata/metadata.go
```diff
@@ -47,18 +51,30 @@ func FromContext(ctx context.Context) (Metadata, bool) {
...
 
-	// capitalise all values
-	newMD := make(map[string]string)
+	newMD := make(Metadata)
+
 	for k, v := range md {
-		newMD[strings.Title(k)] = v
+		key, err := base32.StdEncoding.DecodeString(k)
+		if err != nil {
+			panic(err)
```
asim: Should you panic?

vtolstov: sorry, debug stuff that never can happen

metadata/metadata.go
```diff
@@ -47,18 +51,29 @@ func FromContext(ctx context.Context) (Metadata, bool) {
...
 
-	// capitalise all values
-	newMD := make(map[string]string)
+	newMD := make(Metadata)
+
 	for k, v := range md {
-		newMD[strings.Title(k)] = v
+		key, err := base32.StdEncoding.DecodeString(k)
+		if err != nil {
+			return nil, false
```
asim: Should you just continue? What if we're trying to be backwards compatible and there are no base32 encoded strings

vtolstov: inside context we always have encoded metadata

vtolstov: or you mean if we have older micro aff and never?

vtolstov: push fix

metadata/metadata.go
```diff
@@ -47,18 +52,30 @@ func FromContext(ctx context.Context) (Metadata, bool) {
...
-	// capitalise all values
-	newMD := make(map[string]string)
+	newMD := make(Metadata)
+
 	for k, v := range md {
-		newMD[strings.Title(k)] = v
+		key, err := base32.StdEncoding.DecodeString(strings.ToUpper(k))
+		if err != nil {
+			// try to be compatible with older micro versions
+			newMD[strings.Title(k)] = v
```
asim, line 61: You're not continuing here so it will do this and the thing below

vtolstov: inattentive today , thanks

asim:
We will leave for now as this breaks backwards compatibility with existing services.

Encode/Decode methods are welcomed to add as an optional step

PR #1232, opened 2020-02-21
add metadata set method

vtolstov:


jakexks:
Need to work out what to do about GitHub actions secrets on PRs from forks :(

jakexks (approved):

Issue #1189, opened 2020-02-12
[BUG] in client/options.go handling


vtolstov:
**Describe the bug**
client/options.go contains this on line 155:
o.Selector.Init(selector.Registry(r))
if we didn't provide selector as option, we try to use nil value and panics

vtolstov:
i'm believe that this will be fixed with provided pr

PR #1235, opened 2020-02-22
fix panic in empty selector, when assign registry

vtolstov:

### DIFF
--- a/client/options.go
+++ b/client/options.go
@@ -152,6 +152,9 @@ func Registry(r registry.Registry) Option {
 	return func(o *Options) {
 		o.Registry = r
 		// set in the selector
+		if o.Selector == nil {
+			o.Selector = selector.DefaultSelector
+		}
 		o.Selector.Init(selector.Registry(r))
 	}
 }


vtolstov:
close #1189

asim:
not needed. its here https://github.com/micro/go-micro/blob/master/client/options.go#L103

vtolstov:
i'm sure that i saw panic for such stuff, i'm try to reproduce and provide stack trace

asim:
We should not polute options with nil checks. This should be upfront setup. If it fails at this stage there was previously. As initialisation. The reason to not do it here is because we could then just end up doing it in literally every option.

vtolstov:
I found how to i get this issue, this is our standard method for variadic options parsing.
I'm create options struct and fill it.in other places we dont provide new options func. So we need to be consistent

asim:
Where are you creating it? We provide a NewOptions() function that generates a default. We can't cater to empty options being parsed. We have to conform to a single way of how we expect the tools to work. Options functions need to be kept simple. The initialisation should be upfront.

vtolstov:
i think that in flow package i'm utilize NewOptions() and also recheck other usage across go-micro other options structs

PR #1237, opened 2020-02-22
prune util/log and user logger

vtolstov:


vtolstov:
@asim main question for this pr - is it ok to remove overriding os.Stderr and os.Stdout?
I don't see now any benefits of overriding it now. Mostly it brings problems with race detector, ability to redirect log output

vtolstov:
panic unrelated for this pr, but looks interesting
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0xa0a99f]

goroutine 16 [running]:
github.com/micro/go-micro/v2/router.(*router).advertiseEvents(0xc000160180, 0x0, 0x0)
	/home/runner/work/go-micro/go-micro/router/default.go:470 +0xf4f
github.com/micro/go-micro/v2/router.(*router).Advertise.func1(0xc000160180)
	/home/runner/work/go-micro/go-micro/router/default.go:604 +0x58
created by github.com/micro/go-micro/v2/router.(*router).Advertise
	/home/runner/work/go-micro/go-micro/router/default.go:599 +0x38f
FAIL	github.com/micro/go-micro/v2/router	0.509s
```

vtolstov:
i'm prefer to remove stderr/stdout overriding and after re-read code, don't understand current benefits, as we use across all go-micro logger functions. And panics are recovered and logged too

asim:
Capturing stderr and stdout let us capture all output. Not just what's logger but also what isn't logged by us

asim:
I think we might have to stop capturing though if it causes problems. I'll find another way

vtolstov:
ok, do you have any other objections not merge this ?=)
P.S. i think that after this will be merged we need to restructure log usage for example to add field type of service so brokers uses field broker=nats/eats/kafka etc, transport and so

asim:
Before merging I need you to run a test. Create a service that uses the logger and then call it's Debug.Log endpoint to see if the log lines exist in the buffer

vtolstov:
Ok, I'm add such test.
Does we need always to store all messages regarding log level, but output to console only when level is satisfied?

asim:
You don't need to add a test, just test it locally. Make sure if you create any new service the logs are stored. Thats it. Then we can merge.

On your other question. Regardless of level we just store or output at the level. Otherwise the buffer would be flooded with trace level logs. We just need a way to flip log levels at runtime, but we'll do that later as its do to with overall service config.

vtolstov:
@asim as i see logs there:

console output with trace level:
```
2020-02-23 15:40:24 Starting [service] go.micro.srv.t1
2020-02-23 15:40:24 Server [grpc] Listening on 127.0.0.1:12345
2020-02-23 15:40:24 Registry [memory] Registering node: go.micro.srv.t1-9ac65a3d-6392-422e-b12b-fb613b91031e
2020-02-23 15:40:24 Registry added new service: go.micro.srv.t1, version: latest
```

from Debug.Log:
```
log.Record{Timestamp:time.Time{wall:0x0, ext:63718058424, loc:(*time.Location)(0x1d33cc0)}, Metadata:map[string]string{"level":"info"}, Message:"Starting [service] go.micro.srv.t1"}
log.Record{Timestamp:time.Time{wall:0x0, ext:63718058424, loc:(*time.Location)(0x1d33cc0)}, Metadata:map[string]string{"level":"info"}, Message:"Server [grpc] Listening on 127.0.0.1:12345"}
log.Record{Timestamp:time.Time{wall:0x0, ext:63718058424, loc:(*time.Location)(0x1d33cc0)}, Metadata:map[string]string{"level":"info"}, Message:"Registry [memory] Registering node: go.micro.srv.t1-9ac65a3d-6392-422e-b12b-fb613b91031e"}
log.Record{Timestamp:time.Time{wall:0x0, ext:63718058424, loc:(*time.Location)(0x1d33cc0)}, Metadata:map[string]string{"level":"debug"}, Message:"Registry added new service: go.micro.srv.t1, version: latest"}
```

asim (approved):

PR #1252, opened 2020-02-24
Dynamic Runtime source for k8s with github packages

ben-toogood:


ben-toogood:
@asim what is the best way for me to test this?

ben-toogood:
Do not merge until we're ready to switch to github packages for build source.

asim:
Let me know when the PR is in working state as discussed, will then add comments if necessary.

1. Maintain github source for source code flow with entrypoint
2. Add github package registry flow for per service builds
3. Allow whatever else to drop through verbatim
4. Change client.Format to swap / with - so helloworld/api becomes helloworld-api-latest in deployment

ben-toogood:
@asim I updated based on our discussion. This is safe to merge as it will only edit the deployment for serviced named such as demo/api or which use the github docker registry. 

I'll be working on the micro cli and github actions sides tomorrow.

asim (approved):

asim:
It looks like it works as we're expecting from the code. Good job Ben. Let's see what edge cases we uncover with that name formatting and the commands.

PR #1253, opened 2020-02-24
fix grpc json codec

yzprofile:
`jsonpbMarshaler.MarshalToString` make enum type translate to string in response, and cause many problems in web side.


asim:
Surely this is going to cause more problems than it solves. Most people are expecting this to deal with proto types like oneof. Which doesn't work without this.

vtolstov:
we can accept some options to marshaler to change it behaviour, but current pr brings problems with others

PR #1254, opened 2020-02-24
fix router panic

vtolstov:


router/default.go
```diff
@@ -700,16 +704,15 @@ func (r *router) Stop() error {
 		// extract the events
 		r.drain()
 
+		r.sub.Lock()
```
asim: the lock is per subscriber so I'm not sure this makes sense?

asim: or should be

asim (approved):

PR #1258, opened 2020-02-25
adds Panic & Panicf to logger

gggwvg:


vtolstov (approved):

vtolstov:
thanks!

asim:
Why add panic when we have Fatal?

vtolstov:
@asim, because we have panic log level. dont know why we have such level. may be we miss in review?

vtolstov:
best of all - delete panic level and helper

gggwvg:
Fatal means program have to exit.
Panic logs a message than panic.

vtolstov:
this is bad thing, because panic not goes to logger, so this is not good. for panic we need to something like wrapper, that internally have recover() format message, write to output and exit

asim:
yea, as a framework I'm not sure its right to allow people to panic without having recovery

vtolstov:
@asim so to not have two levels with the same stuff, we remove panic?

asim:
Lets remove Panic. I don't think its right for us to promote panicking right now. This is not what we want. I think this is not a practice we want. We want people to return errors where it makes sense because this is microservices, you want to send the error back to the caller.

PR #1260, opened 2020-02-25
Auth Service Account

ben-toogood:
Allow a client to pass a service account token when making requests.


asim:
https://github.com/micro/go-micro/pull/1261

Added Token option to Auth interface, included its usage in client wrapper. Config now loaded on service.Init since syscalls for files on every request is expensive. Changed Validate to Verify because verification of tokens makes sense whereas validation is parsing related. Added MICRO_AUTH_TOKEN variable to set the auth token.

PR #1263, opened 2020-02-26
Adjusting the BeforeStart () position

gggwvg:
BeforeStart () executes too late, so I can't modify s.opts.Handler in it


gggwvg:
@asim @vtolstov

vtolstov (approved):

PR #1264, opened 2020-02-26
generate pseudo accounts

asim:
Move to https://github.com/nats-io/nkeys later for default noop


crufter (approved):
Tested and it works locally

PR #1265, opened 2020-02-26
various fixes

vtolstov:
* server/grpc fixes
* logger fixes

### DIFF
--- a/server/grpc/grpc.go
+++ b/server/grpc/grpc.go
@@ -555,8 +555,8 @@ func (g *grpcServer) Subscribe(sb server.Subscriber) error {
 	}
 
 	g.Lock()
-
 	if _, ok = g.subscribers[sub]; ok {
+		g.Unlock()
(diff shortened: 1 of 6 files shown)


server/grpc/grpc.go
```diff
@@ -363,7 +363,17 @@ func (g *grpcServer) processRequest(stream grpc.ServerStream, service *service,
 					err = errors.InternalServerError("go.micro.server", "panic recovered: %v", r)
 				}
 			}()
-			returnValues = function.Call([]reflect.Value{service.rcvr, mtype.prepareContext(ctx), reflect.ValueOf(argv.Interface()), reflect.ValueOf(rsp)})
+			vreq := reflect.ValueOf(argv.Interface())
+			if vreq.IsNil() {
```
asim: Surely if the argument is nil this is not an issue because its just a blank arg?

PR #1270, opened 2020-02-27
Replace validation error with regex for cockroach namespace

ben-toogood:


store/cockroach/cockroach.go
```diff
@@ -206,16 +206,17 @@ func (s *sqlStore) configure() error {
...
-	for _, r := range namespace {
-		if !unicode.IsLetter(r) {
-			return errors.New("store.namespace must only contain letters")
-		}
+	// store.namespace must only contain letters
+	reg, err := regexp.Compile("[^a-zA-Z0-9]+")
+	if err != nil {
+		return errors.New("error compiling regex for namespace")
 	}
+	namespace = reg.ReplaceAllString(namespace, "")
```
asim, line 214: Does that do what you expect it to do? Because the assumption here is that it will replace all strings matched in the regexp with nothing.

ben-toogood: Yes, it can't be replaced with a - etc so I think just removing them is okay

asim: I think I totally misinterpreted what regexp.ReplaceAllString did...

asim (approved):

asim:
We may run name clashing but lets fix later

PR #1277, opened 2020-02-29
adding dumb-init to base image

xmlking:
This helps us adopt recommended   `dumb-init`  when building go-micro based microservices. 

```
COPY --from=builder /usr/bin/dumb-init /usr/bin/dumb-init
```

https://github.com/xmlking/micro-starter-kit/blob/develop/Dockerfile#L44


asim:
Is this available as an apk package?

vtolstov:
Can you provide this change as single commit

PR #1279, opened 2020-02-29
config: adding DisableUpdates option

xmlking:
Adding DisableUpdates option to disable Watcher for the file source


asim:
We discussed this being generally applied across config. This PR does it directly in the file source. You need to apply the disabling of updates to the thing that uses the source watcher not inside the source itself.

asim:
Disable updates should  be applied to the loader not source

PR #1283, opened 2020-03-02
k8s runtime - get status from pods

ben-toogood:


jakexks (approved):
I think this should work(!)

PR #1284, opened 2020-03-02
Add a cache to workers KV storage implementation

jakexks:
This adds another option to the workers KV Store implementation which enables an optional cache, so bursts of reads for the same keys don't hammer Cloudflare.


asim:
I am looking forward to when we build this a layer up, but good start

asim (approved):

PR #1285, opened 2020-03-02
logger fixes and improvements

vtolstov:


vtolstov:
@asim @xmlking needs you review as i can forgot something

asim (approved):

PR #1295, opened 2020-03-04
server: not subscribe to topic with own service name

vtolstov:


asim (approved):

asim:
Subscribe only if `if s.opts.Router != nil {` so the proxy/network can still be a message exchange

PR #1296, opened 2020-03-04
[WIP]: broker ErrorHandler option

vtolstov:
ErrorHandler option to brokers can catch errors before message  passed to handler

### DIFF
--- a/broker/broker.go
+++ b/broker/broker.go
@@ -28,6 +28,7 @@ type Event interface {
 	Topic() string
 	Message() *Message
 	Ack() error
+	Error() error
 }
 
 // Subscriber is a convenience return type for the Subscribe method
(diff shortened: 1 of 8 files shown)


vtolstov:
i'm test various signatures of ErrorHandler func, and all other stuff can be filled when init struct like
```
h := &errHandler{ctx: ctx, xx:tt}
broker.ErrorHandler(h)
```

with signature of ErrorHandler(Event, error) we can get topic name, ack message, and get it body (if this is not unmarshal error).
next step is return from Codec specific error type? may be errors.Error ?

most common use-case - when error happens
* log error
* move message to some dead letter topic
* ack msg from original topic
* optional - unsubscribe from topic

vtolstov:
i'm see that we don't need customer Error to this handler, as Queue not really useful when working via error handler. As we have message and error this is enough info to fix/workaround problem.

vtolstov:
ready to merge

asim (approved):

PR #1303, opened 2020-03-05
server/grpc: avoid panic in case of nil Header

vtolstov:


jakexks (approved):

server/grpc/subscriber.go
```diff
@@ -178,6 +178,11 @@ func (g *grpcServer) createSubHandler(sb *subscriber, opts server.Options) broke
 		}()
 
 		msg := p.Message()
+		// if we don't ave headers, create empty map
```
jakexks, line 181:
```suggestion
		// if we don't have headers, create empty map
```

PR #1306, opened 2020-03-05
outline config/secrets

asim:


jakexks (commented):
I don't think this is the right approach. Our config interfaces are designed to retrieve entire nested maps of configuration data. Normal secrets APIs are simple key value (e.g. github secrets, hashicorp vault, etc).

Closing and making a separate PR

PR #1307, opened 2020-03-06
add models

asim:


model/crud/crud.go
```diff
@@ -0,0 +1,29 @@
+// Package crud provides a crud interface
+package crud
+
+type CRUD interface {
+	// Read values
+	Read(...ReadOption) ([]Record, error)
```
jakexks: []*Record?

```suggestion
	Read(...ReadOption) ([]*Record, error)
```

model/crud/crud.go
```diff
@@ -0,0 +1,29 @@
...
+	// Write a value
+	Write(id string, v interface{}) error
+	// Update a record
+	Update(id string, v interface{}) error
+	// Delete a record
+	Delete(id string) error
```
jakexks: Should there be options for Write, Update, Delete even if we can't think of any yet?

```suggestion
	Write(id string, v interface{}, ...WriteOption) error
	// Update a record
	Update(id string, v interface{}, ...UpdateOption) error
	// Delete a record
	Delete(id string, ...DeleteOption) error
```

model/crud/crud.go
```diff
@@ -0,0 +1,29 @@
...
+}
+
+type ReadOption struct {
+	// Id to read
+	Id string
+	// Limit number of returns records
+	Limit int
+	// Offset
+	Offset int
+}
```
jakexks: e.g., ListOption should have offset and limit too

model/crud/crud.go
asim: Record is an interface, should it be a struct? The idea is that it provides a method to scan in the value. Otherwise Record value would be an interface that's not actually the users struct type?

asim: I mean we could do it like https://golang.org/pkg/database/sql/#Row.Scan

model/crud/crud.go
asim: There is no list. CRUD doesn't have anything called list. You use Read and it returns rows.

model/crud/crud.go
jakexks: Ah, that makes sense

PR #1321, opened 2020-03-09
add helper function to determine logger level

vtolstov:


vtolstov:
simple usage
```
if logger.V(logger.TraceLevel, myLog) {
  myLog.Tracef("some big message %v", myBigStruct)
}
```

mostly i think that first user of this is go-micro itself. So we can add many trace/debug stuff to our core without significant overhead. Also users what have long-lived apps and that wants to avoid unnecessary allocations can use it too.

asim:
Makes sense.

asim (approved):

PR #1323, opened 2020-03-09
minimize allocations in logger and tunnel code

vtolstov:


vtolstov:
fixes bug in helper that always copy fields regardless of log level, and also store cipher inside session, so we avoid many allocations in tunnel code
```
BenchmarkNew-16             100000            140182 ns/op            8286 B/op        143 allocs/op
```

so we save 20000 ns/op and 20 allocs/op

asim (approved):

asim:
I think you have to rebase against master before this can be merged

vtolstov:
How you know that I'm finish? Amazing

PR #1325, opened 2020-03-10
Add secrets interface to config/secrets

jakexks:
This adds an interface for using asymmetric (public key cryptography) and symmetric encryption to encipher small messages. I've also included an implementation of each using http://nacl.cr.yp.to/

### DIFF
--- a/config/secrets/box/box.go
+++ b/config/secrets/box/box.go
@@ -1,0 +1,45 @@
+// Package box is an asymmetric implementation of config/secrets using nacl/box
+package box
+
+import "github.com/micro/go-micro/v2/config/secrets"
+
+type box struct {
+	options secrets.Options
+
+	publicKey  [32]byte
+	privateKey [32]byte
+}
+
+// NewCodec returns a nacl-box codec
+func NewCodec(opts ...secrets.Option) secrets.Codec {
+	b := &box{}
+	for _, o := range opts {
+		o(&b.options)
+	}
+	return b
+}
+
+// Init initialises a box
+func (b *box) Init(...secrets.Option) error {
+	return nil
+}
+
+// Options returns options
+func (b *box) Options() secrets.Options {
+	return b.options
+}
+
+// String returns nacl-box
+func (*box) String() string {
+	return "nacl-box"
+}
+
+// Encrypt encrypts
+func (b *box) Encrypt(in []byte, opts ...secrets.EncryptOption) ([]byte, error) {
+	return nil, nil
+}
+
+// Decrypt Decrypts
+func (b *box) Decrypt(in []byte, opts ...secrets.DecryptOption) ([]byte), error) {
+	return nil, nil
+}
--- a/config/secrets/box/box.go
+++ b/config/secrets/box/box.go
@@ -1,7 +1,13 @@
 // Package box is an asymmetric implementation of config/secrets using nacl/box
 package box
 
-import "github.com/micro/go-micro/v2/config/secrets"
+import (
+	"github.com/micro/go-micro/v2/config/secrets"
+	"github.com/pkg/errors"
+	naclbox "golang.org/x/crypto/nacl/box"
+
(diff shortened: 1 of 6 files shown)


config/secrets/box/box.go
```diff
@@ -0,0 +1,87 @@
...
+	return b
+}
+
+// Init initialises a box
+func (b *box) Init(opts ...secrets.Option) error {
+	for _, o := range opts {
+		o(&b.options)
+	}
+	if len(b.options.PrivateKey) != 32 || len(b.options.PublicKey) != 32 {
+		return errors.New("A 32 byte public and private key must be provided")
```
vtolstov: i think that error must contains required len. Because i can provide both of them but not needed len.

config/secrets/box/box.go
```diff
@@ -0,0 +1,87 @@
...
+	return "nacl-box"
+}
+
+// Encrypt encrypts a message with the sender's private key and the receipient's public key
+func (b *box) Encrypt(in []byte, opts ...secrets.EncryptOption) ([]byte, error) {
+	var options secrets.EncryptOptions
+	for _, o := range opts {
+		o(&options)
+	}
+	if len(options.RecipientPublicKey) != 32 {
```
vtolstov: 32 must be some const, to reuse it in all places

config/secrets/box/box.go
```diff
@@ -0,0 +1,87 @@
...
+		return []byte{}, errors.New("Recepient's public key must be provided")
+	}
+	var recipientPublicKey [32]byte
```
vtolstov: use the same const len...

config/secrets/secretbox/secretbox.go
```diff
@@ -0,0 +1,71 @@
...
+	return "nacl-secretbox"
+}
+
+func (s *secretBox) Encrypt(in []byte, opts ...secrets.EncryptOption) ([]byte, error) {
+	// no opts are expected, so they are ignored
+
+	// there must be a unique nonce for each message
+	var nonce [24]byte
+	if _, err := rand.Reader.Read(nonce[:]); err != nil {
+		return []byte{}, errors.Wrap(err, "Couldn't obtain a random nonce from crypto/rand")
```
vtolstov: may be i'm forget something but usually all errors in lower case

config/secrets/secrets.go
```diff
@@ -0,0 +1,77 @@
...
+func SecretKey(key []byte) Option {
+	return func(o *Options) {
+		o.SecretKey = key
+	}
+}
+
+// PublicKey sets the asymmetric Public Key of this codec
+func PublicKey(key []byte) Option {
+	return func(o *Options) {
+		o.PublicKey = key
```
vtolstov: i think that we need to copy, to avoid changing this later in user code.

go.sum
```diff
@@ -284,6 +284,7 @@ github.com/mholt/certmagic v0.9.3 h1:RmzuNJ5mpFplDbyS41z+gGgE/py24IX6m0nHZ0yNTQU
 github.com/mholt/certmagic v0.9.3/go.mod h1:nu8jbsbtwK4205EDH/ZUMTKsfYpJA1Q7MKXHfgTihNw=
 github.com/micro/cli/v2 v2.1.2 h1:43J1lChg/rZCC1rvdqZNFSQDrGT7qfMrtp6/ztpIkEM=
 github.com/micro/cli/v2 v2.1.2/go.mod h1:EguNh6DAoWKm9nmk+k/Rg0H3lQnDxqzu5x5srOtGtYg=
+github.com/micro/go-micro v1.18.0 h1:gP70EZVHpJuUIT0YWth192JmlIci+qMOEByHm83XE9E=
```
vtolstov: you have invalid import somewhere, that pulls this old dep

vtolstov (dismissed):

asim (approved):

PR #1328, opened 2020-03-10
add the concept of a resource

asim:
Resource is the concept of an external resource. Ideally this will enable the management of any type of external service that is not native to micro e.g redis, postgres, etc. Services can then "depend" on a Resource, adding them to a list of dependencies which needs to be started. Dependencies could also include other services.

### DIFF
--- a/resource/resource.go
+++ b/resource/resource.go
@@ -1,0 +1,20 @@
+// Package resource is an external resource
+package resource
+
+// Resource represents an external resource
+// which can be run alongside a service e.g
+// a database.
+type Resource interface {
+	// Initialise
+	Init(...Option) error
+	// Get options
+	Options() Options
+	// Run the resource
+	Run() error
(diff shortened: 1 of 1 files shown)


vtolstov:
I think that we need to extend such description to contains any any service that can be used as dependency.
For example i have strong dependency, that api gw service and other services must be started only after auth service. So if auth service in some location is not working i dont need to start all of them, because I can't validate tokens from users and other services, and also can't use mtls.
For micro service we can have default checker for it like check registry.
For non micro we can have some func to check

asim:
Resource is simply the ability to define and run that resource. We will separately define dependencies and the DAG for how that dependency list is started. It is likely that the go-micro/runtime understands how to do that or something higher up.

vtolstov:
Yes, DAG will be useful for such case....  Need to deeply understand runtime package.

PR #1335, opened 2020-03-12
Rewrite the store interface

jakexks:
This PR updates the store interface to be coherent with the  [design document](https://github.com/micro/development/blob/9582e9ce6f3a5ceaf9879568867c1d4c286d1931/design/store.md).

Changes:
* etcd and memory stores rewritten to be completely complaint with all possible options
* cockroachdb, cloudflare and service stores updated to be compliant with the interface.


store/cockroach/cockroach.go
```diff
@@ -40,44 +40,44 @@ func (s *sqlStore) Init(opts ...store.Option) error {
 }
 
 // List all the known records
-func (s *sqlStore) List() ([]*store.Record, error) {
+func (s *sqlStore) List(opts ...store.ListOption) ([]string, error) {
 	rows, err := s.db.Query(fmt.Sprintf("SELECT key, value, expiry FROM %s.%s;", s.database, s.table))
```
vtolstov: Please move all queries to init phase.to avoid allocating strings and db stuff

asim: Refactor later?

vtolstov (dismissed):
Please move all prepared stuff to init tine to avoid reallocation

store/cockroach/cockroach.go
jakexks: makes sense

PR #1342, opened 2020-03-13
fix log level helper

vtolstov:


vtolstov:
fix check for needed log level to display message

PR #1343, opened 2020-03-13
Extract should not include loopback addresses

gggwvg:
Now Extract can get 127.0.0.1, causing the registered address of the service to be 127.0.0.1 and not accessible


gggwvg:
@asim

gggwvg:
@vtolstov

gggwvg:
image

My service running on k8s, registered as 127.0.0.1 in consul, that is not the real node IP.
My env MICRO_SERVER_ADDRESS=":9090"

https://github.com/micro/go-micro/blob/v2.2.0/server/grpc/grpc.go#L594

In my case, `addr.Extract("")` returns 127.0.0.1, that is not as expected.

asim:
Specify 0.0.0.0 as your address

vtolstov:
Can you provide list of all addresses ?

gggwvg:
> Specify 0.0.0.0 as your address

It's not work.
Although my env MICRO_SERVER_ADDRESS="0.0.0.0:9090", IP in consul is still 127.0.0.1:9090, that's a invalid service address.

image

gggwvg:
> Can you provide list of all addresses ?

image

vtolstov:
Micro try to bind to internal address, so you need to add any internal address

Issue #1209, opened 2020-02-17
[BUG] If nsq broker subscribe with three queues, message will run 3*3=9 times handlers


0987363:
go-micro 1.18.0
go-plugins  1.15
go version: 1.13

I use nsq as broker,  register one subscribe with one topic and multiple queue.
```
      micro.RegisterSubscriber(
          models.TopicDevice,
          service.Server(),
          device.Test1,
          server.SubscriberQueue("test1"),
      )
      micro.RegisterSubscriber(
          models.TopicDevice,
          service.Server(),
          device.Test2,
          server.SubscriberQueue("test2"),
      )
      micro.RegisterSubscriber(
          models.TopicDevice,
          service.Server(),
          device.Test3,
          server.SubscriberQueue("test3"),
      ) 
```

So i had three different channel.
When i publish one message to this topic, every subscriber channel should receive one message only.

But when go-micro receive message from nsq, every subscriber channel and every message will run three times handlers. 
Run 3*3=9 times  handlers in total .

I found code at go-micro/server/rpc_router.go, func ProcessMessage()
Every channel will register a subscribe, one topic will register three subscribe.

```
func (router *router) Subscribe(s Subscriber) error {
    sub, ok := s.(*subscriber)
    if !ok {
        return fmt.Errorf("invalid subscriber: expected *subscriber")
    }
    if len(sub.handlers) == 0 {
        return fmt.Errorf("invalid subscriber: no handler functions")
    }

    if err := validateSubscriber(sub); err != nil {
        return err
    }

    router.su.Lock()
    defer router.su.Unlock()

    // append to subscribers
    subs := router.subscribers[sub.Topic()]
    subs = append(subs, sub)
    router.subscribers[sub.Topic()] = subs

    return nil
}

func (router *router) ProcessMessage(ctx context.Context, msg Message) error {
   ......

    // get the subscribers by topic
    subs, ok := router.subscribers[msg.Topic()]
    if !ok {
        router.su.RUnlock()
        return nil
    }

    // unlock since we only need to get the subs
    router.su.RUnlock()

    var errResults []string

    // we may have multiple subscribers for the topic
    for _, sub := range subs {
        // we may have multiple handlers per subscriber
        for i := 0; i < len(sub.handlers); i++ {
            // get the handler
            handler := sub.handlers[i]
......
```

0987363:
I try add this code, it worked

```
diff --git a/broker/nsq/nsq.go b/broker/nsq/nsq.go
[...]

PR #1379, opened 2020-03-19
Fixed nsq one queue run multiple times handlers, issue #1209

0987363:
If nsq broker subscribe with three queues, message will run 3*3=9 times handlers.

So add queue check before handler.

https://github.com/micro/go-micro/issues/1209

I add pr to go-plugins also.
https://github.com/micro/go-plugins/pull/513

### DIFF
--- a/server/rpc_router.go
+++ b/server/rpc_router.go
@@ -526,6 +526,11 @@ func (router *router) ProcessMessage(ctx context.Context, msg Message) (err erro
 
 	// we may have multiple subscribers for the topic
 	for _, sub := range subs {
+		// Nsq one subscriber had multiple queue, one queue bind to one connection, skip repeat subscriber.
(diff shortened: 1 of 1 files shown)


asim (commented):
This seems like a fix purely for nsq. Where does this queue header come from?

0987363:
> This seems like a fix purely for nsq. Where does this queue header come from?

I also add queue to header with go-plugins pr.
https://github.com/micro/go-plugins/pull/513

vtolstov:
i think that this is not proper way to fix this. because we add header for one single broker implementation.

crazybber:
Maybe you need to prove  a little more

asim:
Closing. Fix needs to provided as a more generic form.

vtolstov:
whats says nsq author for this use-case ?

PR #1382, opened 2020-03-19
WIP: binary minimization and customization

vtolstov:


asim:
I think we want to continue to iterate on this and merge when possible. It would be a good transition for v3 for binary minimization.

asim:
Going to close as unlikely to merge now but we can revisit later as we start to refactor

asim:
Now is probably when we revisit this as we're about to strip config/cmd and top level service. So we can move to per package maps for plugins and remove defaults.

PR #1384, opened 2020-03-20
Updated auth interface

ben-toogood:


auth/auth.go
```diff
@@ -31,29 +31,32 @@ type Resource struct {
...
 	// ID of the account (UUIDV4, email or username)
-	Id string `json:"id"`
+	ID string `json:"id"`
 	// Token used to authenticate
-	Token string `json:"token"`
-	// Time of Account creation
-	Created time.Time `json:"created"`
-	// Time of Account expiry
-	Expiry time.Time `json:"expiry"`
+	Token Token `json:"token"`
```
asim: Token (aka jwt) would not be stored in the account itself. We'd only have secret. Tokens would be created and its up to the client side to store/make use of those.

auth/auth.go
```diff
@@ -4,25 +4,45 @@ package auth
...
+	ErrForbidden = errors.New("resource forbidden")
+)
+
 // Auth providers authentication and authorization
 type Auth interface {
-	// Init the auth package
-	Init(opts ...Option) error
-	// Options returns the options set
+	// String returns the name of the implementation
+	String() string
```
vtolstov: String() is always last,
Init() is mostly first

auth/default.go
```diff
@@ -1,122 +1,229 @@
...
+
+// Init the auth
+func (m *memory) Init(opts ...Option) {
 	for _, o := range opts {
-		o(&n.opts)
+		o(&m.opts)
+	}
+
+	// use a memory store
+	if m.opts.Store == nil {
```
vtolstov: if we pass non memory store, does it looks right to return memory for auth?

vtolstov (changes requested):

auth/auth.go
```diff
@@ -31,40 +51,47 @@ type Resource struct {
...
-	// in metadata
+	// MetadataKey is the key used when storing the account in metadata
 	MetadataKey = "auth-account"
-	// CookieName is the name of the cookie which stores the
-	// auth token
-	CookieName = "micro-token"
+	// TokenCookieName is the name of the cookie which stores the auth token
+	TokenCookieName = "micro-token"
+	// SecretCookieName is the name of the cookie which stores the auth secret
+	SecretCookieName = "micro-secret"
```
asim, line 94: A cookie should not store the secret, this is like storing your password in a cookie. It's going to get passed in every request since cookies are included in headers. I would not feel comfortable having my passwords stored anywhere accept server side.

We may needs refresh tokens, session tokens or to just set UI side to a slightly longer live token.

https://auth0.com/learn/refresh-tokens/
https://auth0.com/blog/refresh-tokens-what-are-they-and-when-to-use-them/

vtolstov: I'm prefer to use session and refresh tokens.
Access token has short lifetime is like 1h, and longer refresh.
When refresh token used to issue new access and refresh, it removed from store,

ben-toogood: Where would the JS client store the secret?

auth/options.go
```diff
@@ -69,31 +70,31 @@ type GenerateOptions struct {
...
+func WithRoles(rs []string) GenerateOption {
 	return func(o *GenerateOptions) {
 		o.Roles = rs
 	}
 }
 
-// Expiry for the generated account's token expires
-func Expiry(ex time.Time) func(o *GenerateOptions) {
+// WithSecretExpiry for the generated account's secret expires
+func WithSecretExpiry(ex time.Duration) GenerateOption {
```
asim, line 95: Should secrets expire? If its like a password?

ben-toogood:
Approved by @asim

crazybber:
Why do we need to update auth  interface? any improvement?

vtolstov:
@crazybber yes, now it support authentication and authorization

PR #1387, opened 2020-03-20
client/grpc: remove json-iterator usage

vtolstov:
* minimize external deps and binary size
* if user wants json-iterator codec it must be used in server and
  client code. so best to use it via go-plugins


vtolstov:
improvements will be in next series when we enable only internal implementations of all stuff (store, transport, broker and so)

PR #1388, opened 2020-03-23
api/handler/rpc: use client selector in api

### LINKED ISSUE #1003: api handler need support SelectOption
**Is your feature request related to a problem? Please describe.**
I want filter `go.micro.api.xxx` service. So I custom micro main() func.
```go
func main() {
	cmd.Init(
		micro.WrapClient(xxx.NewClientWrapper()),
	)
}
```

`xxx` client wrapper add CallOption
```go
client.WithSelectOption(selector.WithFilter(MyFilterFunc()))
```
But when I build and run `micro`, filter work, but service not filtered...😂

Then I see `api` and `rpc` handler use a strategy() func.
**ignore input to this function, use services above**
```go
// strategy is a hack for selection
func strategy(services []*registry.Service) selector.Strategy {
	return func(_ []*registry.Service) selector.Next {
		// ignore input to this function, use services above
		return selector.Random(services)
	}
}
```
```go
// create strategy
so := selector.WithStrategy(strategy(service.Services))
if err := c.Call(cx, req, rsp, client.WithSelectOption(so)); err != nil {
}
```

**Describe the solution you'd like**
I try to remove the strategy, then build and run `micro`, it's ok. But I don't know if it will cause other problems.
Can we remove the strategy, or why use like this?
Or any other better ways?

vtolstov:
close #1003 


vtolstov:
@asim i need you decision why you want to use sticky strategy inside api. And not use passed client with all options (that provides more options and control like strategy, selector and other stuff)

asim:
The reason it's done like this is because the API is pulling a list of endpoints from the registry. It's building a routing table based on metadata which requires it to know what node to call. So by doing this you've effectively broken that.

Please don't make the change without knowing why it's done like this. The API code needs to be refactored to change what the handlers are provided.

vtolstov:
What is the difference with static passing services and get they from registry?

asim:
See the API router. It's building a list of endpoints based on the registry so it doesn't make sense to call it twice.

vtolstov:
Ok, now i'm understand how api works with resolver, router and other stuff.
But as i see its not possible to register endpoint with statically typed service name.

so if i want something like this
```
hrpc := apirpc.WithService(&api.Service{          
  Name: "account",                                
  Endpoint: &api.Endpoint{                        
    Name:    "AccountService.Lookup",             
    Path:    []string{"/example"},                
    Method:  []string{"GET"},                     
    Handler: apirpc.Handler,                      
  },                                              
}, ahandler.WithService(svc),                      
)                                                 
````

that means i want to handle /example http endpoint with method GET via account.AccountService.Lookup rpc call ? Services must be filled from registry.
What i need to do? Create own resolver, because i don't have Endpoint filled in registry because i have many services and this abuses gossip registry (and other registries too)?
So if i have own resolver that able to dynamically add/remove so mapping, i can get what i want ?

crazybber:
> Ok, now i'm understand how api works with resolver, router and other stuff.
> But as i see its not possible to register endpoint with statically typed service name.
> 
> so if i want something like this
> 
> ```
> hrpc := apirpc.WithService(&api.Service{          
>   Name: "account",                                
>   Endpoint: &api.Endpoint{                        
>     Name:    "AccountService.Lookup",             
>     Path:    []string{"/example"},                
>     Method:  []string{"GET"},                     
>     Handler: apirpc.Handler,                      
>   },                                              
> }, ahandler.WithService(svc),                      
> )                                                 
> ```
> 
> that means i want to handle /example http endpoint with method GET via account.AccountService.Lookup rpc call ? Services must be filled from registry.
> What i need to do? Create own resolver, because i don't have Endpoint filled in registry because i have many services and this abuses gossip registry (and other registries too)?
> So if i have own resolver that able to dynamically add/remove so mapping, i can get what i want ?

@hb-chen  come to see ,if you have any comment ,i think you will like this feature.

vtolstov:
so i think that asim is right and this pr is not needed. To not rewrite all api stuff the simplest case is write resolver.

asim:
The api package needs to be reworked. Its something I've left long overdue. The api resolver is used to map a http path to an internal service and endpoint. The router then uses this to find a service. What you need is an implementation of the router if you want to replace the registry.

vtolstov:
this issue is solved by static router that have ability to get custom binding of http endpoint to micro service

PR #1390, opened 2020-03-23
api/handler/rpc: log errors

vtolstov:


vtolstov:
self merge as have only minimal improvements with logger.

Issue #1389, opened 2020-03-23
add call option for proxy, make proxy can be used dynamically


hb-chen:
**Is your feature request related to a problem? Please describe.**

Sometimes only part of the traffic needs to pass through the proxy. I think we can add a CallOption, make proxy can be used dynamically.

**Describe the solution you'd like**
simple code, or CallOption first.
```go
	// get proxy
	if prx := os.Getenv("MICRO_PROXY"); len(prx) > 0 {
		service = prx
	} else if len(opts.Proxy) > 0 {
		service = opts.Proxy
	}
```

**Additional context**

asim:
Not implementing at this time

PR #1396, opened 2020-03-24
[FEATURE] add call options proxy, overrides MICRO_PROXY environment

hb-chen:
Use the proxy dynamically through CallOptions.
My implementation is `CallOptions.Proxy` overrides `MICRO_PROXY` environment
https://github.com/micro/go-micro/issues/1389


asim:
Comments have not been addressed. Please reopen when they have.

PR #1397, opened 2020-03-24
Auth load rules

ben-toogood:


auth/service/service.go
```diff
@@ -44,6 +50,13 @@ func (s *svc) Init(opts ...auth.Option) {
 	if key := s.options.PublicKey; len(key) > 0 {
 		s.jwt = jwt.NewTokenProvider(token.WithPublicKey(key))
 	}
+
+	// load rules periodically from the auth service
+	timer := time.NewTimer(time.Second * 30)
```
asim: NewTicker rather than timer since you are not using the actual time returned

ben-toogood: Thanks

ben-toogood:
Approved by @asim

PR #1415, opened 2020-03-26
api/handler: use http.MaxBytesReader and buffer pool

vtolstov:
protect api handlers from OOM cases

### DIFF
--- a/api/handler/api/api.go
+++ b/api/handler/api/api.go
@@ -24,7 +24,13 @@ const (
 
 // API handler is the default handler which takes api.Request and returns api.Response
 func (a *apiHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	request, err := requestToProto(r)
+	bsize := handler.DefaultMaxRecvSize
+	if a.opts.MaxRecvSize > 0 {
+		bsize = a.opts.MaxRecvSize
+	}
+
+	r.Body = http.MaxBytesReader(w, r.Body, bsize)
+	request, err := a.requestToProto(r)
 	if err != nil {
 		er := errors.InternalServerError("go.micro.api", err.Error())
 		w.Header().Set("Content-Type", "application/json")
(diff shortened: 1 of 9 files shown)


vtolstov:
in next commit i want to add bpool to all usage of iotuil.ReadAll

vtolstov:
i think that this good now. if you have enabled some tracer or monitoring that says api payload size (or count of all api requests in some time and incoming traffic for this time) we can calculate best values for buffer pool (size of preallocated buffer and it count)

asim:
Maybe it makes sense to define the pool somewhere in util or in a shared place so it's just defined once and reused?

vtolstov:
> Maybe it makes sense to define the pool somewhere in util or in a shared place so it's just defined once and reused?

i think about it, but why i duplicate - because some messages for specific handlers may have smaller requirements and some bigger. 
What you think?

asim:
Let's leave as is for now. Whats the current max message size for net/http? You've limited this to 4mb which might pose some issues.

vtolstov:
i'm limit to 10 (fix wrong comment =))
go std lib not limit message body. it limits reading from it in parse form function in request. to 10Mb ib user not uses MaxBytesReader.
So 10Mb is the safe default. also for api i don't think that somembody hits this limit.
If you want to post files bigger - use websocket or raise limit.

asim:
I'm not sure I trust this artificial limit of 10mb. We've now introduced something that potentially breaks existing behaviour for anyone using the API. What if people are throwing a 100mb file upload at it?

vtolstov:
This is api stuff. If you want to upload 100mb go ahead and raise limit.
Also note, that for for large file upload most suitable to use websocket. It allows to resume download and do other cool things

asim:
What I'm saying is its a breaking behaviour for an existing tool. There are 1500+ deployments of the api gateway, you have no idea what people are doing. So to limit it wihtout that knowledge is wrong.

vtolstov:
Do you have some stats data? If not, do you plan to have ability to
gathering some anonymous stats?

чт, 26 мар. 2020 г., 19:56 Asim Aslam <notifications@github.com>:

vtolstov:
For limit, what you want to specify by default, or use 0 as no limit?

asim:
I've set max size to 100mb. Lets see what happens.

PR #1428, opened 2020-03-27
fix: rand race condition

acynothia:
rand is not safe for concurrent use by multiple goroutines.

### DIFF
--- a/util/jitter/jitter.go
+++ b/util/jitter/jitter.go
@@ -3,15 +3,20 @@ package jitter
 
 import (
 	"math/rand"
+	"sync"
 	"time"
 )
 
 var (
-	r = rand.New(rand.NewSource(time.Now().UnixNano()))
+	r  = rand.New(rand.NewSource(time.Now().UnixNano()))
+	mu sync.Mutex
 )
 
 // Do returns a random time to jitter with max cap specified
 func Do(d time.Duration) time.Duration {
-	v := r.Float64() * float64(d.Nanoseconds())
+	mu.Lock()
+	v := r.Float64()
+	mu.Unlock()
+	v *= float64(d.Nanoseconds())
 	return time.Duration(v)
 }


vtolstov (changes requested):
mutex is not needed, simply seed random on func init() once and use rand functions without mutex

vtolstov:
you can use rand.Seed(time.Unix().Nano()) once on init and user rand.XXX functions with thread safe

acynothia:
rand is not safe for concurrent use by multiple goroutines. u can found in the following example.

https://github.com/golang/go/blob/master/src/math/rand/rand.go#L293
```go
var globalRand = New(&lockedSource{src: NewSource(1).(*rngSource)})
...
type lockedSource struct {
	lk  sync.Mutex
	src *rngSource
}

func (r *lockedSource) Int63() (n int64) {
	r.lk.Lock()
	n = r.src.Int63()
	r.lk.Unlock()
	return
}

func (r *lockedSource) Uint64() (n uint64) {
	r.lk.Lock()
	n = r.src.Uint64()
	r.lk.Unlock()
	return
}
```

https://github.com/grpc/grpc-go/blob/master/internal/grpcrand/grpcrand.go#L30
```go
var (
	r  = rand.New(rand.NewSource(time.Now().UnixNano()))
	mu sync.Mutex
)

// Int63n implements rand.Int63n on the grpcrand global source.
func Int63n(n int64) int64 {
	mu.Lock()
	res := r.Int63n(n)
	mu.Unlock()
	return res
}

// Intn implements rand.Intn on the grpcrand global source.
func Intn(n int) int {
	mu.Lock()
	res := r.Intn(n)
	mu.Unlock()
	return res
}

// Float64 implements rand.Float64 on the grpcrand global source.
func Float64() float64 {
	mu.Lock()
	res := r.Float64()
	mu.Unlock()
	return res
}
```

vtolstov:
@Astone-Chou i know, what i want to tell you - not need to create own mutex stuff if we simply can use golang rand and have the same behaviour

asim:
I'm closing

PR #1434, opened 2020-03-27
add mdns registry debug

vtolstov:


vtolstov:
small debug message. may be it can helps later

PR #1437, opened 2020-03-29
api: add static router and improve path parser in rpc handler

vtolstov:


vtolstov:
needs to have some discussion

vtolstov:
i'm add simple test to provide example how to use static router and auto filling proto field based on url path

vtolstov:
in modify own stuff to work with this and it works fine with 5 grpc services behind api with static selector.

asim:
Lets merge and keep iterating!

vtolstov:
cool =) the next step for api is modify protoc-gen-micro  and try to use generated stuff?

asim:
OK, lets try it. I'll let you resurrect the PR or whatever.

crazybber:
It is a very cool stuff

PR #1440, opened 2020-03-30
regenerate all proto

vtolstov:


asim:
You need to use the script in micro/scripts/protoc.sh. Any internally referenced imports need a v2 rewrite because protoc-gen-go does not know about go modules.

PR #1442, opened 2020-03-30
auth proto: provide help to protoc-gen-go

vtolstov:


vtolstov:
self merge as it trivial naming stuff.

PR #1452, opened 2020-03-31
fix api metadata extract from context

vtolstov:

### DIFF
--- a/api/handler/rpc/rpc.go
+++ b/api/handler/rpc/rpc.go
@@ -118,6 +118,17 @@ func (h *rpcHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	// create context
 	cx := ctx.FromRequest(r)
+	// get context from http handler wrappers
+	md, ok := r.Context().Value(metadata.MetadataKey{}).(metadata.Metadata)
+	if !ok {
+		md = make(metadata.Metadata)
+	}
+
+	// merge context with overwrite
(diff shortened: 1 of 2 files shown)


api/handler/rpc/rpc.go
```diff
@@ -118,6 +118,17 @@ func (h *rpcHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	md, ok := r.Context().Value(metadata.MetadataKey{}).(metadata.Metadata)
+	if !ok {
+		md = make(metadata.Metadata)
+	}
+
+	// merge context with overwrite
+	cx = metadata.MergeContext(cx, md, true)
+
+	// set merged context to request
+	*r = *r.WithContext(cx)
```
asim, line 131: Why dereference like this?

vtolstov: to modify original context inside r.
r.WithContext returns pointer to http Request. So we need to modify original pointer with new (that actually have only context differs)

asim: https://golang.org/src/net/http/request.go?s=13039:13098#L346

It already does this for you

asim: Ah crap its making a copy

vtolstov: no, it returns new Request pointer (r2). Mostly i need to use *r = *r.Clone(ctx)
as i saw in docs. But none of this two functions modifies original request.

PR #1453, opened 2020-03-31
metadata: allow to remove key from metadata

vtolstov:


metadata/metadata.go
```diff
@@ -25,6 +25,13 @@ func (md Metadata) Get(key string) (string, bool) {
 	return val, ok
 }
 
+func (md Metadata) Del(key string) {
```
asim, line 28: please name Delete

PR #1459, opened 2020-04-01
Auth Interface Iteration

ben-toogood:


auth/auth.go
```diff
@@ -60,16 +61,22 @@ type Resource struct {
...
+	// ID of the account e.g. email
 	ID string `json:"id"`
-	// Secret used to renew the account
-	Secret string `json:"secret"`
+	// Type of the account, e.g. service
+	Type string `json:"type"`
+	// Provider who issued the account
+	Provider string `json:"provider"`
+	// RefreshToken used to renew the account
+	RefreshToken string `json:"refresh_token"`
```
asim: Rather than storing refresh token in the account we should probably store it with the token itself. As its an implementation detail and related to post token usage.

auth/auth.go
```diff
@@ -34,6 +33,8 @@ type Auth interface {
 	Options() Options
 	// Generate a new account
 	Generate(id string, opts ...GenerateOption) (*Account, error)
+	// Login to an existing account
+	Login(id string, opts ...LoginOption) (*Account, error)
```
asim: I think if we're going to have a Login method it should explicitly ask for id, secret as args rather than as an option. Otherwise seeing this I'm just seeing a mirror of the Token method in which case:

`Token(id string, opts ...TokenOption)` where Secret or RefreshToken are passed in as options and we operate based on whats provided. Feels cleaner? Then `micro login` is only calling the Token method and passing in secret or token as required.

auth/options.go
```diff
@@ -50,10 +50,10 @@ func PrivateKey(key string) Option {
 }
 
 // Credentials sets the auth credentials
-func Credentials(id, secret string) Option {
+func Credentials(id, refresh string) Option {
```
asim: This should likely remain id, secret as auth.Credentials signals id, secret.

auth/auth.go
ben-toogood: If we do this then we'd need to return an initial token as part of generation, either as an additional return value or as an attribute of account.

auth/auth.go
ben-toogood: The reason I done it like this is so that oauth providers can login an account without a password.

auth/options.go
ben-toogood: We don't want to store the secret (password) ever client side. We're just using the refresh token to generate new tokens, so I don't see how to action this?

auth/auth.go
ben-toogood: After thinking about this, I agree. I'll merge those two methods.

asim: I'm just feeling like the methods are redundant in the sense that Login/Token are basically a mirror of each other. Whether you need to login or exchange a token its basically one and the same with different params. So being able to specific Secret or Token as an option to the Token method would allow it in one place.

The other thing I'm thinking:

```
type Token struct {
    // Refresh is a function which renews the token
    Refresh func() (*Token, error)
}
```

Rather than specifying the refresh token maybe we specify a refresh function instead which abstracts it away. So when we call auth.Token the auth provider sets up the Refresh() function to handle the details.

auth/auth.go
asim: I think creating an account is not the same as logging into the account. So having the separate call to create the token is not an issue right?

auth/auth.go
ben-toogood: Actually, for oauth logic we wouldn't have either a refresh token (since they're logging in), or a secret. I'll have to add an edgecase to allow unauthenticated token generation if the account has a provider, such as I did with Login.

asim: In this case Refresh function would be nil. Go does similar things in certain structs where they just do a nil check on a field and if its not present they ignore it, otherwise its executed. I think it cleans up the logic a bit.

ben-toogood: I'm not sure about this since it breaks down for non go usage, e.g. when I pass my token to the frontend or CLI I can't later call Refresh.

auth/options.go
asim: This is a bit confusing. There is an option called Credentials. So my assumption is auth.Credentials lets me set id/secret and these will be used for something. If thats not the case this should be removed.

asim: I see this as relevant for say my local CLI is loading credentials from disk and then using them as part of auth or its calling the Generate/Token methods but its best to remove them to remove ambiguity.

auth/auth.go
ben-toogood: This would require us to do no verification in Token(), since the raw account was have nothing to verify itself. I think it's cleaner to have the refreshtoken in account.

auth/options.go
ben-toogood: We need this method since we need a way to set the account ID and refresh token. What would you rather have the method called?

ben-toogood: This is used by cmd/config to set the services credentials for srv => srv requests. Services have no passwords, just refresh tokens.

auth/auth.go
asim: Theres no place I've ever seen that stores refresh token in an account because its a thing generated at the time of creating a new token. I don't think it makes sense in go-micro to store it as part of the Account struct. It does make sense to exist in the Token itself. And for the auth service to store the refresh token as part of the token separately to verify it.

auth/auth.go
asim: When you call the Auth.Token rpc method you'll get back a refresh token. It's left to the frontend to handle its own logic around refreshing (likely by calling the token method again with the refresh token). From solely a go-micro standpoint if you're talking just about interfaces and usage in code its just cleaner but I can share a refactor later.

auth/auth.go
```diff
@@ -33,7 +32,7 @@ type Auth interface {
 	// Options set for auth
 	Options() Options
 	// Generate a new account
-	Generate(id string, opts ...GenerateOption) (*Account, error)
+	Generate(id, secret string, opts ...GenerateOption) (*Account, error)
```
asim: Leave secret as an option

auth/options.go
```diff
@@ -10,14 +10,12 @@ import (
...
-	// Secret is used to generate new tokens
+	// Secret is used to authenticate the service
 	Secret string
 	// Token is the services token used to authenticate itself
 	Token *Token
-	// Public key base64 encoded
+	// PublicKey for decoding JWTs
 	PublicKey string
-	// Private key base64 encoded
-	PrivateKey string
```
asim, line 20: Leave public/private keypair? This is relevant for Auth itself to use.

auth/options.go
```diff
@@ -112,17 +121,59 @@ func NewGenerateOptions(opts ...GenerateOption) GenerateOptions {
 	return options
 }
 
+type LoginOptions struct {
```
asim: Remove LoginOptions considering we're just going to have the Token method?

auth/options.go
ben-toogood: Would we not want to set this in micro since go-micro has no use for it?

auth/auth.go
ben-toogood: Why so? Without a secret the account would be useless since no tokens can be generated?

auth/options.go
ben-toogood: Thanks, missed this

auth/auth.go
asim: Because I should be able to call Generate with an id and it sets me a unique secret. Its not a mandatory requirement since you get it back in account.

auth/options.go
asim: I think we've nuked a lot of stuff here which is going to come back e.g jwt implementation that needs public/private keypair or any auth implementation for that matter that makes use of this. I would leave it.

ben-toogood:
Approved by @asim

PR #1463, opened 2020-04-01
log.Errorf when pod streaming fails

crufter:


runtime/kubernetes/kubernetes_logs.go
```diff
@@ -32,8 +31,8 @@ func (k *klog) podLogStream(podName string, stream *kubeStream) {
 	}, client.LogParams(p))
 
 	if err != nil {
-		fmt.Fprintf(os.Stderr, err.Error())
-		return
+		stream.err = err
```
crufter, line 34: This is the key btw. This is where we set the err so it can be checked in runtime.Logs handler

Issue #1436, opened 2020-03-29
[BUG]应该需要改成mhttp.WithRegistry(s.opts.Registry),


yuanzhangcai:
https://github.com/micro/go-micro/blob/8282e781e46062c98c9cf8feaf1c015d4259d026/web/service.go#L271

应该需要改成mhttp.WithRegistry(s.opts.Registry),

crazybber:
@asim @vtolstov @ben-toogood   we can close this now.

PR #1468, opened 2020-04-02
fix: use registry from opts not use default directly:(#1436)

crazybber:
ref to : #1436


asim:
What if hte registry is nil?

crazybber:
> What if hte registry is nil?

u r right, nil need to be checked .i will fix it.

asim:
The registry is used elsewhere in the code. This is not enough.

crazybber:
yeah ,my careless , fix it in options.

crazybber:
@asim any comments ?

vtolstov:
Thanks, looks good. Minor thing comments to public funcs.

vtolstov:
Can you update branch and fix comments?

crazybber:
> Can you update branch and fix comments?

comments fixed.

vtolstov:
"This branch is out-of-date with the base branch" may be you need to pull go-micro master and merge it to this branch?

crazybber:
> "This branch is out-of-date with the base branch" may be you need to pull go-micro master and merge it to this branch?

comments also updated with merge

asim:
Has anyone tested this works with the existing web usage? Has anyone switched a registry without setting it to see if it works? Because I have previously seen a broken experience. Someone please test and confirm.

vtolstov:
i dont understand how this can broke something.
if user not set registry option all wark as before, if set - use another registry.
mostly we have two unneeded parts 
https://github.com/micro/go-micro/blob/master/web/service.go#L117
https://github.com/micro/go-micro/blob/master/web/service.go#L144

asim:
Specify --registry=etcd and see if the web service registry is actually set.

vtolstov:
cmdline and env parsing doing via micro.NewService, so i think that it not set with this option.
But if you set registry via go, why it can't be set?

asim:
go-micro/web uses config/cmd internally. So I believe if you don't use go-micro/registry/DefaultRegistry then it fails to initialise. Basically no one has tested the backwards compatibility of this PR and its been merged. Please test the usage of the flags and env vars.

vtolstov:
Ok, im test it

asim:
This change broke everything as I mentioned. It was merged untested. Please do not merge untested code that may have adverse effects everywhere. Future PRs will first go through test environment before being merged. Reverting https://github.com/micro/go-micro/pull/1489/files

PR #1469, opened 2020-04-02
Stop LogStream if there is an error in k8s pod log streaming

crufter:


asim:
You will need to add mutex in the stop method before select on the channel as apparently this is not thread safe.

crufter:
@asim :+1:

debug/log/kubernetes/stream.go
```diff
@@ -22,13 +23,16 @@ type kubeStream struct {
 	stream chan log.Record
 	// the stop chan
 	stop chan bool
+	sync.Mutex
```
asim: Place mutex above the thing you're locking, then when someone looks at the struct they know what its for :)

PR #1470, opened 2020-04-02
delete monitor

asim:


crazybber:
Why we delete monitor.

asim:
Because its not useful here.

PR #1472, opened 2020-04-02
minimize allocations

vtolstov:


vtolstov:
@asim i'm reread sources of http based and grpc based servers. And i don't think that changing advertise address after service starts is the right way. Conventional way to reconfigure service is run Init() with options. In this case previous server needs to be stopped and new started?

asim:
I don't care whats reconfigured. We will not Stop in Init. If someone decides to change something that required rebinding let them stop and start. I repeat, we will not stop the server in Init. Init is NOT the right place to create a side effect especially in the server. Remember the server is started by service.Run and majority of people call that. For anyone changing things they can start and stop the server themselves. DO NOT add complexity to the server.

vtolstov:
@asim i'm understand that pr relatively big and contains many parts.
Mostly my question - to you - what about caching service data for registration. Is it ok for you? Does anybody uses domain names in advertise addrs (if so, this pr can break them when they do roaming between networks)

asim:
I think if we want to be able to support moving cross networks caching that entry is going to break things.

vtolstov:
ok, i'm add fix to that

vtolstov:
@asim i'm add check. if user passes advertise address as hostname:ip, caching not done. If user pass ip:port as address or in advertise caching is enabled.
this can save a lot of memory:
```
ROUTINE ======================== github.com/micro/go-micro/v2/util/addr.Extract in github.com/micro/go-micro/v2@v2.4.0/util/addr/addr.go
         0   252.71MB (flat, cum) 86.71% of Total
         .          .     55:   // if addr specified then its returned
         .          .     56:   if len(addr) > 0 && (addr != "0.0.0.0" && addr != "[::]" && addr != "::") {
         .          .     57:           return addr, nil
         .          .     58:   }
         .          .     59:
         .    79.85MB     60:   ifaces, err := net.Interfaces()
         .          .     61:   if err != nil {
         .          .     62:           return "", fmt.Errorf("Failed to get interfaces! Err: %v", err)
         .          .     63:   }
         .          .     64:
         .          .     65:   //nolint:prealloc
         .          .     66:   var addrs []net.Addr
         .          .     67:   var loAddrs []net.Addr
         .          .     68:   for _, iface := range ifaces {
         .   172.36MB     69:           ifaceAddrs, err := iface.Addrs()
```

PR #1475, opened 2020-04-03
Auth integrate resolver to support micro web & api

ben-toogood:
Note: this implementation does mean the selector gets called twice per request, once by micro web/api and once by the auth service. **However** the selector caches the services, so I don't think this is an issue.


ben-toogood:
@asim (non-urgent) - I'm happy to merge this and the micro PR when you're happy. I'd like to open a second PR to fix the resolver inconsistency as documented in the PR, lmk your thoughts on using "Endpoint" to represent both web paths and rpc methods.

PR #1478, opened 2020-04-03
Fix auth hosts bug

ben-toogood:


api/server/auth/auth.go
```diff
@@ -101,8 +101,14 @@ func (h authHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 }
 
 func namespaceFromRequest(req *http.Request) (string, error) {
+	// determine the host, e.g. dev.micro.mu:8080
+	host := req.URL.Host
+	if len(host) == 0 {
+		host = req.Host
```
asim, line 107: Host can be host:port so this will need to be split before passing to parseIp

Issue #1473, opened 2020-04-03
service.Service and micro.Service


crazybber:
service.Service https://github.com/micro/go-micro/blob/master/service/service.go#L10
micro.Service https://github.com/micro/go-micro/blob/master/micro.go#L16

they have compatible signature  and **same with option**

should we unify  these interface to [go-micro/service/service.go](https://github.com/micro/go-micro/blob/master/service/service.go) or you guys have other design consideration and why  ?

@asim

asim:
We will unify but haven't decided on the right approach. Maybe using import . Or aliasing

crazybber:
a preview PR proposed  , https://github.com/micro/go-micro/pull/1484  which discussion can go  from

crazybber:
talking will be followed up in PR: https://github.com/micro/go-micro/pull/1484

PR #1484, opened 2020-04-04
[discussion][pre-propose]unify micro interface

crazybber:
refer to :  https://github.com/micro/go-micro/issues/1473 Topic : service.Service and micro.Service

**here is a proposed preview.**

how do you think about this preview?

@asim @vtolstov @jakexks @ben-toogood

can we go directly or we should go in another way or more something else need to be done ?

### DIFF
--- a/api/handler/options.go
+++ b/api/handler/options.go
@@ -3,6 +3,7 @@ package handler
 import (
 	"github.com/micro/go-micro/v2"
 	"github.com/micro/go-micro/v2/api/router"
+	mservice "github.com/micro/go-micro/v2/service"
 )
 
 var (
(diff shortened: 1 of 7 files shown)


crazybber:
@asim  @vtolstov @jakexks  @ben-toogood

asim:
Not sure about this yet. Thinking maybe a dotted import might make more sense so top level inherits but one to think about. Considering how central a Service is I'm more tempted to remove the package itself. It was only put there for non command line parsed initialisation but considering we're moving to the Config interface it probably make sense to strip the package.

asim:
I'm going to close for now. I think what we're going to do is continue to have the top level micro.Service and just leverage the internals.

PR #1485, opened 2020-04-04
errors: add helper methods

vtolstov:

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -15,6 +15,31 @@ func (e *Error) Error() string {
 	return string(b)
 }
 
+func (e *Error) BadRequest(format string, a ...interface{}) error {
+	return BadRequest(e.Id, format, a...)
+}
+func (e *Error) Unauthorized(format string, a ...interface{}) error {
+	return Unauthorized(e.Id, format, a...)
+}
+func (e *Error) Forbidden(format string, a ...interface{}) error {
+	return Forbidden(e.Id, format, a...)
+}
+func (e *Error) NotFound(format string, a ...interface{}) error {
+	return NotFound(e.Id, format, a...)
+}
+func (e *Error) MethodNotAllowed(format string, a ...interface{}) error {
+	return MethodNotAllowed(e.Id, format, a...)
+}
+func (e *Error) Timeout(format string, a ...interface{}) error {
+	return Timeout(e.Id, format, a...)
+}
+func (e *Error) Conflict(format string, a ...interface{}) error {
+	return Conflict(e.Id, format, a...)
+}
+func (e *Error) InternalServerError(format string, a ...interface{}) error {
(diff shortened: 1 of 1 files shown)


vtolstov:
btw. may be also need to add NewError method that receives only Id and returns *Error?

asim:
What's the usage model here? I need example code to understand why this is needed. Do we actually need these helper methods? If you read the godoc for Error it now goes from just Error() with some structs for 10 methods. It should not exist on the error itself.

vtolstov:
use case
type Handler struct {
  db *sql.DB
  err *errors.Error
}

main.go
handler := &handlers.Handler{db: db, err: &errors.Error{Id: "my_service_name"}}
pb.RegisterHandlerxxxx

handler.go

func (h *Handler) Call(ctx ....) error {
  if err := h.db.Query(...); err != nil {
    return h.err.NotFound("req not found")
  }
}

so caller always get proper micro error with filled id.

asim:
It doesn't feel idiomatic to how we're already doing things or how Go does things. 

https://blog.golang.org/go1.13-errors
https://godoc.org/github.com/pkg/errors

We defined specific errors as functions to help but what we do not want to do is define these rpc error types on the error itself. We want ways to wrap/unwrap and compare cleanly.

vtolstov:
mostly i'm bored  when need to specify for each error Id. How to deal? i want tha all my errors have one id and not think about it, only use.

asim:
Specify your own error and use that?

asim:
Key thing to remember. Not everything needs to be part of go-micro. The framework doesn't need to deal with stuff that you could very easily maintain in your own code or your own util package. It's not our responsibility to inherit all of the users baggage. 

Going to close for now.

vtolstov:
> Specify your own error and use that?

if all users must specify they own errors we go to go-kit and this is bad. 
If logger produce log , why some interface can't produce errors?

asim:
I think the framework needs to be responsible for framework errors but they're scoped as functions right now. We can maybe define them as variables so they're compatible and have a comparison function but it's best to operate in a model much like Go here. Any higher level errors outside of the framework are not our concern right now. Even for our services in micro/services we'll define an entirely separate set of errors.

vtolstov:
mostly i want to do that not only for user case, but for our too. so user create micro.NewService with micro.Name, so we can create internal var inside server struct, that holds name. So all errors looks not like go.micro.server xxxx from all services, but comes with specific id. So it's easy distinglish source of problem in multiservice thinkg.

asim:
Setting the service names for errors makes sense but its a broader thing right. Like not requiring someone to return a micro error from a handler but allowing us to wrap the error by default. This is something I think needs a design doc to actually discuss the expected outcome rather than just ad-hoc merging something. development/design/errors.md

PR #1486, opened 2020-04-04
api/router/static: allow to specify body dst

vtolstov:


vtolstov:
add body field to api.Endpoint and it usage in rpc handler.

vtolstov:
now we have feature parity with grpc-gateway i think.

api/handler/rpc/rpc.go
```diff
@@ -300,13 +300,17 @@ func requestPayload(r *http.Request) ([]byte, error) {
...
 	matches := make(map[string]interface{}, len(md))
+	bodydst := ""
 
 	// get fields from url path
 	for k, v := range md {
 		// filter own keys
 		if strings.HasPrefix(k, "x-api-field-") {
 			matches[strings.TrimPrefix(k, "x-api-field-")] = v
 			delete(md, k)
+		} else if k == "x-api-body" {
```
asim, line 311: What sets these metadata values?

api/api.go
```diff
@@ -40,6 +40,10 @@ type Endpoint struct {
 	Method []string
 	// HTTP Path e.g /greeter. Expect POSIX regex
 	Path []string
+	// Body destination
+	// "*" or "" - top level message value
+	// "string" - inner message value
+	Body string
```
asim, line 46: Why do we need body?

vtolstov: rpc endpoint need to say how to use body, it can be used as full message or as it part in case of nested message.
for ex:
```
message AccountUpdateReq {
  Account account = 2;
  ....
}

message Account {
  string id = 1;
  string email = 2;
  string birthday = 3;
  ...
}
```

so you can write in proto
```
rpc AccountUpdate(AccountUpdateReq) returns (AccountUpdateRsp) {
  option (google.api.http) = { put: "/api/account/{account.id}"; body: "account" };
}
```

in this case id in struct will be filled via path param, and other fields from body

api/handler/rpc/rpc.go
vtolstov: they set by router (i'm plan to add it also for registry router, now it filled only in static router)

asim: Should the router not just set some informational struct in context?

asim: Should it not just maybe pass in the endpoint in context?

asim:
The branch needs to be rebased with master

api/handler/rpc/rpc.go
vtolstov: i don't have any objections about using router context, this code mostly based on decision in slack when chat with you. We want to have something that working and iterate, request context used because it already present. if we use dedicated router context - this save some memory to avoid growing map

vtolstov:
> The branch needs to be rebased with master

rebased

api/handler/rpc/rpc.go
asim: Lets merge this first iteration then move to using a context passed through

vtolstov:
ok, thanks, i'm combine with adding similar stuff to registry router, and write full test case.

PR #1491, opened 2020-04-06
Disk backed local store

crufter:
I took the memory tests (and codebase) and tried to make as minor modifications as possible in the first round.

Notes:
- Rather ineffective now (each read is a separate file access, that could be improved), but focus was on conforming to spec.
- Init() was doing flush in the in memory version, but obv. I don't want to delete the data on init as it aims to be persistent, so in the test I had to delete file after each Init.


store/file/file.go
```diff
@@ -0,0 +1,351 @@
+// Package local is a file system backed store
+package local
```
asim: rename package to file

PR #1492, opened 2020-04-06
Rename store Namespace / Prefix options to Database and Table

jakexks:


jakexks:
Don't merge yet, as this changes a command line flag which is breaking.

service.go
```diff
@@ -108,9 +108,9 @@ func (s *service) Init(opts ...Option) {
 
 		// If the store has no namespace set, fallback to the
 		// services name
-		if len(store.DefaultStore.Options().Namespace) == 0 {
+		if len(store.DefaultStore.Options().Database) == 0 {
 			name := s.opts.Cmd.App().Name
-			store.DefaultStore.Init(store.Namespace(name))
+			store.DefaultStore.Init(store.Database(name))
```
asim: This should be set to Table name since we're having table per service rather than database per service

PR #1496, opened 2020-04-07
Configurable Namespace & Public Suffix Domain Resolution

ben-toogood:


asim:
I'm struggling with this PR. Something just doesn't feel right. A lot of the resolution logic has ended up in an auth wrapper. Rather than it being wrapped up in a resolver its in this wrapper itself and then we have to do that resolution again in a number of places. It feels like this wrapper should only be making one request to Resolve and be given an endpoint that has the exact info it needs rather than having to map out a bunch of stuff itself. 

I think if resolver.Resolve returned the Namespace with the endpoint or resolver had the namespace itself that might simplify things. And the Name would be returned as the full namespace.type.alias.

ben-toogood:
We didn't want to have the auth namespace equal the service name prefix, e.g the namespace is micro, but the services don't have to be named "micro.web.xxx". I'll can move the namespace resolution logic to the resolver, although I'm not sure this is the best place for it since it is auth specific?

api/resolver/resolver.go
```diff
@@ -11,6 +11,12 @@ var (
 	ErrInvalidPath = errors.New("invalid path")
 )
 
+// NamespaceResolver resolves request to the namespace
+type NamespaceResolver interface {
```
asim: Why not just add Namespace to the Endpoint thats resolved in the current resolver?

api/server/auth/auth.go
```diff
@@ -69,14 +54,7 @@ func (h authHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
 	if err != nil {
-		acc = &auth.Account{Namespace: namespace}
-	}
-
-	// Check the accounts namespace matches the namespace we're operating
-	// within. If not forbid the request and log the occurance.
-	if acc.Namespace != namespace {
-		logger.Debugf("Cross namespace request warning: account %v (%v) requested access to %v in the %v namespace", acc.ID, acc.Namespace, req.URL.Path, namespace)
-		// http.Error(w, "Forbidden namespace", 403)
+		acc = &auth.Account{}
```
asim: What kind of issues will be caused by having an empty account here?

ben-toogood: We use an empty account just so we can call Verify. It doesn't cause any issues.

PR #1498, opened 2020-04-07
Use file store by default (as opposed to memory store)

crufter:


store/file/file.go
```diff
@@ -17,6 +17,14 @@ import (
 	"github.com/pkg/errors"
 )
 
+var (
+	// DefaultDatabase is the namespace that the bbolt store
+	// will use if no namespace is provided.
+	DefaultDatabase = "micro"
```
asim, line 23: You may also want a DefaultTable called micro

crufter: Makes sense

PR #1499, opened 2020-04-07
Setting up file store in constructor and not in init which is optional, make memory store the default again

crufter:
Testing this locally currently.

### DIFF
--- a/store/file/file.go
+++ b/store/file/file.go
@@ -44,7 +44,6 @@ type fileStore struct {
 }
 
 func (m *fileStore) Init(opts ...store.Option) error {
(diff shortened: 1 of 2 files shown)


store/file/file.go
```diff
@@ -46,6 +44,11 @@ type fileStore struct {
 }
 
 func (m *fileStore) Init(opts ...store.Option) error {
+	m.options = store.Options{}
```
asim: Do not blow away the options. Only some of them may be included in the update.

crufter: Cheers I was actually thinking about this

PR #1500, opened 2020-04-08
add Store Close method

asim:


crufter (approved):

store/file/file.go
```diff
@@ -223,6 +223,13 @@ func (m *fileStore) set(r *store.Record) error {
 	})
 }
 
+func (m *fileStore) Close() error {
+	if m.db != nil {
+		return m.db.Close()
```
crufter, line 228: Perhaps do `m.db = nil`?

PR #1510, opened 2020-04-09
Enabling micro run for subfolders

crufter:
To reproduce the issue:

- Step into `micro/services` repo locally.
- Issue `micro run account/api`.

You will see:

```
$ micro run account/api
go: cannot find main module, but found .git/config in /home/crufter/code/src/github.com/micro/services
	to create a module there, run:
	go mod init
```

This is because micro run issues `go run` in the current folder (ie. `micro/services`) rather than in the subfolder. This PR fixes that by setting the `Dir` property for `cmd`s executed.

Context: The reason I needed to do `micro run [subfolder]` is so micro run correctly sets the name of the service.


runtime/local/process/os/os.go
```diff
@@ -15,12 +15,15 @@ import (
 
 func (p *Process) Exec(exe *process.Executable) error {
 	cmd := exec.Command(exe.Package.Path)
+	cmd.Dir = exe.WorkDir
```
crufter: This might not be needed as Exec looks neglected?

runtime/default.go
```diff
@@ -195,6 +195,12 @@ func (r *runtime) Create(s *Service, opts ...CreateOption) error {
 		o(&options)
 	}
 
+	dir, err := filepath.Abs(filepath.Dir(os.Args[0]))
```
crufter: os.Args feels heavy handed when we have proper cli args etc in outer layers. Suggestions welcome.

runtime/options.go
```diff
@@ -66,6 +66,8 @@ type CreateOptions struct {
 	Retries int
 	// Specify the image to use
 	Image string
+	// Initial working directory for command
+	WorkDir string
```
asim: This is probably not needed since you have source

runtime/local/process/process.go
```diff
@@ -26,6 +26,8 @@ type Executable struct {
 	Env []string
 	// Args to pass
 	Args []string
+	// Initial working directory
+	WorkDir string
```
asim: Rename to Dir?

PR #1524, opened 2020-04-11
broker: swap default broker from eats to http

vtolstov:


vtolstov:
@asim why we have Handle option in go-plugins/broker/http ? i don't see any usage of it

asim:
It was to enable adding a handler for things like prometheus

config/cmd/cmd.go
```diff
@@ -318,7 +318,7 @@ var (
 	DefaultBrokers = map[string]func(...broker.Option) broker.Broker{
 		"service": brokerSrv.NewBroker,
 		"memory":  memory.NewBroker,
-		"nats":    nats.NewBroker,
```
asim, line 321: Still need this

vtolstov:
do we need this option too? i'm ask because this is not default broker option, but it exposed in broker package

vtolstov:
may be move it to go-micro/broker/http ?

config/cmd/cmd.go
asim: Actually never mind, we'll move the imports into micro itself

asim:
yea add the options

vtolstov:
done

broker/http.go
```diff
@@ -0,0 +1,709 @@
...
+}
+
+type httpEvent struct {
+	m   *Message
+	t   string
+	err error
+}
+
+var (
+	DefaultSubPath   = "/_sub"
```
asim: /

broker/http.go
```diff
@@ -0,0 +1,709 @@
...
+	serviceName      = "go.micro.http.broker"
```
asim: s/go//

broker/http.go
```diff
@@ -0,0 +1,709 @@
...
+
+	var m *Message
+	if err = h.opts.Codec.Unmarshal(b, &m); err != nil {
+		errr := merr.InternalServerError("go.micro.broker", "Error parsing request body: %v", err)
+		w.WriteHeader(500)
+		w.Write([]byte(errr.Error()))
+		return
+	}
+
+	topic := m.Header[":topic"]
```
asim: s/:topic/Micro-Topic/

asim:
Merge?

vtolstov:
waiting for ci

PR #1526, opened 2020-04-11
api minor improvements

vtolstov:
* api/handler/rpc: unblock all http methods and set Host meta
* api/router/static: add debug log


vtolstov:
@asim https://github.com/micro/go-micro/pull/1526/files#diff-b3e297584c2a86f2612ea1d663b963acR103
this line changed because with non nil router i can handle post/put/patch requests with api

PR #1530, opened 2020-04-14
api/router/registry: extract path based parameters from url to req

vtolstov:


vtolstov:
this pr depends on #1531 , so they need to be merged first and this rebased

PR #1537, opened 2020-04-15
Add Priority to auth rules

ben-toogood:


asim:
Can you please document in the proto or somewhere the ordering of priority

PR #1547, opened 2020-04-17
Runtime Namespace

ben-toogood:


runtime/kubernetes/kubernetes.go
```diff
@@ -24,20 +24,59 @@ type kubernetes struct {
 	closed chan bool
 	// client is kubernetes client
 	client client.Client
+	// namespaces which exist
+	namespaces []client.Namespace
+}
+
+// namespaceExists returns a boolean indicating if a namespace exists
+func (k *kubernetes) namespaceExists(name string) (bool, error) {
```
asim, line 32: Are we still making use of the "default" namespace if none exists or switching everything to a "micro" namespace?

asim:
LGTM. Waiting for one more look from Jake.

PR #1548, opened 2020-04-18
simple multicast registry

vtolstov:


asim:
I don't think this is what we need. We want to optimise mdns. We still want to use a standard to discovery.

vtolstov:
Why you want mdns?what benefits do you have with it?With plain multicast we
can avoid double encoding, and have very simple registry that uses all
other core micro stuff

сб, 18 апр. 2020 г., 11:43 Asim Aslam <notifications@github.com>:

asim:
Because I want to operate on standards as opposed to ad hoc custom things. What I learned from my client/server experience is that inevitably standards win out. And in the case of local discovery mdns is a staple. So I only want to optimise. I also see further long term benefits to it.

vtolstov:
ok,

PR #1562, opened 2020-04-23
Runtime - Image Pull Secrets

ben-toogood:


runtime/options.go
```diff
@@ -101,6 +103,13 @@ func CreateImage(img string) CreateOption {
 	}
 }
 
+// CreateImagePullSecret sets a secret to use
+func CreateImagePullSecret(secrets ...string) CreateOption {
+	return func(o *CreateOptions) {
+		o.ImagePullSecrets = append(o.ImagePullSecrets, secrets...)
```
jakexks, line 109: append() on a "create" function feels weird to me

PR #1570, opened 2020-04-24
Micro log fixes

crufter:


runtime/service/service.go
```diff
@@ -90,14 +89,25 @@ func (s *svc) Logs(service *runtime.Service, opts ...runtime.LogsOption) (runtim
...
-			err := ls.RecvMsg(&record)
-			if err != nil {
-				log.Error(err)
+			select {
+			// @todo this never seems to return, investigate
+			case <-ls.Context().Done():
+				return
+			default:
+				record := pb.LogRecord{}
+				err := ls.RecvMsg(&record)
```
crufter, line 115: This default cause will actually block, I think I have to turn this blocking part into a channel send.

PR #1573, opened 2020-04-26
delete invalid copy

Allenxuxu:


asim:
Why,?

Allenxuxu:
All places using the `Send` method have been copied `Message` once, and there will be no other goroutines to modify `m`.

```
// write back some 200
if err := sock.Send(&transport.Message{
		Header: msg.Header,
}); err != nil {
	gerr = err
	break
}
```

We should reduce memory copy to indicate speed.

PR #1595, opened 2020-04-29
Auth Client Wrapper

ben-toogood:


auth/options.go
```diff
@@ -79,6 +79,13 @@ func Credentials(id, secret string) Option {
 	}
 }
 
+// ClientToken sets the auth token to use when making requests
```
ben-toogood, line 82: I had to name this `ClientToken` because Token is already being used. Open to suggestions for a better name.

PR #1596, opened 2020-04-30
unsubscribe can async

socket515:
unsubscribe can async, will make server fast close.


asim:
It's really not worth the optimization

PR #1602, opened 2020-05-05
Design: model interface ideas

asim:
## Overview

This is a design idea for a "model" interface which serves as a data model for accessing and storing data. It layers on store/sync/codec as client/server do on top of registry/broker/transport/codec.

## Design

We're using the entity relationship model to represent our model interface https://en.wikipedia.org/wiki/Entity%E2%80%93relationship_model. The idea is to create a simple high level CRUD for storage much like the client does for RPC calls. The entity represents access to a specific domain.

The Model interface is inline with the Client and Server interfaces and further encapsulated by the Service {Client, Server, Model} to encompass the overall requirements for a service.

Model encapsulates:
- Store - for the data storage
- Sync - for locking and synchronization
- Codec - for serialisation / marshaling

## Proto

We code generate service to service calls to reduce the boilerplate usage of the client/server. Considering the verbosity of CRUD development and repitition we want to encapsulate this concern in a similar way. The protoc-gen-micro command should code generate usage of the Model to create out of the box CRUD interfaces for storage.

```
// in the proto
service User {}
```

generates

```
// Generate a new user mode
NewUserModel(model.Model) UserModel

// Generates a register method
RegisterUserModel(model.Model)
```

Additionally the UserModel will include access and storage methods much like an ORM.

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -1,0 +1,57 @@
+// Package model is a data modelling interface
+package model
+
+import (
+	"github.com/micro/go-micro/v2/codec"
+	"github.com/micro/go-micro/v2/store"
+)
+
+// Model represents a data model interface for data access
+type Model interface {
+	// Initialise options
+	Init(...Option) error
+	// Retrieve the options
+	Options() Options
+	// Register a data type
+	Register(v interface{}) error
+	// Create an entity
+	Create(v interface{}, opts ...CreateOption) error
+	// Read an entity
+	Read(v interface{}, opts ...ReadOption) error
+	// Update an enity
+	Update(v interface{}, opts ...UpdateOption) error
+	// Delete an entity
+	Delete(v interface{}, opts ...DeleteOption) error
+	// The implementation e.g crud
+	String() string
+}
+
+// Options to the model
+type Options struct {
+	// The store for data storage
+	Store store.Store
+	// The codec for marshalling
+	Codec codec.Marshaler
+}
+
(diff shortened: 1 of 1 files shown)


model/model.go
```diff
@@ -0,0 +1,57 @@
...
+	// The implementation e.g crud
+	String() string
+}
+
+// Options to the model
+type Options struct {
+	// The store for data storage
+	Store store.Store
+	// The codec for marshalling
+	Codec codec.Marshaler
```
domwong, line 43: I'm not crazy about passing the `micro/go-micro/codec` since that's designed around messages whereas here we're talking about models. 

I'm still arguing in my own head about whether I like the idea of a codec at al or would rather prefer to have a `Marshal/Unmarshal` funcs on the model interface itself. I think I need to have a longer think about possible use cases that would require different marshalling depending on circumstances

asim: We tend to pass around interfaces for reuse, its useful to have codecs consolidated in once place. Especially if we're saying types are mostly generated from proto and we want to use the proto codec to serialise. This should ideally be hidden from the user, as in the default codec would be proto and the user would just call model.Create with their type and the model would take care of the encoding for them.

asim: We could potentially expose the marshaler on the model but I think since its an internal thing done by the model itself it might not make sense.

asim:
Useful design ideas but unlikely to be merged in current state

PR #1604, opened 2020-05-06
WIP: Add metadata to store record

asim:


store/cockroach/cockroach.go
```diff
@@ -121,6 +122,12 @@ func (s *sqlStore) initDB(database, table string) error {
 		return err
 	}
 
+	// Create Metadata Index
+	_, err = s.db.Exec(fmt.Sprintf(`CREATE INDEX IF NOT EXISTS "%s" ON %s.%s USING GIN ("metadata");`, "metadata_index_"+table, database, table))
```
sadwxqezc, line 126: I'm curious  will the result of  `map=>JSONB` change with the upgrade of depend libs? Because that will effect the use of metadata index。

sadwxqezc (commented):
LGTM

PR #1622, opened 2020-05-13
Registry mutli-tenancy

ben-toogood:
Refactored the auth wrapper to prevent a loop when using the service client for the runtime clients. Now a token is generated when a service starts and the auth wrapper will only refresh it once it expires.

The registry now uses the service client when making RPC's so that the registry service can register it in the correct namespace.


ben-toogood:
Approved by Asim.

PR #1621, opened 2020-05-13
fix: update dependency certmagic

x1nchen:
module github.com/mholt/certmagic has been renamed github.com/caddyserver/certmagic,
so upgrade on this module will fail.

fix: https://github.com/micro/micro/issues/835

PR #1625, opened 2020-05-13
fix: update dependency certmagic

x1nchen:
module github.com/mholt/certmagic has been renamed github.com/caddyserver/certmagic,
so upgrade on this module will fail.

fix: micro/micro#835

caddyserver/certmagic@v0.10.6 is Maximum upgradeable version with go version 1.13

Higher version use *tls.ClientHelloInfo.SupportsCertificate which only supported in go 1.14


x1nchen:
@asim Hi, this PR refers to https://github.com/micro/go-micro/pull/1621

Issue #1560, opened 2020-04-23
[BUG] Data races when using two services


PatrLind:
I am trying to run a micro.Service together with a web.Service. Sometimes I get data race errors in different places inside the code. Usually around the logging or option handling.
Perhaps I am not using the framework correctly, in that case I would like to know an alternative way to have a micro.Service together with a web.Service.

I am using: github.com/micro/go-micro/v2 v2.4.0

This code usually reproduces the issue:
```
package main

import (
	"context"
	"fmt"
	"sync"
	"time"

	"github.com/micro/go-micro/v2"
	"github.com/micro/go-micro/v2/logger"
	"github.com/micro/go-micro/v2/web"
)

func main() {
	for i := 0; i < 10; i++ {
		fmt.Println("Test nr", i)
		testFunc()
	}
}

func testFunc() {
	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond*250)
	defer cancel()
	s := micro.NewService(
		micro.Name("test"),
		micro.Context(ctx),
		micro.HandleSignal(false),
	)
	w := web.NewService(
		web.MicroService(s),
		web.Context(ctx),
		web.HandleSignal(false),
	)
	s.Init()
	w.Init()

	var wg sync.WaitGroup
	wg.Add(2)
	go func() {
		defer wg.Done()
		err := s.Run()
		if err != nil {
			logger.Errorf("micro run error: %v", err)
		}
	}()
	go func() {
		defer wg.Done()
		err := w.Run()
		if err != nil {
			logger.Errorf("web run error: %v", err)
		}
	}()

	wg.Wait()
}
```

**Environment:**
go version go1.14 windows/amd64

vtolstov:
Please, provide race log

PatrLind:
It is a bit random, but here are two:

```
==================
WARNING: DATA RACE
Write at 0x00c000073400 by main goroutine:
  github.com/micro/go-micro/v2/web.(*service).Init.func1()
      C:/Users/patrik/go/pkg/mod/github.com/micro/go-micro/v2@v2.4.0/web/service.go:343 +0x458
  github.com/micro/cli/v2.(*App).RunContext()
      C:/Users/patrik/go/pkg/mod/github.com/micro/cli/v2@v2.1.2/app.go:312 +0xab0
  github.com/micro/cli/v2.(*App).Run()
[...]

PR #1644, opened 2020-05-19
Fixes for #1560

PatrLind:
This should fix some issues reported for https://github.com/micro/go-micro/issues/1560


PatrLind:
I am not so experienced with how PRs work in git so I hope I did the right thing.

PR #1648, opened 2020-05-20
catch default config init errors

colinhoglund:
Constructing a `NewConfig` with source options can panic in certain situations due to the default config constructor not handling `Init()` errors.

Example: trying to load a non-existent file source
```
  1 package main
  2
  3 import (
  4   "github.com/micro/go-micro/v2/config"
  5   "github.com/micro/go-micro/v2/config/source/file"
  6 )
  7
  8 func main() {
  9   c, err := config.NewConfig(config.WithSource(file.NewSource(file.WithPath("thisdoesnotexist"))))
 10   if err != nil {
 11     panic(err)
 12   }
 13
 14   c.Map()
 15 }
```

calling `Map()` panics because `config.vals` is `nil`
```
$ go run main.go
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x1599c70]

goroutine 1 [running]:
github.com/micro/go-micro/v2/config.(*config).Map(0xc000168280, 0x0)
	/Users/colinhoglund/go/pkg/mod/github.com/micro/go-micro/v2@v2.7.0/config/default.go:134 +0x70
main.main()
	/tmp/microtest/main.go:14 +0xdb
exit status 2
```


asim:
Our config loading is a background process so we don't really want to do this. The file might appear in future as opposed to being there right now. Ideally we just want to address the panic.

colinhoglund:
> Our config loading is a background process so we don't really want to do this. The file might appear in future as opposed to being there right now. Ideally we just want to address the panic.

@asim Gotcha, I didn't notice that `nil` `vals` were being handled elsewhere in the package. Does it make more sense then to simply check for `nil` wherever `vals` is used?

something like
```
	if c.vals != nil {
		return c.vals.Map()
	}

	return make(map[string]interface{})
```

asim:
Yea I think its important to ensure vals is not nil but we can maybe just pre-initialise someting so it doesn't have to be evaluated everywhere

colinhoglund:
@asim Thanks for the feedback!

I dug a little deeper and it looks like the `Init()` error was coming from the `c.opts.Loader.Load(c.opts.Source...)` call, which would explain why `c.vals` is not being initialized.

https://github.com/micro/go-micro/blob/3d36398818f4f9f646b2e2f0268eec27922a1ccb/config/default.go#L52-L65

I pushed a new commit that gathers errors during `Init()` rather that returning immediately after failure. Let me know if this looks better.

colinhoglund:
> I pushed a new commit that gathers errors during `Init()` rather that returning immediately after failure.

@asim Any thoughts on this approach?

domwong:
Two things here.

Firstly, if `Init()` errors I wouldn't expect anyone to try and do anything else with it. Carrying on and collecting all the errs feels like non-obvious behaviour. So on that basis I'm inclined to not merge this PR. 

Secondly, as you've seen `NewConfig()` doesn't return error ever. This is on purpose as the other `New*` funcs across the framework don't return errors either. BUT `NewConfig()` actually has `error` in its return signature which is wrong. We'll add a note to the func to let users know this is the case and we'll remove it from the sig in the next major release as a breaking change. 

For you @colinhoglund I think you're best to do the init yourself and do essentially what `NewConfig()` is doing yourself but check for errors from `Init`
```
var c config
c.Init(opts...)
go c.run()
```

colinhoglund:
> Carrying on and collecting all the errs feels like non-obvious behaviour

I totally agree, I noticed this used elsewhere in the code base so I figured I'd commit it as a suggestion since it seemed better than an unhandled panic.

> I think you're best to do the init yourself

Thanks, this works for my use case to avoid running into a panic. So if this doesn't seem like something worth fixing feel free to close this PR.

domwong:
Thanks for the understanding, will be closing rather than merging.

PR #1654, opened 2020-05-21
Auth Improvements

ben-toogood:
- Add Rules() to the auth interface and update Grant/Revoke to accept Rule as an argument.
- Remove unused errors from the auth interface
- Change account.Namespace to account.Scopes so a single account can have multiple scopes.
- Refactor rule checking to auth/rules so all implementations are checking rules the same way on Verify.
- Improve test coverage for rule checking and the auth wrapper.


auth/auth.go
```diff
@@ -99,15 +69,47 @@ type Token struct {
...
+	// AccessGranted to a resource
+	AccessGranted Access = iota
+	// AccessDenied to a resource
+	AccessDenied
 )
 
+// Rule is used to verify access to a resource
+type Rule struct {
+	// ID of the rule, e.g. "public"
+	ID string
```
domwong, line 106: Is this unique to environment? 

e.g. spotify/staging and spotify/prod  can have rules with the same ID?

auth/jwt/jwt.go
```diff
@@ -80,84 +71,44 @@ func (j *jwt) Generate(id string, opts ...auth.GenerateOption) (*auth.Account, e
 	return account, nil
 }
 
-func (j *jwt) Grant(role string, res *auth.Resource) error {
+func (j *jwt) Grant(rule *auth.Rule) error {
 	j.Lock()
 	defer j.Unlock()
-	j.rules = append(j.rules, &rule{role, res})
+	j.rules = append(j.rules, rule)
```
domwong, line 77: Should we check for dupes by checking ID?

auth/rules/rules.go
```diff
@@ -0,0 +1,93 @@
...
+	validTypes := []string{"*", res.Type}
+
+	// the rule is only to be applied if the name matches the resource or is catch-all (*)
+	validNames := []string{"*", res.Name}
+
+	// rules can have wildcard excludes on endpoints since this can also be a path for web services,
+	// e.g. /foo/* would include /foo/bar. We also want to check for wildcards and the exact endpoint
+	validEndpoints := []string{"*", res.Endpoint}
+	if comps := strings.Split(res.Endpoint, "/"); len(comps) > 1 {
+		for i := 1; i < len(comps); i++ {
```
domwong: If endpoint = foo/bar/baz you'll get `[*, foo/bar/baz, foo/*, foo/bar/*]` - do you want `foo/bar/baz/*` too?

auth/service/proto/auth.proto
```diff
@@ -35,25 +35,22 @@ message Token {
 message Account {
 	string id = 1;
 	string type = 2;
-	repeated string roles = 3;
 	map<string, string> metadata = 4;
-	string namespace = 5;
-	string provider = 6;
+	repeated string scopes = 5;
```
domwong, line 39: Be careful, this is a breaking change

domwong: Well, this whole file is really

auth/service/service.go
```diff
@@ -79,90 +77,61 @@ func (s *svc) Generate(id string, opts ...auth.GenerateOption) (*auth.Account, e
...
+			Access:   access,
+			Resource: &pb.Resource{
+				Type:     rule.Resource.Type,
+				Name:     rule.Resource.Name,
+				Endpoint: rule.Resource.Endpoint,
+			},
 		},
 	})
+
+	go s.loadRules(s.options.Namespace)
```
domwong: Haven't checked for err in previous call. May or may not be an issue.

domwong (approved):
I think this looks fine. Some minor comments. 

Broader comment is that there are a lot of changes to protos which would cause existing clients to break. Something that we should be aware of and decide when we're happy to make such changes vs not

auth/auth.go
ben-toogood: Yes

auth/jwt/jwt.go
ben-toogood: Since the JWT isn't backed by the store, the rules are normally hardcoded in the service (like we do in micro/micro/internal/auth), so I don't think this is really necessary.

auth/rules/rules.go
ben-toogood: No, good catch. I'll amend.

auth/service/proto/auth.proto
ben-toogood: Yes, since we're still using this only internally it's okay to do the breaking change as i'll bump all the dependancies when we roll this out.

sadwxqezc (commented):
There may be a problem in auth/jwt/jwt.go 85

I guess the correct code should be like：

```
	rules := []*auth.Rule{}
	for _, r := range j.rules {
		if r.ID != rule.ID {
			rules = append(rules, r)
		}
	}
```

PR #1657, opened 2020-05-26
logger caller not trim in windows

xpunch:
uber-go/zap use package/file style caller, I think this will be more useful when last file name is very simple.
zap/logger([zap/logger.go](https://github.com/uber-go/zap/blob/9a9fa7d4b5f07a9b634983678a65b5525f81e58b/zapcore/entry.go#L101))


domwong (approved):
Looks like a good change. Thanks @xpunch

Issue #1641, opened 2020-05-15
"importing 'github.com/go-git/go-git/v5'" introduced MegaByte size dependences


masknu:
https://github.com/micro/go-micro/blob/fccab8ad27d13cc366180bbcaa9c1341b37b8b78/runtime/local/git/git.go#L12

"go-git/v5" depends on "go-billy/v5" which depends on "go-git-fixtures/v4", and go-git-fixtures includes massive .tgz files in it.
It takes too long to build a micro project due to downloading these files.

crufter:
@masknu we will look into this! Thanks

vtolstov:
yes, this is really bad thing

mcuadros:
go-git-fixtures it's a test dependency of go-git. The tgz files are inside of the zip created by proxy.golang.org even when they do not need it since they are embedded in a data file.

The file is 47mb `https://proxy.golang.org/github.com/go-git/go-git-fixtures/v4/@v/v4.0.1.zip` and takes around 8s, the server doesn't look super fast, either. So I agree it's not convenient, even more, when it's a test dependency.

What is even worst is that you have as well the go-git-fixtures from the old org `src-d`, so you have the package twice.

From go-git org, we will try to find a solution to this problem. Feel free to open an issue.

vtolstov:
@mcuadros thanks! i think that best of all - create dedicated test repo with fixtures.

vtolstov:
fixed

PR #1658, opened 2020-05-27
go modules cleanup and remove wrong self import to v1

mcuadros:
This PR's fix some problems with the go modules dependencies:
- Update code to use go-git/go-git.v5 since ` runtime/local/source/git/git.go` was still using the old version.
- Remove a self-dependency of go-micro to v1, I guess by error introducing by the autocomplete tool.

This has removed a lot of dependencies on old packages and for sure reduced the download time. Related to #1641. 

Also I didn't a bit of code improvement on `runtime/local/git/git.go` that is kind of unrelated to the purpose of this PR, let me know if you rather have this PR splitter in two.


go.mod
```diff
@@ -18,24 +23,28 @@ require (
...
 	github.com/google/uuid v1.1.1
 	github.com/gorilla/handlers v1.4.2
+	github.com/gorilla/websocket v1.4.1 // indirect
+	github.com/grpc-ecosystem/go-grpc-middleware v1.1.0 // indirect
+	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0 // indirect
 	github.com/grpc-ecosystem/grpc-gateway v1.9.5 // indirect
 	github.com/hashicorp/hcl v1.0.0
 	github.com/hpcloud/tail v1.0.0
-	github.com/imdario/mergo v0.3.8
+	github.com/imdario/mergo v0.3.9
```
domwong, line 43: This actually causes a break in behaviour causing a failing test. Can you revert this module back to v0.3.8 please ?

domwong (changes requested):
Looks good, just need to revert one of the module updates and retest.

go.mod
vtolstov: some deps wants mergo 0.3.9, i'm fixup via replace. but may be we can have better fix

vtolstov:
@mcuadros can you lowe mergo dep in git repo and release next version?

mcuadros:
Maybe if you have a problem with a specific problem you can make a replace

vtolstov:
Do you have specific reason to bump mergo dep? What is the political decision?
Have minimal needed deps or latest released?

mcuadros:
In go-git? Usually, we update to the latest versions. Here ask to `go mod tidy`? :D

vtolstov:
ok, so we pin now prev mergo version, thanks for pr!

PR #1661, opened 2020-05-27
fix locking of socket pool

domwong:
Current implementation doesn't behave as expected. Could have

Thread 1 : call `Get()`
Thread 2: call `Get()`
Thread 1: ReadLock. Sees `p.pool[id]` is empty. ReadUnlock
Thread 2: ReadLock. Sees `p.pool[id]` is empty. ReadUnlock
Thread 1: Lock. Assigns to `p.pool[id]`. Unlock
Thread 2: Lock. Assigns to `p.pool[id]`. Unlock

Adding double checked locking removes the possibility of overwriting.

### DIFF
--- a/util/socket/pool.go
+++ b/util/socket/pool.go
@@ -19,12 +19,18 @@ func (p *Pool) Get(id string) (*Socket, bool) {
 	}
 	p.RUnlock()
 
-	// create new socket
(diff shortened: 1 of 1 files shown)


vtolstov:
looks fine

PR #1670, opened 2020-06-01
Fix config watch

sunfuze:
We run go-micro application in k8s, some time container have started to crash for can't get the right env source config. We load file sources first, then env source. 
And we haven't reproduced crash in our local dev env. It is really hard to find out what make the crash happen. 

We found the less cpu quota container have more  probability to run to crash, and i spent so many time to read the source code. 

Because every time invoke config.Load, it trigger watcher update, but the update is in another goroutine, so if the main goroutine hangged and back, and not all update operation are finished, the main goroutine can get the dirty config.

I write test code in config/default_test.go, it can easy produce the bug if you run it in a low cpu shares container.


crufter:
Thanks for this @sunfuze will review and merge tomorrow.

asim (approved):

PR #1674, opened 2020-06-02
Replace client/selector with router 

ben-toogood:
This PR refactors the selector and router, currently there is overlap with the selector looking up nodes from the registry. The router is now responsible for finding the routes to a service and the selector is responsible for selecting which route should be used when executing the request.


auth/default.go
```diff
@@ -53,7 +53,7 @@ func (n *noop) Generate(id string, opts ...GenerateOption) (*Account, error) {
 		Secret:   options.Secret,
 		Metadata: options.Metadata,
 		Scopes:   options.Scopes,
-		Issuer:   n.Options().Namespace,
+		Issuer:   n.Options().Issuer,
```
domwong, line 56: Why has this changed? It looks perfectly reasonable but just wondering how it's gone from Namespace to Issuer

api/handler/http/http.go
```diff
@@ -4,13 +4,14 @@ package http
 import (
 	"errors"
 	"fmt"
+	"math/rand"
```
domwong, line 7: Note "math/rand" is pseudo random. If you want it to be truly random you need to seed it. https://golang.org/pkg/math/rand/

auth/default.go
ben-toogood: It's to be more inline with the domain based naming, e.g. domain for registry, network for router etc. I feel like issuer is a lot more in-line with what you'd expect for an auth account.

api/handler/http/http.go
ben-toogood: Good catch, will amend.

client/grpc/grpc.go
```diff
@@ -432,19 +450,19 @@ func (g *grpcClient) Call(ctx context.Context, req client.Request, rsp interface
...
+			return err
 		}
 
 		// make the call
-		err = gcall(ctx, node, req, rsp, callOpts)
-		g.opts.Selector.Mark(service, node, err)
+		err = gcall(ctx, route, req, rsp, callOpts)
+
+		// record the result of the call to inform future routing decisions
+		g.opts.Selector.Record(route, err)
```
domwong, line 472: Why do we bother recording this? I thought the selector isn't involved with determining the route anyway.

registry/cache/cache.go
```diff
@@ -101,38 +100,44 @@ func (c *cache) quit() bool {
...
-	// otherwise delete entries
-	delete(c.cache, service)
-	delete(c.ttls, service)
+
+	if _, ok := c.services[domain]; ok {
+		delete(c.services[domain], service)
+	}
+
+	if _, ok := c.ttls[domain]; ok {
+		delete(c.ttls, service)
```
domwong: Should this be `delete(c.ttls[domain], service)` ?

registry/cache/cache.go
```diff
@@ -101,38 +100,44 @@ func (c *cache) quit() bool {
...
 
-	// check the cache first
-	services := c.cache[service]
-	// get cache ttl
-	ttl := c.ttls[service]
-	// make a copy
-	cp := util.Copy(services)
+	// lookup the values in the cache before calling the underlying registrry
+	c.RLock()
+	if srvs, ok := c.services[domain]; ok {
```
domwong, line 127: Some accesses to c.services are done under lock but others aren't (see `del()`). Needs to be consistent

vtolstov:
can you share more description why this needed? (i mean change client/selector to router)

asim:
> can you share more description why this needed? (i mean change client/selector to router)

We are consolidating the functionality that is overlapped between the selector and the router.

client/grpc/grpc.go
ben-toogood: The selector is responsible for picking one of the routes. Right now we're using the default (random) selector, but in the future we'll implement a more intelligent selector which uses recent errors / response times to determine which route is optimal.

registry/cache/cache.go
ben-toogood: Yes, good catch.

registry/cache/cache.go
ben-toogood: Actioned

registry/cache/cache.go
```diff
@@ -153,43 +158,59 @@ func (c *cache) get(service string) ([]*registry.Service, error) {
...
 
 	// get and return services
-	return get(service, cp)
+	return get(domain, service, services)
 }
 
-func (c *cache) set(service string, services []*registry.Service) {
-	c.cache[service] = services
-	c.ttls[service] = time.Now().Add(c.opts.TTL)
+func (c *cache) set(domain string, service string, srvs []*registry.Service) {
```
domwong, line 203: Is this likely to be called concurrently by different threads? If so we need some locking here.

registry/etcd/etcd.go
```diff
@@ -143,9 +154,26 @@ func (e *etcdRegistry) registerNode(s *registry.Service, node *registry.Node, op
...
+	}
+	if len(options.Domain) == 0 {
+		options.Domain = e.options.Domain
+	}
+
 	e.RLock()
-	leaseID, ok := e.leases[s.Name+node.Id]
+	// ensure the leases and registers are setup for this domain
+	if _, ok := e.leases[options.Domain]; !ok {
+		e.leases[options.Domain] = make(leases)
```
domwong, line 169: You're going to need a write lock not a read lock

registry/cache/cache.go
ben-toogood: I've made a bunch of improvements to the registry cache so this should now be resolved.

registry/etcd/etcd.go
ben-toogood: Thanks. I'll amend.

registry/mdns_registry.go
```diff
@@ -298,13 +314,24 @@ func (m *mdnsRegistry) Deregister(service *Service, opts ...DeregisterOption) er
...
 }
 
 func (m *mdnsRegistry) GetService(service string, opts ...GetOption) ([]*Service, error) {
+	// parse the options, fallback to the default domain
+	var options GetOptions
+	for _, o := range opts {
+		o(&options)
+	}
+	if len(options.Domain) == 0 {
+		m.Lock()
```
domwong, line 330: At some point we should revisit how we're using the locks. Here you're locking to read from `m.opts` but elsewhere you use the same lock to lock access to `m.services`. It's thread safe, but could be made more efficient with specific locks for whatever you're trying to lock.

selector/random.go
```diff
@@ -0,0 +1,44 @@
+package selector
```
domwong, line 1: why is this in package `selector` and not `selector/random`?

selector/random/random.go
```diff
@@ -0,0 +1,10 @@
+package random
+
+import (
+	"github.com/micro/go-micro/v2/selector"
+)
+
+// NewSelector returns a random selector
+func NewSelector(opts ...selector.Option) selector.Selector {
+	return selector.DefaultSelector
```
domwong, line 9: is this not a bit fragile? assumes `selector.DefaultSelector` is a `random` ?

ben-toogood: Unfortunately this is a standard practice in go-micro because of the DefaultX pattern. Each package has a default which has to be implemented at a package level to prevent a circular dependancy. It's something which needs to be amended globally, there is a PR @vtolstov opened which was working towards this, but it's not yet been merged.

selector/random.go
ben-toogood: I didn't realise this was possible, e.g. the encoding/json package is just named json: https://github.com/golang/go/blob/9326d71874730901c159643703a16dc34646e5bd/src/encoding/json/encode.go#L11

registry/mdns_registry.go
ben-toogood: You're correct. I need to revisit how this has been implemented. I think options we shouldn't  need a lock on as they shouldn't be mutated whilst the service is running.

selector/random.go
domwong: sorry, was unclear in my comment. Should read "why is the implementation of random in package `selector` when it should be in package `random`?" But I guess it's explained per your comment about Default* impls

go.mod
```diff
@@ -26,27 +21,23 @@ require (
...
 	github.com/hashicorp/hcl v1.0.0
 	github.com/hpcloud/tail v1.0.0
 	github.com/imdario/mergo v0.3.9
-	github.com/jonboulle/clockwork v0.1.0 // indirect
 	github.com/json-iterator/go v1.1.9 // indirect
 	github.com/kr/pretty v0.1.0
 	github.com/lib/pq v1.3.0
 	github.com/lucas-clemente/quic-go v0.14.1
 	github.com/micro/cli/v2 v2.1.2
+	github.com/micro/go-micro v1.18.0
```
sadwxqezc, line 40: Why depend on v1?

ben-toogood: gopls imported by accident, this will be removed.

router/registry/registry.go
```diff
@@ -0,0 +1,8 @@
+package registry
+
+import "github.com/micro/go-micro/v2/router"
+
+// NewRouter returns an initialised registry router
+func NewRouter(opts ...router.Option) router.Router {
```
sadwxqezc, line 6: It's the same function as `router.NewRouter`，why is it needed?

router/default.go
```diff
@@ -45,11 +45,21 @@ func newRouter(opts ...Option) Router {
...
+		networks:      map[string]bool{options.Network: true},
+		options:       options,
+		exit:          make(chan bool),
+		table:         newTable(options),
+		subscribers:   make(map[string]chan *Advert),
+		resetWatchers: make(map[string]chan bool),
 	}
+
+	// load routes from the registry and then watch them async
+	r.populateRoutes(r.options.Registry, options.Network)
```
sadwxqezc, line 59: Maybe it's better to change the name of `options.Network` to `options.Domain`?

router/service/service.go
```diff
@@ -36,9 +36,10 @@ func NewRouter(opts ...router.Option) router.Router {
 	// NOTE: might need some client opts here
 	cli := client.DefaultClient
 
-	// set options client
-	if options.Client != nil {
-		cli = options.Client
+	// get options client from the context. We set this in the context to prevent an import loop, as
+	// the client depends on the router
+	if c, ok := options.Context.Value(clientKey{}).(client.Client); ok {
```
sadwxqezc, line 41: How does this prevent an import loop?

registry/memory/memory.go
```diff
@@ -71,14 +78,16 @@ func (m *Registry) ttlPrune() {
...
-								logger.Debugf("Registry TTL expired for node %s of service %s", n.Id, name)
+			for domain, services := range m.records {
+				for service, versions := range services {
+					for version, record := range versions {
+						for id, n := range record.Nodes {
+							if n.TTL != 0 && time.Since(n.LastSeen) > n.TTL {
+								if logger.V(logger.DebugLevel, logger.DefaultLogger) {
+									logger.Debugf("Registry TTL expired for node %s of service %s", n.Id, service)
+								}
+								delete(m.records[domain][service][version].Nodes, id)
```
sadwxqezc, line 89: This part of code is deeply nested

PR #1685, opened 2020-06-04
util/wrapper: Add Static Client wrapper

ben-toogood:
A replacement for the selector/static which doesn't involve overriding the service name (this improves tracing among other things).

Example usage:
`client := pb.NewAuthService("go.micro.auth", wrapper.StaticClient("auth-srv:8010", options.Client))`

The selector/selector package is soon to be deprecated, so those who were using this previously should no longer use MICRO_SELECTOR=static, and instead wrap their clients as demonstrated above. In go-micro v2.9 and above, this will be done automatically when the MICRO_SELECTOR=static flag is detected (a depreciation warning will be logged), however this will be removed in v3.


vtolstov:
tests added

server/grpc/server.go
```diff
@@ -97,7 +97,7 @@ func prepareEndpoint(method reflect.Method) *methodType {
 		streamType := reflect.TypeOf((*server.Stream)(nil)).Elem()
 		if !argType.Implements(streamType) {
 			if logger.V(logger.ErrorLevel, logger.DefaultLogger) {
-				logger.Error(mname, "argument does not implement Streamer interface:", argType)
+				logger.Error(mname, " argument does not implement Streamer interface: ", argType)
```
domwong: Let's switch this to `logger.Errorf`

domwong: In fact can we change all of the new logger invocations in this file please

server/rpc_router.go
```diff
@@ -148,38 +148,38 @@ func prepareMethod(method reflect.Method) *methodType {
 		// check stream type
 		streamType := reflect.TypeOf((*Stream)(nil)).Elem()
 		if !argType.Implements(streamType) {
-			log.Error(mname, "argument does not implement Stream interface:", argType)
+			log.Error(mname, " argument does not implement Stream interface: ", argType)
```
domwong: Ditto these logs

domwong (approved):
Looks fine. Just change those log lines to use the formatter versions

PR #1687, opened 2020-06-05
Sigterm instead of Sigkill

crufter:
Submitting this to master as I think it qualifies as a hotfix.
No tests for it however due to time constraints.

Fixes https://github.com/micro/development/issues/199

Repro steps:
Do `micro update`s and see if `micro get service [servicename]` has more more than 1 nodes - if it does, the fix doesn't work. Also see if it actually terminates services etc.


ben-toogood (approved):

crufter:
I'm somewhat worried about this on Windows as I had a lot of pain with signals on Windows and Kill was the most reliable there.

domwong:
Works for me - tested on macos

domwong (approved):

crufter:
Only weird thing I have spotted is:

```
2020-06-05 16:09:07  file=registry/registry.go:102 level=error service=api unable to get service: service not found
```

This is when doing a micro update or even killing micro server.
But doesn't seem to have adverse effect.

crufter:
@jakexks wanted to test this on Windows but micro server is completely broken there

asim:
This change does not affect windows as windows as its own arch specific file with different method implementations https://github.com/micro/go-micro/blob/master/runtime/local/process/os/os_windows.go

ekrem95 (approved):

PR #1693, opened 2020-06-09
 Initialize selector before we make an auth.Generate call

crufter:
Web, proxy and api was failing because the `selector.Select` in `auth.Generate` still had the `mdns` registry because the selector initialization (where we change from default `mdns` to `etcd`) happened after that call.

This fixes that.

Relevant stack trace printed by debug: https://gist.github.com/crufter/c408b6355d28b0a90ae101beb5b5e7f2#file-gistfile1-txt-L33


domwong (approved):
🚀

asim (approved):

PR #1708, opened 2020-06-17
registry/mdns: add domain support

ben-toogood:
Add domain options to the registry interface and add support in mDNS.

### DIFF
--- a/registry/mdns_registry.go
+++ b/registry/mdns_registry.go
@@ -360,7 +360,13 @@ func (m *mdnsRegistry) Deregister(service *Service, opts ...DeregisterOption) er
 	}
 
 	// we have more than one entry remaining, we can exit
-	if len(newEntries) > 1 || newEntries[0].id == "*" {
+	if len(newEntries) > 1 {
+		m.domains[options.Domain][service.Name] = newEntries
+		return nil
+	}
+
+	// our remaining entry is not a wildcard, we can exit
+	if len(newEntries) == 1 && newEntries[0].id != "*" {
 		m.domains[options.Domain][service.Name] = newEntries
 		return nil
 	}
--- a/registry/mdns_registry.go
+++ b/registry/mdns_registry.go
@@ -21,8 +21,12 @@ import (
 )
 
 var (
-	// use a .micro domain rather than .local
-	mdnsDomain = "micro"
+	// use a .micro tld rather than .local by default
+	defaultTLD = "micro"
+	// every service is written to the global domain so * domain queries work, e.g.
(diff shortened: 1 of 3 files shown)


registry/mdns_registry.go
```diff
@@ -138,18 +149,19 @@ func newRegistry(opts ...Option) Registry {
 	}
 
 	// set the domain
-	domain := mdnsDomain
+	defaultTLD := defaultTLD
```
domwong: this seems wrong? Or at least confusing

registry/mdns_registry.go
```diff
@@ -258,25 +281,68 @@ func (m *mdnsRegistry) Register(service *Service, opts ...RegisterOption) error
...
+		o(&options)
+	}
+	if len(options.Domain) == 0 {
+		options.Domain = m.defaultTLD
+	}
+
+	// register in the global tld
+	if options.Domain != m.globalTLD {
+		// no errors are returned by this function so we don't need to handle the error
+		defer m.Deregister(service, append(opts, DeregisterDomain(m.globalTLD))...)
```
domwong: You're going to lose the error returned by `Deregister()` by doing the defer

registry/mdns_registry.go
```diff
@@ -293,18 +359,37 @@ func (m *mdnsRegistry) Deregister(service *Service, opts ...DeregisterOption) er
 		}
 	}
 
+	// we have more than one entry remaining, we can exit
+	if len(newEntries) > 1 || newEntries[0].id == "*" {
```
domwong: Is it possible `len(newEntries) == 0` if so then you'll get a nil pointer exception

registry/mdns_registry.go
ben-toogood: Yes, this was expected but because the function doesn't return an error, it's okay. I did add a comment for this. Would it be better to simply return m.Deregister everywhere return is called?

registry/mdns_registry.go
ben-toogood: Good catch, i'll amend.

registry/mdns_registry.go
ben-toogood: Why so?

ben-toogood: I'll rename TLD => Domain to avoid confusion.

ben-toogood:
@domwong thanks for the review, i've actioned all your feedback. Let me know if you're happy to merge.

registry/mdns_registry.go
domwong: I just looked at the function signature and saw it returned err (I missed your comment).

domwong (approved):

PR #1710, opened 2020-06-17
auth: rename Namespace to Issuer

ben-toogood:
Rename namespace to issuer within the context of auth to better fit with the domain specific language we're using across go-micro (e.g. registry=domain, router=network). 

Replaced the MICRO_AUTH_NAMESPACE env var with MICRO_NAMESPACE as this will be used in the future for a range of things beyond just auth.


domwong (approved):
Looks fine. Breaking change.

ben-toogood:
auth_namespace isn't yet being used outside micro so this should be fine.

PR #1711, opened 2020-06-17
config/cmd: remove package duplicate initialisation

ben-toogood:
Stop module re-initialisation in config/cmd. This PR ensures Init() is only called once per module if any options are passed. 

Logging has also been improved, if an implementation is not found, logger.Fatalf is now used instead of fmt.Print.

### DIFF
--- a/config/cmd/cmd.go
+++ b/config/cmd/cmd.go
@@ -2,7 +2,6 @@
 package cmd
 
 import (
-	"fmt"
 	"math/rand"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


config/cmd/cmd.go
```diff
@@ -468,10 +467,79 @@ func (c *cmd) Options() Options {
 }
 
 func (c *cmd) Before(ctx *cli.Context) error {
-	// If flags are set then use them otherwise do nothing
-	var serverOpts []server.Option
+	// Setup client options
 	var clientOpts []client.Option
 
+	if r := ctx.Int("client_retries"); r >= 0 {
```
domwong, line 473: `ctx.Int()` will return 0 in case `client_retries` is not set. Is that what we want? Not sure if there is a default client retries val

config/cmd/cmd.go
```diff
@@ -468,10 +467,79 @@ func (c *cmd) Options() Options {
...
+
+	if len(ctx.String("server_address")) > 0 {
+		serverOpts = append(serverOpts, server.Address(ctx.String("server_address")))
+	}
+
+	if len(ctx.String("server_advertise")) > 0 {
+		serverOpts = append(serverOpts, server.Advertise(ctx.String("server_advertise")))
+	}
+
+	if ttl := time.Duration(ctx.Int("register_ttl")); ttl >= 0 {
```
domwong, line 535: Similar to above `ctx.Int` returns 0 by default so register TTL is set to 0 here if not specified

config/cmd/cmd.go
```diff
@@ -468,10 +467,79 @@ func (c *cmd) Options() Options {
...
+		serverOpts = append(serverOpts, server.RegisterTTL(ttl*time.Second))
+	}
+
+	if val := time.Duration(ctx.Int("register_interval")); val >= 0 {
```
domwong, line 539: ditto

domwong (approved):
Changes look OK. Think I missed it but where were things being initialised multiple times?

config/cmd/cmd.go
ben-toogood: The flag uses client.DefaultRetries as the default value, so it would be 1 by default.

config/cmd/cmd.go
ben-toogood: The value is defaulted to 60 when the flag is declared.

config/cmd/cmd.go
ben-toogood: > Changes look OK. Think I missed it but where were things being initialised multiple times?

A few places, e.g. when we set the store / registry addresses, when we set the selector. We should either pass the options to the default implementation when we create it, or set it using Init() if we don't override the implementation.

PR #1713, opened 2020-06-18
registry/memory: add support for domain options

ben-toogood:
Add support for domain options in the memory registry.

### DIFF
--- a/registry/memory/memory.go
+++ b/registry/memory/memory.go
@@ -14,6 +14,7 @@ import (
 var (
 	sendEventTime = 10 * time.Millisecond
 	ttlPruneTime  = time.Second
+	defaultDomain = "micro"
 )
 
 type node struct {
@@ -34,19 +35,24 @@ type Registry struct {
 	options registry.Options
 
 	sync.RWMutex
-	records  map[string]map[string]*record
+	// records is a KV map with domain name as the key and a services map as the value
+	records  map[string]services
 	watchers map[string]*Watcher
 }
 
+// services is a KV map with service name as the key and a map of records as the value
+type services map[string]map[string]*record
+
+// NewRegistry returns an initialized in-memory registry
 func NewRegistry(opts ...registry.Option) registry.Registry {
 	options := registry.Options{
 		Context: context.Background(),
 	}
-
 	for _, o := range opts {
 		o(&options)
 	}
 
+	// records can be passed for testing purposes
 	records := getServiceRecords(options.Context)
 	if records == nil {
 		records = make(map[string]map[string]*record)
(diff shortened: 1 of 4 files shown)


registry/memory/memory.go
```diff
@@ -34,27 +35,32 @@ type Registry struct {
...
+	// records can be passed for testing purposes
 	records := getServiceRecords(options.Context)
 	if records == nil {
 		records = make(map[string]map[string]*record)
 	}
 
 	reg := &Registry{
 		options:  options,
-		records:  records,
+		records:  make(map[string]services),
```
domwong: Not using `records`? Then you can probably delete the block above

registry/memory/memory.go
```diff
@@ -192,93 +218,199 @@ func (m *Registry) Register(s *registry.Service, opts ...registry.RegisterOption
...
+	// if the nodes not empty, we replace the version in the store and exist, the rest of the logic
+	// is cleanup
+	if len(version.Nodes) > 0 {
+		m.records[options.Domain][s.Name][s.Version] = version
+		return nil
+	}
+
+	// if this version was the only version of the service, we can remove the whole service from the
+	// registry and exit
+	if len(versions) == 0 {
```
domwong: When would `len(versions)` be 0? We haven't deleted this version from `versions` so not sure when this is true.

registry/memory/memory.go
```diff
@@ -192,93 +218,199 @@ func (m *Registry) Register(s *registry.Service, opts ...registry.RegisterOption
...
+	for _, o := range opts {
+		o(&options)
+	}
+	if len(options.Domain) == 0 {
+		options.Domain = defaultDomain
+	}
+
+	// if it's a wildcard domain, return from all domains
+	if options.Domain == registry.WildcardDomain {
+		m.Lock()
```
domwong: Can probably be a `RLock` instead

registry/memory/memory.go
```diff
@@ -192,93 +218,199 @@ func (m *Registry) Register(s *registry.Service, opts ...registry.RegisterOption
...
+	for _, o := range opts {
+		o(&options)
+	}
+	if len(options.Domain) == 0 {
+		options.Domain = defaultDomain
+	}
+
+	// if it's a wildcard domain, list from all domains
+	if options.Domain == registry.WildcardDomain {
+		m.Lock()
```
domwong: Can probably be an RLock instead

ben-toogood:
Thanks for the review @domwong. I've actioned all your points.

domwong (approved):

PR #1714, opened 2020-06-18
registry/etcd: support domain options

ben-toogood:


registry/etcd/etcd.go
```diff
@@ -326,67 +373,122 @@ func (e *etcdRegistry) GetService(name string, opts ...registry.GetOption) ([]*r
...
+	if options.Domain == registry.WildcardDomain {
+		rsp, err := e.client.Get(ctx, prefix, clientv3.WithPrefix(), clientv3.WithSerializable())
+		if err != nil {
+			return nil, err
+		}
+
+		// filter using a check for the service name
+		keyPath := fmt.Sprintf("/%v/", serializeServiceName(name))
+		for _, kv := range rsp.Kvs {
+			if strings.Contains(string(kv.Key), keyPath) {
```
domwong, line 395: What does `kv.Key` look like? Do you have some examples? I'm always a bit worried with `strings.Contains` whether it might have a false positive. for example if `serializeServiceName(name)` returned `github.com/food/bar` and you were looking for something which was a substring there like `foo` you'd still get a hit

domwong (approved):

registry/etcd/etcd.go
ben-toogood: kv.Key is something like `micro/registry/foo/go.micro.service.helloworld/1.0.0/nodeid`. I added / to either side of the service name so it'll only match exact service names. I spent some time thinking about this and I can't find a better solution.

PR #1718, opened 2020-06-19
server: register in the services namespace

ben-toogood:
The service will now register in the domain specified using the MICRO_NAMESPACE env var.

### DIFF
--- a/config/cmd/cmd.go
+++ b/config/cmd/cmd.go
@@ -278,7 +278,7 @@ var (
 			Name:    "service_namespace",
 			EnvVars: []string{"MICRO_NAMESPACE"},
 			Usage:   "Namespace the service is operating in",
-			Value:   "go.micro",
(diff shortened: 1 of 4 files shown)


config/cmd/cmd.go
```diff
@@ -278,7 +278,7 @@ var (
 			Name:    "service_namespace",
 			EnvVars: []string{"MICRO_NAMESPACE"},
 			Usage:   "Namespace the service is operating in",
-			Value:   "go.micro",
+			Value:   "micro",
```
domwong, line 281: Changing the default. Will this have any adverse impact on people who are already running micro?

domwong (approved):

config/cmd/cmd.go
ben-toogood: No, the usage of this flag is still being implemented, right now it's not being used.

PR #1723, opened 2020-06-19
config/cmd: fix selector setup bug

ben-toogood:
Selector must be set before the auth account is generated since the selector will be used to lookup the route.


crufter (approved):

crufter:
Same bugfix as https://github.com/micro/go-micro/pull/1693
Guess +1 for running micro tests on each go-micro PR

PR #1725, opened 2020-06-19
client/selector: query across multiple domains

ben-toogood:
Right now, a service in the foo namespace can only discover services in its own namespace. This PR will allow it to call services in foo and also micro, this is important because in a multi-tenant environment the server always runs in the default namespace (micro).


client/selector/default.go
```diff
@@ -46,16 +46,22 @@ func (c *registrySelector) Select(service string, opts ...SelectOption) (Next, e
...
-	// if that fails go directly to the registry
-	services, err := c.rc.GetService(service)
-	if err != nil {
-		if err == registry.ErrNotFound {
-			return nil, ErrNotFound
-		}
+	// get the service. Because the service could be running in the current or the default domain,
+	// we call both. For example, go.micro.service.foo could be running in the services current domain,
+	// however the runtime (go.micro.runtime) will always be run in the default domain.
+	services, err := c.rc.GetService(service, registry.GetDomain(c.so.Domain))
```
domwong, line 52: Is services always a non nil slice?

domwong (approved):

client/selector/default.go
ben-toogood: I believe so, however I'll add a check. This code is being removed soon.

PR #1727, opened 2020-06-20
Add recover in grpc service handler

enix223:
Hi,

If panic occurred in the wrappers, the whole service would be crashed.

To fix this issue, recover statement has been added in the very beginning of grpc handler function.

This PR also prevents crashing the service when grpc stream service call crashed.


server/grpc/grpc_test.go
```diff
@@ -59,6 +59,12 @@ func (s *testServer) Call(ctx context.Context, req *pb.Request, rsp *pb.Response
 		return &errors.Error{Id: "1", Code: 99, Detail: "detail"}
 	}
 
+	if req.Name == "Panic" {
+		// make it panic
+		var a int
+		fmt.Print(1 / a)
```
sadwxqezc: `panic(...)` would be better

server/grpc/grpc_test.go
```diff
@@ -201,3 +207,103 @@ func TestGRPCServer(t *testing.T) {
...
+	s := gsrv.NewServer(
+		server.Broker(b),
+		server.Name("foo"),
+		server.Registry(r),
+		server.Transport(tr),
+		server.WrapHandler(func(hf server.HandlerFunc) server.HandlerFunc {
+			return func(ctx context.Context, req server.Request, rsp interface{}) error {
+				// make it panic
+				var a int
+				fmt.Print(1 / a)
```
sadwxqezc: Same as above

enix223:
@sadwxqezc Update per your request

ben-toogood (approved):

PR #1732, opened 2020-06-24
router: update interface

ben-toogood:
Update the router interface, removing the need for Start() and renaming Stop() to Close(). This brings it more inline with other go-micro interfaces. This is also needed go-micro services will soon be using the router internally (not just via the proxy).

Also, updating the default network from "go.micro" to "micro", inline with the default registry domain.

This is a breaking change, however as discussed with Asim, the router is only currently being used by micro, so I'll open a secondary PR to update this shortly.


domwong (approved):
LGTM

PR #1733, opened 2020-06-24
router: add to service options; add dns and static implementations

ben-toogood:
• Add the router to service options (so it's in place to replace the selector). 
• Add the dns and static implementations (to replace client/selector/{dns,static}. 
• Update the way client is set on the router service implementation to prevent a future import loop when the client imports router.

### DIFF
--- a/config/cmd/cmd.go
+++ b/config/cmd/cmd.go
@@ -2,6 +2,7 @@
 package cmd
 
 import (
+	"fmt"
 	"math/rand"
 	"strings"
 	"time"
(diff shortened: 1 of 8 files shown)


ben-toogood:
IMO, the dns and static router implementations are a clear indication that we need to slim down / refactor the router interface.

domwong (approved):
LGTM. Would be good to start documenting what these different types of things are useful for and when I would use each. i.e. we've defined 4 different routers there, which should I use? 

Perhaps a readme in the router directory with a quick summary would be useful.

PR #1734, opened 2020-06-24
auth/service: use address option since router may not be configured

ben-toogood:
Auth is the first thing which gets setup during Initialisation, hence the router may not be configured. To combat this, we can now manually set the auth address.

### DIFF
--- a/auth/service/service.go
+++ b/auth/service/service.go
@@ -30,10 +30,6 @@ func (s *svc) Init(opts ...auth.Option) {
 		o(&s.options)
 	}
 
-	if s.options.Client == nil {
-		s.options.Client = client.DefaultClient
-	}
-
 	s.auth = pb.NewAuthService("go.micro.auth", s.options.Client)
 	s.rules = pb.NewRulesService("go.micro.auth", s.options.Client)
 
@@ -60,7 +56,7 @@ func (s *svc) Generate(id string, opts ...auth.GenerateOption) (*auth.Account, e
 		Scopes:   options.Scopes,
 		Metadata: options.Metadata,
 		Provider: options.Provider,
-	})
+	}, s.callOpts()...)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


auth/service/service.go
```diff
@@ -30,10 +30,6 @@ func (s *svc) Init(opts ...auth.Option) {
 		o(&s.options)
 	}
 
-	if s.options.Client == nil {
```
ben-toogood, line 33: This check is already done in NewAuth()

config/cmd/cmd.go
```diff
@@ -570,6 +575,63 @@ func (c *cmd) Before(ctx *cli.Context) error {
 	microClient := wrapper.CacheClient(cacheFn, grpc.NewClient())
 	microClient = wrapper.AuthClient(authFn, microClient)
 
+	// Setup auth options
+	authOpts := []auth.Option{auth.WithClient(microClient)}
+	if len(ctx.String("auth_address")) > 0 {
+		authOpts = append(authOpts, auth.Addrs(ctx.String("auth_address")))
+	}
+	if len(ctx.String("auth_id")) > 0 || len(ctx.String("auth_secret")) > 0 {
```
domwong, line 583: Should this be an `&&` rather than an `||`?

config/cmd/cmd.go
```diff
@@ -570,6 +575,63 @@ func (c *cmd) Before(ctx *cli.Context) error {
...
+		a, ok := c.opts.Auths[name]
+		if !ok {
+			logger.Fatalf("Unsupported auth: %s", name)
+		}
+		*c.opts.Auth = a(authOpts...)
+		serverOpts = append(serverOpts, server.Auth(*c.opts.Auth))
+	} else if len(authOpts) > 0 {
+		(*c.opts.Auth).Init(authOpts...)
+	}
+
```
domwong, line 634: Why did you move this block of code up the order?

config/cmd/cmd.go
ben-toogood: No, if either value is set we want to set the option. If it's incorrect the auth service will be responsible for returning an error on startup. We only really have the check so we don't call Init without reason.

config/cmd/cmd.go
ben-toogood: So that auth is configured before calls to the router / registry get made since these could be authorised.

domwong (approved):

PR #1739, opened 2020-06-24
cached file store

domwong:
A memory cached version of the file store to reduce DB load for reads.

Read through and write through.


asim:
Could store/cache be used for whatever purpose this is for? And then initialise with the memory and file stores. Its a more generic caching abstraction.

domwong:
> Could store/cache be used for whatever purpose this is for? And then initialise with the memory and file stores. Its a more generic caching abstraction.

I don't think so (happy to be proven wrong). I want something that will be a read through cache in front of a file store so for a cache hit on read we never hit the file store and writes will write through to file.  The implementation of `Read` on `cache` store seems to cause a number of writes for every read, keeping all the stores in sync so it would be seem to be more IO heavy rather than less.

asim:
I think it would be good modify store/cache inline with your expectations. The implementation for store/cache was done some time ago but doesn't necessarily need to to stick. Basically I think in the way we have registry/cache, if we're doing any sort of store caching there should just be one way. So I'd replace store/cache with your implementation. Then maybe we can evolve as we start to think beyond read cache. Nobody is using the current store/cache so we'll be able to refactor without issue.

domwong:
@jakexks you were the original author of `cache`. Was there a specific use case in mind here? I can't see it being used anywhere; it isn't one of the default stores in `micro`.

asim:
It was done based on an idea for global storage when we were running multi cloud, multi region platform in November 2019 - April 2020. That's no longer the case for now and we never got around to using it so I'd say now is the time to replace with a simple read cache and then add options for read/write settings later.

asim (approved):

PR #1742, opened 2020-06-25
server/{grpc,rpc}: fix deregister domain bug

ben-toogood:
The domain option was being passed when a service registered, but not when it deregistered.


server/grpc/grpc.go
```diff
@@ -794,7 +794,9 @@ func (g *grpcServer) Deregister() error {
 	if logger.V(logger.InfoLevel, logger.DefaultLogger) {
 		logger.Infof("Deregistering node: %s", node.Id)
 	}
-	if err := config.Registry.Deregister(service); err != nil {
+
+	opts := []registry.DeregisterOption{registry.DeregisterDomain(g.opts.Namespace)}
```
domwong: probably don't need to construct the slice here only to deconstruct it in the next line

domwong (approved):

PR #1745, opened 2020-06-25
router: improve router configuration

ben-toogood:


ben-toogood:
I think the build issue with micro is because we also have the "router_address" cli var declared there, however I can't remove that until this PR is merged.

domwong (approved):

PR #1746, opened 2020-06-25
proxy/mucp: add support for multi-tenancy

ben-toogood:


proxy/mucp/mucp.go
```diff
@@ -129,7 +129,7 @@ func (p *Proxy) filterRoutes(ctx context.Context, routes []router.Route) []route
 
 		// only process routes with this network
 		if net, ok := md.Get("Micro-Network"); ok && len(net) > 0 {
-			if route.Network != net {
+			if route.Network != router.DefaultNetwork && route.Network != net {
```
ben-toogood, line 132: this line ensures all requests can be routed to either the requested network (e.g. "foo") or the default ("micro")

domwong (approved):

PR #1747, opened 2020-06-26
api/resolver: update resolver to enable subdomain routing

ben-toogood:
Add the router which resolves requests made against "foo.m3o.app" to services in the "foo" domain.

### DIFF
--- a/api/resolver/subdomain/subdomain.go
+++ b/api/resolver/subdomain/subdomain.go
@@ -1,0 +1,86 @@
+// Package subdomain is a resolver which uses the subdomain to determine the domain to route to. It
+// offloads the endpoint resolution to a child resolver which is provided in New.
+package subdomain
+
+import (
+	"net"
+	"net/http"
+	"strings"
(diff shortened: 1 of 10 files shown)


ben-toogood:
Tests are failing because an update is required in micro. This is a common requirement as lots of micro specific logic (e.g. the api), still lives in go-micro. It would be good if we could find a way to link go-micro and micro PRs @domwong ?

api/resolver/subdomain/subdomain.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+	}
+
+	// check for an ip address
+	if net.ParseIP(host) != nil {
+		return r.opts.Domain
+	}
+
+	// check for dev enviroment
+	if host == "localhost" || host == "127.0.0.1" {
```
domwong, line 56: Do we need to check for `0.0.0.0`?

domwong (approved):

Issue #1750, opened 2020-06-27
[BUG] Failed to call Response.Write() in outbound


wudi:
**Describe the bug**

1. What are you trying to do?

Try to implement UDP transport

2. What did you expect to happen?

Exit the coroutine safely.

3. What happens instead?

Write failed

**Description**

The following code

https://github.com/micro/go-micro/blob/5967a68e7830f39672a415b78eb748f6cee1ea87/server/rpc_server.go#L149-L169

This defer func is not cover exceptions while the defer invoked faster than process of outbound and inbound.
Because of the psock in pool is closed when write data from coroutine of outbound.

```
Loop 1

Transport---Recv---pkg1---go inbound
                       ---go outbound----------T1---------Response.Write() // socket closed by loop2

Loop 2

Transport---Recv---Error---Close socket (not wait)
```

wudi:
Issues

1. Invoke defer func faster than execute two coroutines.  PR #1752
2. Not wait while occur error on Recv

wudi:
Here the code  to reproduce the issue 1

```go
package main

import (
        "fmt"
        "sync"
        "time"
)

func main() {
        foo()
}

func foo() {
        var wg sync.WaitGroup

        defer func() {
                fmt.Printf("defer: %s\n", time.Now().Format(time.StampMicro))
                wg.Wait()
        }()

        time.Sleep(5 * time.Millisecond)

        go func() {
                wg.Add(1)
                fmt.Printf("co1: %s\n", time.Now().Format(time.StampMicro))
                wg.Done()
        }()

        go func() {
                wg.Add(1)
                fmt.Printf("co2: %s\n", time.Now().Format(time.StampMicro))
                wg.Done()
        }()
}

```

`go build -o test main.go`


Try several times `./test`, u will see the defer invoked faster than other coroutines.

vtolstov:
Thanks! What about number 2?

wudi:
@vtolstov  tks for reply

Two coroutines to wait for, one is the serve http the other is process of outbound. Any close early will occur the same issue.
I've been chewing things over, how about always wait? 
Looks safety, even if the connection is broken, two coroutines always exit normally.

wudi:
All fix merged, tks guys

PR #1752, opened 2020-06-27
Fix invalid usage for sync.WaitGroup

wudi:
Issues: https://github.com/micro/go-micro/issues/1750


vtolstov:
@asim i think that this changes is right.

asim (approved):

vtolstov:
@asim as i see we have new test case, but it not update status

PR #1758, opened 2020-06-29
router/registry: add fallback if routes aren't found in the cache

ben-toogood:
This PR fixes the flakey tests in micro.

### DIFF
--- a/router/table.go
+++ b/router/table.go
@@ -19,8 +19,8 @@ var (
 // table is an in-memory routing table
 type table struct {
 	sync.RWMutex
-	// router which manages the table
-	router *router
+	// fetchRoutes for a service
+	fetchRoutes func(string) error
 	// routes stores service routes
 	routes map[string]map[uint64]Route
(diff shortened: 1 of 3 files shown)


router/table.go
```diff
@@ -19,15 +19,18 @@ var (
 // table is an in-memory routing table
 type table struct {
 	sync.RWMutex
+	// router which manages the table
+	router *router
```
domwong: Not a fan of this circular dependency. Any way we can break this ?

ben-toogood: I can either:
• Pass the fetchRoutes function as an argument when creating the table, it's still a kind of circular dependancy but less direct.
• Merge the table struct and the router struct. Right now there the table is "dumb" and has no knowledge of the registry or where any of its data comes from.

domwong: Let's pass in `fetchRoutes`. Thanks

domwong (approved):

PR #1759, opened 2020-06-29
MDNS registry fix for users on VPNs

domwong:
or other software (firewalls etc) that may block MDNS responses. 

In some cases, if a user was using a VPN, the MDNS server would try to respond to the query using the public IP address of the machine i.e. as assigned by their VPN. In most cases the VPN server would block the response which meant that the client never received the packets and thus MDNS registry would never work. This means running `micro server` would fail.

The proposed fix is to detect if the IP address that we are responding to is the local machine and, if so, attempt to also send our responses to `0.0.0.0` thus avoiding the VPN. This is done on a best efforts basis with no error checking. We detect the IP at start up using a simple ping to Google's DNS server (perhaps we should pick a different address since this may not be available in all locations - e.g. China).

### DIFF
--- a/util/mdns/server.go
+++ b/util/mdns/server.go
@@ -235,9 +235,9 @@ func (s *Server) handleQuery(query *dns.Msg, from net.Addr) error {
 		// 18.1: ID (Query Identifier)
 		// 0 for multicast response, query.Id for unicast response
 		id := uint16(0)
-		// if unicast {
-		id = query.Id
-		// }
+		if unicast {
+			id = query.Id
+		}
 
 		var answer []dns.RR
 		if unicast {
@@ -278,8 +278,8 @@ func (s *Server) handleQuery(query *dns.Msg, from net.Addr) error {
 			// caveats in the RFC), so set the Compress bit (part of the dns library
 			// API, not part of the DNS packet) to true.
 			Compress: true,
-
-			Answer: answer,
+			Question: query.Question,
+			Answer:   answer,
 		}
 	}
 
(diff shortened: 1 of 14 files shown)


sadwxqezc:
Google DNS is available in China

util/mdns/server.go
```diff
@@ -430,13 +432,23 @@ func (s *Server) sendResponse(resp *dns.Msg, from net.Addr) error {
...
+	_, err = conn.WriteToUDP(buf, addr)
+	// If the address we're responding to is this machine then we can also attempt sending on 0.0.0.0
+	// This covers the case where this machine is using a VPN and certain ports are blocked so the response never gets there
+	// Sending two responses is OK
+	if addr.IP.Equal(s.outboundIP) {
+		// ignore any errors, this is best efforts
+		conn.WriteToUDP(buf, &net.UDPAddr{IP: backupTarget, Port: addr.Port})
 	}
+	return err
+
```
sadwxqezc, line 466: I think it's better to make sending on `0.0.0.0` an option

util/mdns/server.go
```diff
@@ -474,3 +486,17 @@ func setCustomPort(port int) {
...
+	if err != nil {
+		// no net connectivity maybe so fallback
+		return nil
+	}
+	defer conn.Close()
+
+	localAddr := conn.LocalAddr().(*net.UDPAddr)
+
+	return localAddr.IP
+}
```
sadwxqezc, line 517: Better to make this an interface

domwong:
> Google DNS is available in China

Thanks for confirming. 

Have taken onboard your other suggestions. Thanks for the input!

crufter (approved):

PR #1761, opened 2020-06-30
selector: new selector interface with random & roundrobin implementation

ben-toogood:
This is the new selector which will be used in combination with the router. Once this has been merged, i'll be removing the client/selector and updating the clients to use go-micro/{selector,router}.


ben-toogood:
@domwong, I looked into using the ring buffer but it didn't work very well for this use-case and I foresee other cleanup once we start using the Record() method, so i've added a Close method to the interface.

crufter (approved):

selector/selector.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+// Selector selects a route from a pool
+type Selector interface {
+	// Init a selector with options
+	Init(...Option) error
+	// Options the selector is using
+	Options() Options
+	// Select a route from the pool using the strategy
+	Select(...*router.Route) (*router.Route, error)
```
domwong: Out of interest why is this variadic rather than a slice?

domwong: Apart from that, interface looks good

ben-toogood: No real reason. It might be worth changing this actually as we could add Select options in the future.

domwong: Yeah maybe change to slice then. I *think* most uses of this will be with an existing slice rather than cobbling together multiple items so it won't involve too many extra slice creations

selector/selector.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+// Selector selects a route from a pool
+type Selector interface {
+	// Init a selector with options
+	Init(...Option) error
+	// Options the selector is using
+	Options() Options
+	// Select a route from the pool using the strategy
+	Select([]*router.Route) (*router.Route, error)
```
asim, line 24: It is simpler to define a "Route" or "Node" here so that router is not imported?

ben-toogood: I'd rather define an interface which is satisfied by the router.Route, however I think this is a wider issuer which we need to look at outside this PR.

PR #1767, opened 2020-06-30
Deprecate client/selector

ben-toogood:
• client: remove Registry from options
• client/{rpc,grpc}: now uses the router to find routes
• client/{rpc,grpc}: now uses the selector to choose a route
• client/{rpc,grpc}: record call results in the selector
• api/handler: replace SelectOptions with a custom router. 
• client: update CallWrapper to use router.Route and not registry.Node
• util/http: Update roundtripper to use router and selector, not registry


ben-toogood:
Note: the client CallWrapper had to change, since the clients are now using routes and not nodes. This is a breaking change for anyone wrapping their clients manually. This is the reason behind one of the examples failing to build.

domwong (approved):
LGTM. Lots of changes so I'd anticipate some bugs but let's get it merged

PR #1779, opened 2020-07-02
Sparse checkout of repos, all repo downloads happen to different folders for concurrency support

crufter:
The compression methods should likely go somewhere else than the git package.

### DIFF
--- a/runtime/local/git/git.go
+++ b/runtime/local/git/git.go
@@ -1,157 +1,70 @@
 package git
 
 import (
+	"archive/tar"
+	"compress/gzip"
 	"errors"
 	"fmt"
+	"io"
+	"net/http"
 	"os"
 	"os/exec"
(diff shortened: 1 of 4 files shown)


runtime/local/git/git.go
```diff
@@ -1,157 +1,79 @@
...
-	}
-	cmd.Dir = fold
-	_, err = cmd.Output()
-	if err != nil {
-		return err
-	}
-	return err
-}
+	defer resp.Body.Close()
+	// Github returns 404 for tar.gz files...
```
domwong, line 47: wat

crufter: I can actualy remove this now as zip files are not throwing 404s....

asim:
+1 for sparse checkout

domwong (approved):

PR #1783, opened 2020-07-02
options: add router option

ben-toogood:
Enable router to be set pragmatically when initialising a service.


options.go
```diff
@@ -195,6 +195,15 @@ func Runtime(r runtime.Runtime) Option {
 	}
 }
 
+// Router sets the router
+func Router(r router.Router) Option {
+	return func(o *Options) {
+		o.Router = r
+		// Update client
+		o.Client.Init(client.Router(r))
```
domwong, line 203: Do we need to do nil checking here or is `o.Client` always non nil?

ben-toogood: It's always non-nil

domwong (approved):

PR #1788, opened 2020-07-02
 router/registry: fix rlock bug when registry errors

ben-toogood:


jakexks (approved):

router/table.go
```diff
@@ -256,27 +253,34 @@ func (t *table) Query(q ...QueryOption) ([]Route, error) {
 
 	if opts.Service != "*" {
 		// try and load services from the cache
-		if routes, ok := t.routes[opts.Service]; ok && len(routes) > 0 {
+		t.RLock()
+		routes, ok := t.routes[opts.Service]
+		t.RUnlock()
+		if ok && len(routes) > 0 {
 			return findRoutes(routes, opts.Address, opts.Gateway, opts.Network, opts.Router, opts.Strategy), nil
```
domwong: I think `findRoutes` needs to be done under a read lock since it reads from the map `routes` which might also be accessed from another thread

ben-toogood: Okay, i'll amend now 👍

domwong (approved):
Added a test to make sure it doesn't happen again

PR #1791, opened 2020-07-03
fix register bug

sadwxqezc:
There is a bug in mdns register, the copied `srv` is not used to register into globalDomain


domwong (approved):
🚀  thanks

PR #1800, opened 2020-07-06
auth: pass namespace options in auth service requests

ben-toogood:
Prerequisite for authorisation in micro/auth.


auth/service/proto/auth.proto
```diff
@@ -87,6 +92,7 @@ message TokenRequest {
 	string secret = 2;
 	string refresh_token = 3;
 	int64 token_expiry = 4;
+	Options options = 8;
```
domwong: 8?

ben-toogood: Typo - I'll amend

domwong (approved):

crufter (approved):

PR #1801, opened 2020-07-07
Fix invalid register check err in log

wudi:
should log the variable `rerr` for register check rather than `err`


vtolstov (approved):

vtolstov:
@asim its trivial change and safe for merge

PR #1803, opened 2020-07-07
config/cmd: custom certificate authorities & secure registry

ben-toogood:


config/cmd/cmd.go
```diff
@@ -650,6 +675,18 @@ func (c *cmd) Before(ctx *cli.Context) error {
 
 	// Setup registry options
 	registryOpts := []registry.Option{registrySrv.WithClient(microClient)}
+
+	// Parse registry TLS certs
+	if ctx.Bool("registry_secure") {
+		cert, err := tls.LoadX509KeyPair("/certs/registry/cert.pem", "/certs/registry/key.pem")
```
domwong, line 681: In a future version I wonder whether we want this configurable too.

domwong (approved):

config/cmd/cmd.go
```diff
@@ -110,6 +113,11 @@ var (
 	DefaultCmd = newCmd()
 
 	DefaultFlags = []cli.Flag{
+		&cli.StringFlag{
+			Name:    "certificate_authorities",
+			EnvVars: []string{"MICRO_CERTIFICATE_AUTHORITIES"},
```
asim, line 118: MICRO_CLIENT_CA or MICRO_AUTH_CA might be better

config/cmd/cmd.go
asim: I think these should probably be configurable or loaded from a relative path.

config/cmd/cmd.go
ben-toogood: I know what you're saying however since the CA isn't just being used by the client, I think it's misleading. I'll merge as-is but we should come back and review all the changes once mTLS is done. I just didn't want to end up merging it as one huge PR akin to client-router.

asim: No one wants to set a value called MICRO_CERTIFICATE_AUTHORITIES for one thing, too long, another thing easy to misspell. From a complete usability standpoint its unusable so we should change it now to at the very least be MICRO_CA

asim: considering its probably used as either the RootCA or ClientCA field in tls certs then either MICRO_ROOT_CA or MICRO_CLIENT_CA might actually be more appropriate, they also use the CA acronym.

PR #1804, opened 2020-07-08
server.Init(Wait(nil)) update server options context

uxgnaw:
default grpc server wg is nil, server.Init can't update it，we nedd update it


sadwxqezc:
LGTM

asim (approved):

PR #1805, opened 2020-07-08
Cache interface

asim:
The cache interface is the first version of an interface specifically designed for caching. In future implementations may include memcache, redis, groupcache, bigcache and other things deemed for caching. We may run a cache service on the platform specifically for this purpose.

Our store interface is a persistent storage layer but it is designed for the backend of a database as opposed to fast caching. What it highlights is that every interface dictates a certain abstraction and design pattern. Cache demonstrates this as a next iteration. In the same way Model will likely be for SQL and Graph will be for graph querying. This is not exclusive to data storage. Micro being a distributed systems framework dictates abstractions / design patterns as interfaces.

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -1,0 +1,17 @@
+// Package cache is a caching interface
+package cache
+
+type Cache interface {
+	// Initialise options
+	Init(...Option) error
+	// Get a value
+	Get(key string) (interface{}, error)
+	// Set a value
+	Set(key string, val interface{}) error
+	// Delete a value
+	Delete(key string) error
+	// Name of the implementation
+	String() string
+}
+
+type Option struct{}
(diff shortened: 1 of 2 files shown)


cache/cache.go
```diff
@@ -0,0 +1,17 @@
...
+	// Initialise options
+	Init(...Option) error
+	// Get a value
+	Get(key string) (interface{}, error)
+	// Set a value
+	Set(key string, val interface{}) error
+	// Delete a value
+	Delete(key string) error
+	// Name of the implementation
+	String() string
```
sadwxqezc, line 14: I guess mget/mset/mdelete is needed

asim: Not right now. Simple is better than verbose.

sadwxqezc: Make sense

PR #1806, opened 2020-07-08
config/cmd: secure broker

ben-toogood:


config/cmd/cmd.go
```diff
@@ -673,14 +678,26 @@ func (c *cmd) Before(ctx *cli.Context) error {
 		brokerOpts = append(brokerOpts, broker.Addrs(ctx.String("broker_address")))
 	}
 
+	// Parse broker TLS certs
+	if ctx.Bool("broker_secure") {
+		// cert, err := tls.LoadX509KeyPair("/certs/broker/cert.pem", "/certs/broker/key.pem")
+		cert, err := tls.LoadX509KeyPair("./platform/kubernetes/resource/nats/certs/client.pem", "./platform/kubernetes/resource/nats/certs/client-key.pem")
```
asim: This is go-micro. "platform" should not be references here, and neither should be kubernetes if possible". Ideally you could do ./micro/broker/certs/ unless there's some limitation with dirs based on k8s.

ben-toogood: My bad, I forgot to remove the replace, this was only meant to be temp.

asim: Its still an absolute path of /certs. This is going to not really work for a lot of people because to write to / you need root privileges, hence a local directory is always better. So ./micro/broker/certs

asim (approved):

PR #1807, opened 2020-07-08
feat: refactor register func

sadwxqezc:
Make the register func cleaner


asim:
why?

sadwxqezc:
I felt the former code a bit too long when reading

PR #1810, opened 2020-07-09
close r/w

sadwxqezc:
Close writer&reader

### DIFF
--- a/registry/mdns_registry.go
+++ b/registry/mdns_registry.go
@@ -85,6 +85,7 @@ func encode(txt *mdnsTxt) ([]string, error) {
 
 	w := zlib.NewWriter(&buf)
 	if _, err := w.Write(b); err != nil {
+		w.Close()
 		return nil, err
 	}
 	w.Close()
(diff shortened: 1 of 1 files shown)


asim (approved):

vtolstov:
@asim @sadwxqezc if you interesting in mdns registry i think that https://github.com/micro/go-micro/pull/1810/files#diff-2da2b18e238cfc41d240e444a3c1d651R86 needs 
```
defer func () {
  if err := w.Close(); err != nil {
    logger.Errorf("[mdns] registry error: %v", err)
  }
}()

sadwxqezc:
> @asim @sadwxqezc if you interesting in mdns registry i think that https://github.com/micro/go-micro/pull/1810/files#diff-2da2b18e238cfc41d240e444a3c1d651R86 needs
> 
> ```
> defer func () {
>   if err := w.Close(); err != nil {
>     logger.Errorf("[mdns] registry error: %v", err)
>   }
> }()
> ```

Actually you can't use `defer` here，as this part of code needs to call `close` to trigger `flush`。However, logging the error would be perfect

vtolstov:
this is not problem, you can add flush before close

sadwxqezc:
I will open another pr

PR #1812, opened 2020-07-09
kubernetes: fixes for production

ben-toogood:


crufter (approved):

util/kubernetes/client/client.go
```diff
@@ -23,7 +23,7 @@ var (
 	// ErrReadNamespace is returned when the names could not be read from service account
 	ErrReadNamespace = errors.New("Could not read namespace from service account secret")
 	// DefaultImage is default micro image
-	DefaultImage = "micro/go-micro"
+	DefaultImage = "micro/cells:go"
```
asim: We should leave this as micro/go-micro and either pass the appropriate commands or specify micro/cells:go when we actually do `micro run` and pass image to the runtime. I don't think go-micro should reference this image.

crufter: Overwriting it in `micro run` should be simple enough assuming it's correctly passed down the chain.

runtime/kubernetes/service.go
```diff
@@ -66,6 +66,9 @@ func newService(s *runtime.Service, c runtime.CreateOptions) *service {
 			kdeploy.Spec.Template.PodSpec.Containers[i].Command = []string{}
 			kdeploy.Spec.Template.PodSpec.Containers[i].Args = []string{}
 		}
+	} else {
+		// use the default image (micro/cell:go), passing the source as the first argument
+		kdeploy.Spec.Template.PodSpec.Containers[0].Args = []string{s.Source}
```
asim, line 71: is this going to cause problems? The idea was to remove any clever logic from here and keep it in the CLI since that has knowledge of any requirements.

PR #1821, opened 2020-07-12
V3: Remove config/cmd and top-level interfaces and initialisers

asim:
In preparation for V3 we look to remove config/cmd and the top level service definition.

Anyone looking to use micro.NewService should now refer to the go-micro/service package or to micro/micro/service which will include the flag/env var parsing as we did here. The idea is now to simplify go-micro and turn it into a standard library for distributed systems, where each package can be used independently if needed.

Micro V3 becomes the framework.

This should only be merged after config/cmd is added to micro/micro/cmd and the top level go-micro/Service is moved to micro/micro/service.


vtolstov:
Do you have some migration info?

asim:
Not yet

asim:
Closing, taking another shot at it.

PR #1831, opened 2020-07-14
Let bolt do locking per each List Get etc op, instead of managing fil…

crufter:
I have spotted a problem with `micro server` where the `auth` service and the `store` service both try to touch the `auth` db - the former when doing day to day ops the latter when I issue a `micro store list --table auth`.

[Unfortunately only 1 process can hold a lock to a bolt file](https://github.com/etcd-io/bbolt#opening-a-database) so the `micro store list --table auth` command was timing out. This PR makes it so the db lock is only acquired while the `List` `Get` etc operations are happening.

### DIFF
--- a/store/file/file.go
+++ b/store/file/file.go
@@ -7,7 +7,6 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
-	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


domwong (approved):
LGTM

PR #1861, opened 2020-07-20
runtime: support for dynamic secrets

ben-toogood:
When creating a runtime service, you can now use the WithSecret("key", "value") option and it'll inject the secrets into the service. In the local runtime, they'll be injected directly as env vars. In the k8s runtime, they'll be created as k8s secrets which are then mounted to the deployment as env vars.


ben-toogood:
This PR is a draft as i'm yet to test it in staging, i'll do that shortly and then mark it as ready for review.

asim:
This change is very good. I welcome it. LGTM.

domwong:
How does this work for secrets with the same name? e.g. If I want to create a single secret and use it across multiple services how does that work? Will the creation of the secret fail for the second service and then cause it to fail creation of the service? 

What happens when you delete the service and recreate? Will it fail on secret creation?

ben-toogood:
@domwong due to the current runtime interface is built around purely running services (whereas k8s supports CRUD for namespaces, secrets etc), I didn't add support for shared secrets. When the service is deleted, it'll delete the secret for that service also. 

To your point regarding two secrets with the same name, since secrets are scoped to each service, it won't error if you use the same key for another service.

PR #1873, opened 2020-07-27
Added 2 fuzzers

AdamKorcz:
This PR adds 2 fuzzers to go-micro

The fuzzers send thousands of pseudo-random inputs to their target functions each second in order to expose any bugs.

The fuzzers can be run locally, but I suggest setting up continuous fuzzing. This will allow the fuzzers to run continuously without anymore code added to go-micro. I can confirm that the fuzzers run on OSS-fuzz's platform, and as I have the build scripts ready, it will be my pleasure to integrate go-micro.
OSS-fuzz is free, but is offered with an implied expectation that bugs are fixed, so that the resources spent on fuzzing go-micro are put to good use. All I need to setup continuous fuzzing are the email addresses for the bug reports. These will be public and can be changed anytime.


vtolstov:
thanks for pr, can you add some docs how to use fuzz ? Also i'm prefer to have go-fuzz integration and not depend on external services

vtolstov:
@asim what you say?

asim:
I want to understand the usage first. All I'm seeing is some functions added to some packages.

vtolstov:
> I want to understand the usage first. All I'm seeing is some functions added to some packages.

mostly go-fuzz needs some function with specific signature that accepts []byte slice. So you can test functions with garbage in input. This pr adds func to api handler

asim:
I think this will easily end up with Fuzz methods everywhere so I think realistically there needs to be a better approach to this and we'd rather handle our own fuzzing.

PR #1879, opened 2020-07-28
[WIP] Runtime now uses `server.Download` instead of assuming file has been uploaded to the same host as runtime

crufter:


asim:
We're going to move to new endpoints in micro so no longer needed

PR #1885, opened 2020-07-30
client: add proxy option

ben-toogood:


vtolstov:
why adding new arg to LookupRoute ? Why not create CallOption for proxy address? I think that it more idiomatic

asim:
Set as call option as @vtolstov  mentioned. Basically set as the address in call opts.

client/options.go
```diff
@@ -41,6 +41,9 @@ type Options struct {
 	// Other options for implementations of the interface
 	// can be stored in a context
 	Context context.Context
+
+	// ProxyAddress to send requests via
+	ProxyAddress string
```
asim: Just cal Proxy and comment // Address to proxy all requests through

vtolstov: why not say Address and comment about that this is endpoint address or some proxy?

ben-toogood:
@asim / @vtolstov I've actioned feedback.

vtolstov:
LGTM, waiting @asim response

asim (approved):

asim:
Looks good, merging. Happy to get rid of the env vars.

PR #1893, opened 2020-08-04
router/registry: fix initialisation bug

ben-toogood:
Fix a bug where calling Init on the router with a new registry didn't take effect.


domwong (approved):

router/registry/registry.go
```diff
@@ -62,17 +64,17 @@ func NewRouter(opts ...router.Option) router.Router {
...
-	// stop the router before we initialize
-	if err := r.Close(); err != nil {
-		return err
-	}
-
 	r.Lock()
-	defer r.Unlock()
-
 	for _, o := range opts {
 		o(&r.options)
```
domwong, line 69: Should we reset options on Init so that we don't have any hangover from previous inits?

PR #1903, opened 2020-08-06
codec - Allow to Write() nil body

### LINKED ISSUE #1902: [BUG] proto codec is not able to write errors
**Describe the bug**

1. What are you trying to do?

I am trying to return an error from my RPC service with a protobuf protocol (codec `proto`)

2. What did you expect to happen?

Error is encoded and written properly

3. What happens instead?

A hidden (see #1900) log message is written (i.e. nothing happens)

**How to reproduce the bug:**

Using v2.9.1, just return an error in from the protobuf handler
```golang

func MyHandler(ctx context.Context, req *proto.Request, rsp *proto.Response) error {
	return errors.New("this is an error")
}
```

I debug the code a little:
The hidden error:
https://github.com/micro/go-micro/blob/94bd1025a64eaaba1d89189891ef49f3ab5eee7f/server/rpc_server.go#L437

It is cause because
https://github.com/micro/go-micro/blob/94bd1025a64eaaba1d89189891ef49f3ab5eee7f/server/rpc_server.go#L425
fills in `nil` for `b` (second function parameter)

This ends up (via rpc_codec.go) in (so `b = nil`):
https://github.com/micro/go-micro/blob/94bd1025a64eaaba1d89189891ef49f3ab5eee7f/codec/proto/proto.go#L36

Which returns the (hidden) error.

When looking at JSON codec
https://github.com/micro/go-micro/blob/94bd1025a64eaaba1d89189891ef49f3ab5eee7f/codec/json/json.go#L34
it just returns (`nil`) in this case, preventing the error from occurring.

veger:
I added `nil` checks to the codecs that did not have them, and added a test to check that all codecs accept a `nil` body.

Note that I am using v2.9.1 this PR is targets the `release-2.9` branch

closes #1902

### DIFF
--- a/codec/bytes/bytes.go
+++ b/codec/bytes/bytes.go
@@ -44,6 +44,8 @@ func (c *Codec) ReadBody(b interface{}) error {
 func (c *Codec) Write(m *codec.Message, b interface{}) error {
 	var v []byte
 	switch vb := b.(type) {
+	case nil:
+		return nil
 	case *Frame:
 		v = vb.Data
 	case *[]byte:
--- a/codec/codec_test.go
+++ b/codec/codec_test.go
@@ -1,0 +1,54 @@
+package codec_test
+
+import (
+	"io"
+	"testing"
+
+	"github.com/micro/go-micro/v2/codec"
+	"github.com/micro/go-micro/v2/codec/bytes"
+	"github.com/micro/go-micro/v2/codec/grpc"
+	"github.com/micro/go-micro/v2/codec/json"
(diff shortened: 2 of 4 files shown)


asim:
Merged by mistake. I don't know how we're tracking 2.9 updates yet

veger:
Should I make a new PR?

veger:
BTW, the issue is also in `master`, so after merging to 2.9.1 or master, a cherry-pick to the other branch is required I guess?

asim:
Please PR to master and then we'll think about the 2.x after

PR #1913, opened 2020-08-08
fix g.opts.CallOptions can't change

xiaobudongzhang:
## Pull Request template
Please, go through these steps before clicking submit on this PR.

1. Give a descriptive title to your PR.
where i use github.com/micro/go-micro/v2/client/grpc.DialOptions want to change grpc.DialOption , then i use github.com/micro/go-micro/v2/client/grpc.getGrpcDialOptions  to get grpc.DialOption, i  cant found my change
2. Provide a description of your changes.
	
        callOpts := g.opts.CallOptions
	for _, opt := range opts {
		opt(&callOpts )
	}
      callOpts  cant change g.opts.CallOptions.Conent, so i change it to opt(&g.opts.CallOptions)
3. Make sure you have some relevant tests.

i test in my local 
4. Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if applicable).

**PLEASE REMOVE THIS TEMPLATE BEFORE SUBMITTING**

### DIFF
--- a/client/grpc/grpc.go
+++ b/client/grpc/grpc.go
@@ -357,7 +357,7 @@ func (g *grpcClient) Call(ctx context.Context, req client.Request, rsp interface
 	// make a copy of call opts
 	callOpts := g.opts.CallOptions
 	for _, opt := range opts {
-		opt(&callOpts)
+		opt(&g.opts.CallOptions)
 	}
 
 	// check if we already have a deadline


asim:
Please move the call options access below this. It returns a copy which means if its modified its unclear if we get the changes.

xiaobudongzhang:
> Please move the call options access below this. It returns a copy which means if its modified its unclear if we get the changes.

ok, i done that.

asim:
OK, the problem with your PR is that it modifies that actual call options of the client not for this specific call. Thats not what its intended for. This is why we make a copy.

PR #1929, opened 2020-08-13
Metrics interface and Prometheus implementation

chrusty:
Here is a proposal for a metrics interface, with a sample Prometheus implementation.

Capabilities:
- Go runtime metrics are collected and exposed
- Custom user-defined metrics are supported
- Metrics are made available to scrape by Prometheus (or any HTTP client)
- Config parameters handled with an `options...` type interface
- Includes a server handler wrapper to instrument requests, but this should be moved to another repo I believe?
- Good test coverage, though I would like to actually test hitting the prometheus metrics endpoint and looking at the metrics provided there.


metrics/README.md
```diff
@@ -0,0 +1,22 @@
...
+=======
+
+The metrics package provides a simple metrics "Reporter" interface which allows the user to submit counters, gauges and timings (along with key/value tags).
+
+Implementations
+---------------
+
+* Prometheus (pull): will be first
+* Prometheus (push): certainly achievable
+* InfluxDB: could quite easily be done
```
asim, line 11: For now lets stick with "memory" and "prometheus". If the interface has to change we want to make our lives easier.

metrics/README.md
```diff
@@ -0,0 +1,22 @@
...
+* Telegraf: almost identical to the InfluxDB implementation
+* Micro: Could we provide metrics over Micro's server interface?
+
+
+Todo
+----
+
+* Include a handler middleware which uses the Reporter interface to generate per-request level metrics
```
asim, line 19: Handler wrapper will end up in micro

metrics/prometheus/reporter.go
```diff
@@ -0,0 +1,85 @@
...
+type Reporter struct {
+	defaultLabels      prometheus.Labels
+	prometheusRegistry *prometheus.Registry
+	counters           metricFamily
+	gauges             metricFamily
+	timings            metricFamily
+}
+
+// New returns a configured prometheus reporter:
+func New(service service.Service) (*Reporter, error) {
```
asim: Micro builds on go-micro so theres no back reference here. See other packages where we define an Options struct and then Option func,

metrics/prometheus/metrics.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"errors"
+	"time"
+
+	"github.com/micro/go-micro/v3/metrics"
+)
+
+// ErrPrometheusIsCruft is a terrible thing to have to do. The Prometheus client should really appreciate errors more:
+var ErrPrometheusIsCruft = errors.New("The venerable Prometheus client panicked. Did you do something like change the tag cardinality or the type of a metric?")
```
domwong: Best. Error. Name. Ever.

metrics/README.md
chrusty: Yes, agreed. I'm really just showing that this doesn't have to be Prometheus-centric here.

vtolstov:
What is the benefits and use case ? I mean that we have metrics wrappers in go-plugins

metrics/prometheus/metrics.go
chrusty: I need to check that this is still a thing, expect this to change to something less crufty.

chrusty:
> What is the benefits and use case ? I mean that we have metrics wrappers in go-plugins

Hi @vtolstov. This is more than a "wrapper" really... it provides a way to expose the Go runtime statistics for your services, as well as a consistent way to add your own arbitrary metrics.

asim:
> What is the benefits and use case ? I mean that we have metrics wrappers in go-plugins

go-micro is defining the standard uses for all distributed systems concerns including metrics. A wrapper merely automates standard metrics. It does not allow for user defined metrics. A metrics interface allows it to become a building block for not only wrappers but also user defined values.

asim:
> > What is the benefits and use case ? I mean that we have metrics wrappers in go-plugins
> 
> Hi @vtolstov. This is more than a "wrapper" really... it provides a way to expose the Go runtime statistics for your services, as well as a consistent way to add your own arbitrary metrics.

Note https://github.com/micro/go-micro/blob/master/debug/stats/stats.go which is a built in handler as Debug.Stats

metrics/prometheus/reporter.go
```diff
@@ -0,0 +1,86 @@
...
+
+// Reporter is an implementation of metrics.Reporter:
+type Reporter struct {
+	defaultLabels      prometheus.Labels
+	prometheusRegistry *prometheus.Registry
+	metrics            metricFamily
+}
+
+// New returns a configured prometheus reporter:
+func New(service *service.Service) (*Reporter, error) {
```
asim: Remove the reference to micro/service. Accept `...Option` like all the other interfaces and provide an option to set labels as `func WithLabel(key, val string) Option`

chrusty: OK I've switched to an `options` based configuration, and it seems to work well!

asim (approved):

metrics/options.go
```diff
@@ -0,0 +1,65 @@
...
+
+	for _, o := range opt {
+		o(&opts)
+	}
+
+	return opts
+}
+
+// PrometheusPath is the path that Prometheus metrics will be made available on:
+func PrometheusPath(value string) Option {
```
asim: Rename to Path and add comment "// Path used to serve metrics over http

metrics/options.go
```diff
@@ -0,0 +1,65 @@
...
+	return func(o *Options) {
+		o.PrometheusPath = value
+	}
+}
+
+// PrometheusListenAddress is the listen address for the Prometheus client:
+func PrometheusListenAddress(value string) Option {
+	return func(o *Options) {
+		o.PrometheusListenAddress = value
+	}
```
asim, line 50: Rename to Address

metrics/options.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// DefaultTags will be added to every metric:
+func DefaultTags(value Tags) Option {
```
asim, line 54: Rename to Tags or Labels

metrics/options.go
```diff
@@ -0,0 +1,65 @@
...
+	return func(o *Options) {
+		o.DefaultTags = value
+	}
+}
+
+// TimingObjectives defines the desired spread of statistics for histogram / timing metrics:
+func TimingObjectives(value map[float64]float64) Option {
```
asim, line 61: I dont understand what TimingObjectives are, is there another name for this?

metrics/options.go
chrusty: done

metrics/options.go
chrusty: done

metrics/options.go
chrusty: This one probably _should_ remain "Default" tags because whatever is provided here will be attached to every metric (in addition to any metric-specific tags).

metrics/options.go
chrusty: This one is difficult to describe succinctly. It is where we define the list of quantiles and their respective error ranges. Each one of these becomes a statistical metric on timings.

Our defaults are `{0.0: 0, 0.5: 0.05, 0.75: 0.04, 0.90: 0.03, 0.95: 0.02, 0.98: 0.001, 1: 0}`

This means that for each timing histogram metric we'll get 50th, 75th, 90th, 95th, 98th, and 100 quantiles (along with min/max/average/count).

Annoyingly this is quite fiddly with Prometheus, less so with other metrics implementations. We could just leave this out (as a hard-coded default), but I guarantee that it will annoy _someone_ if they can't configure it.

From the Prometheus client repo:
```
	// Objectives defines the quantile rank estimates with their respective
	// absolute error. If Objectives[q] = e, then the value reported for q
	// will be the φ-quantile value for some φ between q-e and q+e.  The
	// default value is an empty map, resulting in a summary without
	// quantiles.
```

metrics/options.go
asim: ok

metrics/options.go
asim: Can you provide a comment on this function to provide that clarity. More helpful with some explanation

PR #1932, opened 2020-08-14
events: {memory, nats}

ben-toogood:
The foundations of a solution for https://github.com/m3o/dev/issues/347.


events/options.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+}
+
+// SubscribeOptions contains all the options which can be provided when subscribing to a topic
+type SubscribeOptions struct {
+	// Prefix when true will return any event which has the prefix of the topic, not just exact matches
+	Prefix bool
+	// Queue is the name of the subscribers queue, if two subscribers have the same queue the message
+	// should only be published to one of them
+	Queue string
```
domwong, line 44: Minor : I wonder if queue is the correct terminology. It probably is, but Kafka talks about consumer groups for instance. Still think it's fine but flagging in case anyone can think of something better

events/events.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+// Broker of events
+type Broker interface {
+	Publish(topic string, opts ...PublishOption) error
+	Subscribe(topic string, opts ...SubscribeOption) (<-chan Event, error)
+}
+
+// Store of events
+type Store interface {
```
asim, line 24: Events Store or Events Stream?

events/events.go
```diff
@@ -0,0 +1,37 @@
+package events
```
asim, line 2: Need // Package level comment

events/events.go
ben-toogood: I think store is more descriptive, Stream indicates something you subscribe to and I think that's more of the brokers job.

events/options.go
crufter: I'm happy with queue, seems like most widely used term?

events/events.go
crufter: +1 for store from my side

asim: If this is timeseries like moreso than event streaming then I agree. If its more kafka like then its event streaming. Which is it?

asim: Or do we say that events.Store and events.Stream are two separate things that can coexist here?

events/options.go
asim: Yea queue makes sense, "I subscribe to a queue". The alternative is to give "id" for stream id or similar.

chrusty: "topic" works for logs, streams _and_ queues

asim (approved):

events/nats/nats.go
```diff
@@ -0,0 +1,148 @@
...
+	stan "github.com/nats-io/stan.go"
+	"github.com/pkg/errors"
+
+	"github.com/micro/go-micro/v3/events"
+	"github.com/micro/go-micro/v3/logger"
+)
+
+const (
+	defaultClusterID = "micro"
+	firehoseTopic    = "firehose"
```
asim: Rename to events

PR #1942, opened 2020-08-18
events: implement package with memory & nats streams

ben-toogood:


domwong (approved):
We need a design doc or some commentary/readme about how this works and what it's intended for. Without that context it's a bit abstract for users (although I get that we are the first users of this). 

I see we're also doing dual publish. We should do a check to make sure they're not publishing on the `events` topic otherwise we'll get a double message. We also need to think about recovery scenarios if the first publish succeeds but the second fails.

PR #1959, opened 2020-08-20
Memory and file store list fixes

domwong:
Simplify memory and file store impls, should remove a lot of limit and offset bugs

### DIFF
--- a/store/memory/memory.go
+++ b/store/memory/memory.go
@@ -165,6 +165,8 @@ func (m *memoryStore) list(prefix string, limit, offset uint, prefixFilter, suff
 }
 
 func (m *memoryStore) Close() error {
+	m.Lock()
+	defer m.Unlock()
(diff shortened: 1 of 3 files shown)


store/memory/memory.go
```diff
@@ -113,33 +122,52 @@ func (m *memoryStore) set(prefix string, r *store.Record) {
...
 	}
-
-	sort.Slice(keys, func(i, j int) bool { return keys[i] < keys[j] })
-	return applyLimitAndOffset(keys, limit, offset)
+	return keys
 }
 
 func (m *memoryStore) Close() error {
-	m.store.Flush()
+	for _, s := range m.stores {
```
ben-toogood, line 170: Did you want to lock at this point?

domwong: ooh, good spot

ben-toogood (approved):

PR #1960, opened 2020-08-20
store: remove write TTL & expiry options

ben-toogood:
Closes https://github.com/m3o/dev/issues/387


domwong:
We also need to deprecate or remove from store.proto

EDIT - never mind that's in micro not here.

domwong (approved):

PR #1962, opened 2020-08-21
Observability/metrics update

chrusty:
Fixing a few things with the initial metrics package:

- The `noop` implementation logged annoying initialisation messages (removed those completely)
- The "TimingObjectives" option was confusing, so I've changes this to be a simple list of percentiles (and created a mapping in the Prometheus implementation which picks up the associated error-ranges from a predefined list)
- Added a `logging` implementation which could be used locally to debug things (logs at "Trace" level to prevent it from crufting up people's logs)

### DIFF
--- a/metrics/logging/reporter.go
+++ b/metrics/logging/reporter.go
@@ -1,0 +1,53 @@
+package logging
+
+import (
+	"time"
+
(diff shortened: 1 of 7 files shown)


domwong (approved):
LGTM

PR #1963, opened 2020-08-21
runtime/local: add support for idiomatic folder structures

ben-toogood:
Local implementation for https://github.com/m3o/dev/issues/353


domwong (changes requested):
Needs a test

domwong (approved):

PR #1964, opened 2020-08-21
runtime/local: fix unknown dir path

ben-toogood:
When compressing source, the micro CLI names a tar "foo.tar.gz". However, since this is likely a mono-repo the source which in-fact needs to be run is inside the foo folder once uncompressed. As per the comments in the micro run, this folder name should be extracted from the name of the zip.


crufter (approved):

crufter:
I have tested this with `m3o/services` so unless it breaks `micro/micro` tests, this can be merged.

PR #1969, opened 2020-08-24
check stop singal first

v8lab:
if the stream is closed check first, avoid random selection.

### DIFF
--- a/util/ring/buffer.go
+++ b/util/ring/buffer.go
@@ -33,6 +33,15 @@ type Stream struct {
 	Stop chan bool
 }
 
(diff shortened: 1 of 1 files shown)


util/ring/buffer.go
```diff
@@ -33,6 +33,15 @@ type Stream struct {
 	Stop chan bool
 }
 
+func (s *Stream) isStoped() bool {
```
crufter, line 36: Typo? `isStopped`?

v8lab:
i rethink it, the previous version is better

PR #1981, opened 2020-09-01
Event stream updates

domwong:
Add 
- Manual ack support with `ackWait` timeout
- retry limits (poison message processing)
- tests


events/options.go
```diff
@@ -35,6 +35,16 @@ type SubscribeOptions struct {
 	// StartAtTime is the time from which the messages should be consumed from. If not provided then
 	// the messages will be consumed starting from the moment the Subscription starts.
 	StartAtTime time.Time
+	// ManualAck if true (default false), specifies that each message need ts to be manually acknowledged by the subscriber.
+	// If processing is successful the message should be ack'ed to remove the message from the stream.
+	// If processing is unsuccessful the message should be nack'ed (negative acknowledgement) which will mean it will
+	// remain on the stream to be processed again.
+	ManualAck bool
```
asim: I called this AutoAck in the broker and set it to true. Any preferences on Manual vs Auto?

asim (approved):

events/options.go
domwong: No preference really. I'll change to auto ack to keep it consistent

PR #1992, opened 2020-09-10
add Name to auth.Account as a user friendly alias

domwong:
Means that we can have ID as a UUID that never changes for the account and Name can be used as a user friendly alias which you could change. e.g. User would like to change their email or username.


auth/jwt/jwt.go
```diff
@@ -54,13 +54,17 @@ func (j *jwtAuth) Generate(id string, opts ...auth.GenerateOption) (*auth.Accoun
 	if len(options.Issuer) == 0 {
 		options.Issuer = j.Options().Issuer
 	}
-
+	name := options.Name
+	if name == "" {
```
ben-toogood, line 58: Across the rest of the codebase, this is normally written as `len(name) == 0`

ben-toogood (approved):

PR #1993, opened 2020-09-10
Fixing top level run outside repo

crufter:


runtime/local/service.go
```diff
@@ -50,25 +49,6 @@ func newService(s *runtime.Service, c runtime.CreateOptions) *service {
 
 	dir := s.Source
 
-	// For uploaded packages, we upload the whole repo
```
crufter, line 53: Removed this logic as it's pretty much the exact duplicate as what happens above.

ben-toogood (approved):

PR #1998, opened 2020-09-11
Fix running subfolders

crufter:


runtime/local/local.go
```diff
@@ -95,14 +95,14 @@ func (r *localRuntime) checkoutSourceIfNeeded(s *runtime.Service, secrets map[st
 			return err
 		}
 		if len(sourceParts) > 1 {
-			lastFolderPart := s.Name
+			lastFolderPart := tarName
```
crufter, line 98: This is not related to the fix but an existing thing to be fixed.

ben-toogood (approved):

PR #2003, opened 2020-09-14
runtime/builder with golang implementation

ben-toogood:


build/packages.go
```diff
@@ -0,0 +1,31 @@
...
+	Path string
+}
+
+// Location of the binary
+func (b *Binary) Location() string {
+	return b.Path
+}
+
+// DockerImage is a container
+type DockerImage struct {
```
domwong: i know we're unlikely to move away from Docker for now but should this be `ContainerImage` instead?

build/build.go
```diff
@@ -1,32 +1,52 @@
...
-	// Path to the source if local
-	Path string
-	// Language is the language of code
-	Language string
-	// Location of the source
-	Repository string
+// Source code to be built
+type Source interface {
+	// String to describe the source, e.g. foo/api
+	String() string
```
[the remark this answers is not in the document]
ben-toogood: It depends on the object, e.g. LocalSource vs GitSource

build/packages.go
ben-toogood: Yeah, I did consider this - i'm happy with either.

asim: Just use the word Image. Do not be tied to Docker or Container

build/build.go
```diff
@@ -1,32 +1,52 @@
...
 package build
 
-// Build is an interface for building packages
-type Build interface {
+// Builder is an interface for building packages
+type Builder interface {
 	// Package builds a package
-	Package(name string, src *Source) (*Package, error)
-	// Remove removes the package
-	Remove(*Package) error
```
asim: If you have no way of removing docker images then there's nothing that enables a cleanup process, same for tarballs and other sources. Its important that we have a way of actually removing them.

ben-toogood: I'm not sure if this falls outside the responsibility of the builder. I'll have a think about this and get back to you.

asim: My interface had Remove. Please add back remove.

runtime/builder/golang/golang.go
```diff
@@ -0,0 +1,189 @@
...
+	if len(options.Entrypoint) == 0 {
+		ep, err := local.Entrypoint(dir)
+		if err != nil {
+			return nil, err
+		}
+		options.Entrypoint = ep
+	}
+
+	// build the binary
+	cmd := exec.Command(g.cmdPath, "build", "-o", "build", filepath.Dir(options.Entrypoint))
```
domwong: If the archive that we decompress has a directory called build I think this might fail. So I think we might want to pick something with a name that's less likely to clash. `micro_build` or something

```
The -o flag forces build to write the resulting executable or object
to the named output file or directory, instead of the default behavior described
in the last two paragraphs. If the named output is a directory that exists,
then any resulting executables will be written to that directory.
```

domwong (approved):

PR #2005, opened 2020-09-15
store: implement s3 blob store

ben-toogood:
Closes https://github.com/m3o/dev/issues/336

### DIFF
--- a/store/s3/s3.go
+++ b/store/s3/s3.go
@@ -99,7 +99,7 @@ func (s *s3) Write(key string, blob io.Reader, opts ...store.BlobOption) error {
 	// check the bucket exists, create it if not
 	if exists, err := s.client.BucketExists(context.TODO(), options.Namespace); err != nil {
 		return err
-	} else if !exits {
+	} else if !exists {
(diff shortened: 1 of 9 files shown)


store/s3/s3.go
```diff
@@ -0,0 +1,149 @@
...
+	var options store.BlobOptions
+	for _, o := range opts {
+		o(&options)
+	}
+	if len(options.Namespace) == 0 {
+		options.Namespace = "micro"
+	}
+
+	// check the bucket exists, create it if not
+	if exits, err := s.client.BucketExists(context.TODO(), options.Namespace); err != nil {
```
domwong:
```suggestion
	if exists, err := s.client.BucketExists(context.TODO(), options.Namespace); err != nil {
```

store/s3/s3.go
```diff
@@ -0,0 +1,149 @@
...
+		return err
+	} else if !exits {
```
domwong:
```suggestion
	} else if !exists {
```

domwong (approved):
LGTM. Refreshingly simple using the minio stuff 👍

PR #2010, opened 2020-09-17
Config interface change

crufter:
Heavily wip


config/config.go
```diff
@@ -2,49 +2,107 @@
...
 )
 
+type Values interface {
+	Get(path string, options ...Option) Value
+	Set(path string, val interface{}, options ...Option)
+	Delete(path string, options ...Option)
+}
+
 // Config is an interface abstraction for dynamic configuration
 type Config interface {
```
ben-toogood: Can we just combine the two interfaces?

config/config.go
```diff
@@ -2,49 +2,107 @@
...
-	Next() (reader.Value, error)
-	Stop() error
+	Values
 }
 
-type Options struct {
-	Loader loader.Loader
-	Reader reader.Reader
-	Source []source.Source
+// NewConfig returns new config
```
ben-toogood: I don't think we need this in the top level, can you please move to go-micro/config/store

ben-toogood (approved):

config/config.go
crufter: I don't get this

config/config.go
crufter: Yeah we can now

config/config.go
ben-toogood: The implementation should't be in the top level package, only the interface. We'll import the implementations separately from within micro.

crazybber:
any plan for file config implement ? @crufter @asim

PR #2011, opened 2020-09-18
proxy/grpc: fix client streaming bug (EOF not sent to the server)

ben-toogood:


proxy/grpc/grpc.go
```diff
@@ -55,11 +53,8 @@ func readLoop(r server.Request, s client.Stream) error {
 			Body:   body,
 		}
 
-		// write the raw request
-		err = req.Codec().Write(msg, nil)
-		if err == io.EOF {
-			return nil
-		} else if err != nil {
+		// send the message to the stream
+		if err := req.Codec().Write(msg, nil); err != nil {
```
domwong, line 57: why don't we care about EOF error here? Because it's a write and not a read?

domwong (approved):

proxy/grpc/grpc.go
ben-toogood: Exactly, if we're writing information then we shouldn't get an EOF error. That should only be returned when reading from a stream.

PR #2016, opened 2020-09-23
Runtime refactoring and NetworkPolicy support

chrusty:
This PR gets us ready to start managing K8S `networkpolicy` from Runtime.

- Added NetworkPolicy to util/kubernetes
- Refactored the Runtime interface so Create / Delete / Update / Logs etc all make use of a common `Resource` interface
- Updated the Kubernetes unit tests so the namespace and networkpolicy creation / deletion is tested through the new interface methods


asim:
Let's not add another set of endpoints that explicitly leak kubernetes requirements. Lets instead rework the interface. https://github.com/m3o/dev/issues/435.

chrusty:
> Let's not add another set of endpoints that explicitly leak kubernetes requirements. Lets instead rework the interface. [m3o/dev#435](https://github.com/m3o/dev/issues/435).

OK I'm happy to have a go at this, since the underlying utils/kubernetes package already does most of the work it shouldn't be too tough.

I'll remove my networkpolicy methods and replace them with generic `CreateResource` and `DeleteResource` methods.

asim:
Namespace creation/deletion also needs to be wrapped up into this since its also a snowflake in the api.

runtime/runtime.go
```diff
@@ -18,6 +20,10 @@ type Runtime interface {
 	CreateNamespace(string) error
 	// DeleteNamespace deletes a namespace in the runtime
 	DeleteNamespace(string) error
+	// CreateResource creates a runtime resource
+	CreateResource(*client.Resource, ...client.CreateOption) error
```
asim: We should not be leaking references to util. runtime should define Resource type itself. Talk to @ben-toogood about this as he has the most knowledge about what we're attempting to do here.

We have two potential avenues. Entirely rewrite the interface to only deal with Resource or add CreateResource/DeleteResource methods and separately refactor the other methods to Run/Kill/Update/Status.

For the sake of progress, Create/DeleteResource is fine but Create/DeleteNamespace need to be removed in that case, we're otherwise just bloating this interface.

chrusty: Agreed. I've included examples of how to do the namesapces using Create/DeleteResource in the unit test. I'd leave removing CreateNamespace until I have a solution that works for you though.

asim: The issue is specifically related to importing "github.com/micro/go-micro/v3/util/kubernetes/client" and requiring the interface to accept client.Resource. This needs to be a Resource type we define in this runtime package and map through. The fact that we have the util package is just a nice to have, the implementation could very well live in the same place but its separate because its useful for multiple components.

chrusty: OK, I've made some changes which improve this.

runtime/resource.go
```diff
@@ -0,0 +1,113 @@
...
+const (
+	TypeNamespace     = "namespace"
+	TypeNetworkPolicy = "networkpolicy"
+	TypeService       = "service"
+)
+
+// Resource represents any resource handled by runtime
+type Resource interface {
+	ID() string
+	Type() string
```
chrusty, line 15: This will save us from doing reflection / trial-and-error assertions to figure out what each resource is, but it _does_ mean that this needs to be set. The `New*` methods of course set it, but this will pose problems for anyone putting these structs together by hand.

runtime/kubernetes/namespace.go
```diff
@@ -0,0 +1,41 @@
...
+	return err
+}
+
+// deleteNamespace deletes a namespace resource
+func (k *kubernetes) deleteNamespace(namespace *runtime.Namespace) error {
+	err := k.client.Delete(&client.Resource{
+		Kind: "namespace",
+		Name: namespace.Name,
+	})
+	if err != nil {
```
ben-toogood, line 123: Duplicate if statement

runtime/resource.go
ben-toogood: I think this is fine, as long as the struct, e.g. Service always returns a const doesn't depend on New() to set a field on the struct. The reason being that there will be tens or hundreds of places that runtime.Service gets initialised in go-micro / micro and this will prevent having to change all those clients to use the New func.

runtime/resource.go
```diff
@@ -0,0 +1,113 @@
...
+import "fmt"
+
+const (
+	TypeNamespace     = "namespace"
+	TypeNetworkPolicy = "networkpolicy"
+	TypeService       = "service"
+)
+
+// Resource represents any resource handled by runtime
+type Resource interface {
```
ben-toogood, line 13: It might be handy to have a String() func, which could return something in the format of "service://namespace@id:version".

runtime/resource.go
```diff
@@ -0,0 +1,113 @@
...
+	return TypeNamespace
+}
+
+// NetworkPolicy represents an ACL of label pairs allowing ignress to a namespace
+type NetworkPolicy struct {
+	// The labels allowed ingress by this policy
+	AllowedLabels map[string]string
+	// Name of the network policy
+	Name string
+	// Namespace the network policy belongs to
```
ben-toogood, line 50: I think this should not be an attribute on the struct but be retrieved from the options as we currently do.

runtime/resource.go
```diff
@@ -0,0 +1,113 @@
...
+	// url location of source
+	Source string
+	// Metadata stores metadata
+	Metadata map[string]string
+	// Status of the service
+	Status ServiceStatus
+}
+
+// NewService mints a new service
+func NewService(name string) (*Service, error) {
```
ben-toogood: This should take Version also. Default a blank value to "latest".

util/kubernetes/api/request.go
```diff
@@ -168,6 +168,9 @@ func (r *Request) request() (*http.Request, error) {
 	case "deployment":
 		// /apis/apps/v1/namespaces/{namespace}/deployments/{name}
 		url = fmt.Sprintf("%s/apis/apps/v1/namespaces/%s/%ss/", r.host, r.namespace, r.resource)
+	case "networkpolicy", "networkpolicies":
```
ben-toogood, line 171: Why the duplicate values?

ben-toogood (dismissed):

runtime/resource.go
chrusty: Sorry this was a non-issue

runtime/kubernetes/kubernetes_test.go
```diff
@@ -47,19 +50,45 @@ func setupClient(t *testing.T) {
 
 func TestNamespaceCreateDelete(t *testing.T) {
 	defer func() {
+		exec.Command("kubectl", "-n", "foobar", "delete", "networkpolicy", "baz").Run()
```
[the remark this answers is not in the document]
chrusty: Yes we could, but this is a fail-safe I suppose. runtime.Delete(networkPolicy) _is_ actually tested here as well.

runtime/kubernetes/namespace.go
chrusty: Oh you found some cruft... guess who must have been copy/pasting. I fixed this now, sorry

util/kubernetes/api/request.go
chrusty: In case anything leaks up from the client I guess, for completeness sake.

runtime/resource.go
chrusty: Sure, added this

runtime/resource.go
chrusty: I've decided to get rid of `ID()`, and replace it with `String()`. I've implemented your suggestion of the format for the Service resource.

domwong (approved):

PR #2019, opened 2020-09-24
runtime: minor fixes for local runtime

ben-toogood:


runtime/cache/cache.go
```diff
@@ -0,0 +1,156 @@
...
+
+// String defines the runtime implementation
+func (c *cache) String() string {
+	return "cache"
+}
+
+// cacheIsValid returns a boolean indicating if a cache initialized at the time provided is still
+// valid now
+func cacheIsValid(t time.Time) bool {
+	return t.Unix() > time.Now().Add(time.Second*-30).Unix()
```
domwong: you can do a `t.After()` call rather than having to convert with `Unix()`

runtime/cache/cache.go
```diff
@@ -0,0 +1,156 @@
...
+	var options runtime.ReadOptions
+	for _, o := range opts {
+		o(&options)
+	}
+	if options.Namespace == "" {
+		options.Namespace = "micro"
+	}
+
+	// if a query was used we can't use the cache
+	if len(options.Service) > 0 || len(options.Type) > 0 || len(options.Version) > 0 {
```
domwong: if there are this many cases where we can't use the cache is it even worth having it?

domwong (commented):
I think it's worth detailing in the PR exactly why this is useful

runtime/cache/cache.go
ben-toogood: Yes, although these options exist, the vast majority of callers won't pass this option.

ben-toogood:
> I think it's worth detailing in the PR exactly why this is useful

Will add more detail

runtime/cache/cache.go
ben-toogood: Also, these options are usually used in tandem with each other

domwong (approved):

asim:
The new focus of go-micro is to just be a simple set of interfaces that rarely change along with their implementations. Anything superfluous like this needed by micro should just exist in micro instead.

PR #2025, opened 2020-09-25
fix collides with imported package name

fengpf:


asim:
This package has moved

fengpf:
ok,thanks

PR #2029, opened 2020-09-29
Fixes for runtime builder

ben-toogood:
Misc fixes required for https://github.com/micro/micro/pull/1397


runtime/options.go
```diff
@@ -11,16 +11,16 @@ type Option func(o *Options)
 
 // Options configure runtime
 type Options struct {
+	// Client to use when making requests
+	Client client.Client
```
asim, line 15: Is there a reason we need client here and is it avoidable?

runtime/options.go
```diff
@@ -70,6 +70,8 @@ type CreateOptions struct {
 	Args []string
 	// Environment to configure
 	Env []string
+	// Entrypoint within the folder (e.g. in the case of a mono-repo)
+	Entrypoint string
```
asim, line 74: What does entrypoint mean? Is it a directory or a main.go? If its main.go let's call this Program, if its a directory let's call it Dir.

asim (approved):

asim (approved):

PR #2035, opened 2020-10-05
runtime/kubernetes: rewrite to improve support of multiple versions of a single service

ben-toogood:
Previously each version of a service had a k8s service resource created, e.g. "foobar-latest". This was problematic for static routing so now multiple versions of the same service can share a single k8s service object, e.g. "foobar". A single deployment is still created for each version.


runtime/kubernetes/namespace.go
```diff
@@ -0,0 +1,116 @@
+package kubernetes
+
+import (
+	"strings"
+
+	"github.com/micro/go-micro/v3/logger"
+	"github.com/micro/go-micro/v3/util/kubernetes/client"
+)
+
+func (k *kubernetes) ensureNamepaceExists(ns string) error {
```
crufter, line 10: I'm not against the concept of namespaces being in go micro as I think it's a perfectly valid thing to have in OSS, but I thought we try to hide it in general?

ben-toogood: This function already existed, I just moved all the namespace related cruft to a namespace.go file. Chris' PR will tidy this up further.

ben-toogood: This function is to ensure the k8s namespace exists before the runtime creates a service & deployment within it

runtime/kubernetes/kubernetes.go
```diff
@@ -313,153 +97,152 @@ func (k *kubernetes) Create(s *runtime.Service, opts ...runtime.CreateOption) er
...
+	}
 
+	// create the deployment
+	if err := k.client.Create(client.NewDeployment(s, options), client.CreateNamespace(options.Namespace)); err != nil {
+		if parseError(err).Reason == "AlreadyExists" {
+			return runtime.ErrAlreadyExists
+		}
 		if logger.V(logger.DebugLevel, logger.DefaultLogger) {
-			logger.Debugf("Generated auth credentials for service %v", s.Name)
+			logger.Debugf("Runtime failed to create deployment: %v", err)
```
domwong: Probably want to bump this to Error level

runtime/kubernetes/kubernetes.go
```diff
@@ -313,153 +97,152 @@ func (k *kubernetes) Create(s *runtime.Service, opts ...runtime.CreateOption) er
...
 
-	// create new service
-	service := newService(s, options)
+	// create the service, one could already exist for another version so ignore ErrAlreadyExists
+	if err := k.client.Create(client.NewService(s, options), client.CreateNamespace(options.Namespace)); err != nil {
+		if parseError(err).Reason == "AlreadyExists" {
+			return nil
+		}
+		if logger.V(logger.DebugLevel, logger.DefaultLogger) {
+			logger.Debugf("Runtime failed to create service: %v", err)
```
domwong: ditto

runtime/kubernetes/kubernetes.go
```diff
@@ -313,153 +97,152 @@ func (k *kubernetes) Create(s *runtime.Service, opts ...runtime.CreateOption) er
...
-		// update the service
-		if err := service.Update(k.client, client.UpdateNamespace(options.Namespace)); err != nil {
+		// update the deployment
+		res := &client.Resource{
+			Kind:  "deployment",
+			Name:  resourceName(s),
+			Value: &dep,
+		}
+		if err := k.client.Update(res, client.UpdateNamespace(options.Namespace)); err != nil {
+			if logger.V(logger.DebugLevel, logger.DefaultLogger) {
```
domwong: ditto

runtime/kubernetes/kubernetes.go
```diff
@@ -469,29 +252,58 @@ func (k *kubernetes) Update(s *runtime.Service, opts ...runtime.UpdateOption) er
 
 // Delete removes a service
 func (k *kubernetes) Delete(s *runtime.Service, opts ...runtime.DeleteOption) error {
+	k.Lock()
```
domwong, line 255: What are we protecting with this lock?

ben-toogood: Nothing, will remove.

runtime/kubernetes/kubernetes.go
```diff
@@ -469,29 +252,58 @@ func (k *kubernetes) Update(s *runtime.Service, opts ...runtime.UpdateOption) er
...
+	if err != nil || len(services) > 0 {
+		return err
+	}
+
+	// delete the service
+	srv := client.NewService(s, &runtime.CreateOptions{
+		Type:      k.options.Type,
+		Namespace: options.Namespace,
+	})
+	if err := k.client.Delete(srv, client.DeleteNamespace(options.Namespace)); err != nil {
```
domwong, line 303: In theory there's a tiny race condition here where someone is adding a new version at the same time someone else is deleting a version which would cause a delete of the k8s service when in fact there is a newly created micro service. small enough of a chance that I think we'll get away with it for now

runtime/kubernetes/namespace.go
```diff
@@ -0,0 +1,116 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// namespaceExists returns a boolean indicating if a namespace exists
+func (k *kubernetes) namespaceExists(name string) (bool, error) {
+	// populate the cache
+	if k.namespaces == nil {
```
domwong, line 40: Does this need a mutex?

domwong: also - how does this cache work if we're running multiple runtime instances? Looks like it only gets populated once so if a different runtime instance creates a namespace then it won't show up in the cache?

runtime/kubernetes/namespace.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	return err
+}
+
+func (k *kubernetes) DeleteNamespace(ns string) error {
+	err := k.client.Delete(&client.Resource{
+		Kind: "namespace",
+		Name: ns,
+	})
+	if err != nil {
```
domwong: woops looks like a double if. Also I think this is missing a delete from `k.namespaces` (appreciate you might have just copy pasted this)

ben-toogood: Basically, I just moved the namespace cruft to a single file. Fixed the double if.

runtime/kubernetes/util.go
```diff
@@ -0,0 +1,206 @@
...
+	p := &client.Resource{
+		Kind:  "pod",
+		Value: podList,
+	}
+	if err := k.client.Get(p, opts...); err != nil {
+		logger.Errorf("Error fetching pods: %v", err)
+		return nil, nil
+	}
+
+	for _, item := range podList.Items {
```
domwong, line 66: minor: if/when we support multiple instances for a service this will only take the status from one pod

ben-toogood: Good point, I think this will require a wider rethink within the package.

util/kubernetes/client/templates.go
```diff
@@ -84,7 +84,7 @@ spec:
           - {{.}}
           {{- end }}
           image: {{ .Image }}
-          imagePullPolicy: Always
+          imagePullPolicy: IfNotPresent
```
domwong, line 87: Not 100% sure how `micro update` works but would this break that? I suspect not because I think the image pulls down a binary rather than having a prebaked image for a service right?

ben-toogood: Now we have a build service this isn't an issue, it's the same cell every time. For micro/platform image we always use tags.

ben-toogood: It was taking 30s longer than needed to start a service for a user. I think it's better to have this as default and then override using an option if we need to.

domwong (approved):

runtime/kubernetes/util.go
fengpf: func transformStatus(depStatus string)  has so many case, if we can use the map optimization.

PR #2049, opened 2020-10-19
Defer Mutex Unlock in cache

fenngwd:
Use defer to make sure mutex unlocked after an unexpected panic in methdo


asim:
Submit something functional please

PR #2074, opened 2020-11-11
Add AutoAck to Broker

domgolonka:
Adding AutoAck to Broker so some future plugins are able to turn off/on ACKnowledgement.


domgolonka:
Example for AutoAck https://github.com/asim/nitro-plugins/pull/602

Issue #2101, opened 2021-01-05
[BUG] zk registry watch some node loss


JasonChen86899:
**Describe the bug**
zk registry watch loss, deleted node will be in cache until ttl
1. What are you trying to do?  
 when children change,just get new children and override old children
2. What did you expect to happen?
when new node add to zk, they will add to cache. when node was deleted, they wiil be deleted in cache
3. What happens instead?
when new node was added to zk, it was not in cache. when node was deleted, it was still in cache

**How to reproduce the bug:**
1. use zk to registry services
2. during the request period, nodes are added and reduced continuously. then some deleted node will be in cache

If possible, please include a minimal code snippet here.
```go
func (zw *zookeeperWatcher) watchDir(key string) {
	for {
		// get current children for a key
		children, _, childEventCh, err := zw.client.ChildrenW(key)
                ......
		select {
		case e := <-childEventCh:
                        // get again
			newChildren, _, err := zw.client.Children(e.Path)
			for _, i := range newChildren {
                        // proccess
                        ......
                        }
                }
          // next get children will loss changes during before proccessing time
         }
}
```

PR #2102, opened 2021-01-05
Fix zk watch loss

JasonChen86899:
#2101
1. override old children with new children when children changed
2. add override case for cache


JasonChen86899:
close #2101

PR #2116, opened 2021-02-04
add read yaml config from file

august6th:
add read yaml config from file


asim:
See https://github.com/asim/go-micro/tree/master/plugins/config/encoder/yaml

PR #2141, opened 2021-03-26
fixing server info overwrite when change default server

### LINKED ISSUE #2140: [BUG]Service name overwrite when using grpc server plugin
**Describe the bug**

1. What are you trying to do?
use grpc server plugin
2. What did you expect to happen?
service name is "testing"
3. What happens instead?
service name is "go.micro.server"

**How to reproduce the bug:**
```go
package main

import (
	"github.com/asim/go-micro/v3"
	_ "github.com/asim/go-micro/plugins/server/grpc/v3"
	"github.com/asim/go-micro/v3/logger"
)

func main() {
	srv := micro.NewService()
	srv.Init(micro.Name("testing"))
	if err := srv.Run(); err != nil {
		logger.Fatal(err)
	}
}
```
```shell
go run . --server grpc
2021-03-26 18:48:47  file=v3@v3.5.0/service.go:192 level=info Starting [service] go.micro.server
2021-03-26 18:48:47  file=v3@v3.0.0-20210317093720-0a41e6d80f39/grpc.go:869 level=info Server [grpc] Listening on [::]:59999
2021-03-26 18:48:47  file=v3@v3.0.0-20210317093720-0a41e6d80f39/grpc.go:702 level=info Registry [mdns] Registering node: go.micro.server-5a7fe6a8-ff9f-44be-a6cc-43505d08a3e3
```

xpunch:
Apply configuration after service get initialized, will not overwrite after server replaced with custom server. 

Close #2140

### DIFF
--- a/service.go
+++ b/service.go
@@ -57,11 +57,6 @@ func (s *service) Name() string {
 // which parses command line flags. cmd.Init is only called
 // on first Init.
 func (s *service) Init(opts ...Option) {
-	// process options
-	for _, o := range opts {
-		o(&s.opts)
-	}
-
 	s.once.Do(func() {
 		// setup the plugins
 		for _, p := range strings.Split(os.Getenv("MICRO_PLUGIN"), ",") {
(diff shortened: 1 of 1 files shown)


asim:
It's a known bug, but what this implies is that if you change the order of execution, flags don't override code changes

xpunch:
So should we redesign the way go-micro share configurations with plugins?

PR #2168, opened 2021-05-15
fix: "Solve the problem that the resources have not been fully released due to early exit"

JeffreyBool:


JeffreyBool:
@asim

asim:
There's a bunch of random go mod changes and the build is broken

JeffreyBool:
> There's a bunch of random go mod changes and the build is broken

It's not mod, it's because I exposed the run error

JeffreyBool:
@asim  It's solved

PR #2180, opened 2021-06-17
Windows event log plugin

korableg:
Hello, I created logger plugin for windows event log.

The plugin is a wrapper over https://pkg.go.dev/golang.org/x/sys/windows/svc/eventlog.
It's library implements windows event log only partly. For example not implemented Fields function (it just returns the same object)

Good wishes for you!


asim:
Looking for an external reviewer

PR #2184, opened 2021-06-22
support etcd auth with env args

xpunch:
set etcd auth username and password from env args
set default registry address with env arg instead of 127.0.0.1

default etcd address 127.0.0.1 is invalid in prod environment usually.

### DIFF
--- a/plugins/registry/etcd/etcd.go
+++ b/plugins/registry/etcd/etcd.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"errors"
 	"net"
+	"os"
 	"path"
 	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


plugins/registry/etcd/etcd.go
```diff
@@ -36,10 +37,14 @@ type etcdRegistry struct {
 
 func NewRegistry(opts ...registry.Option) registry.Registry {
 	e := &etcdRegistry{
-		options:  registry.Options{},
+		options:  registry.Options{Addrs: []string{os.Getenv("MICRO_REGISTRY_ADDRESS")}},
```
asim: if the env var is blank this will create a blank entry, regardless of whether its parsed later, that's not useful. Please remove.

xpunch: I've removed it, it will be added into registry options as default address instead of 127.0.0.1:2379 when it's not blank.

xpunch: BTW, the unit test passed in my local environment, shouldn't be an issue caused by my codes.

PR #2187, opened 2021-06-30
fix plugins go get bug.

iGoogle-ink:
1.fix plugins go get bug.
2.update all mode.
3.add tidy tools


asim:
You committed a 7mb file to this repo. Unbelievable.

PR #2193, opened 2021-07-12
fix logger v3

JeffreyBool:
fix logger v3


JeffreyBool:
@asim  rabbitmq logger import wrong address

asim (approved):

PR #2206, opened 2021-08-11
plugins/client/grpc: fix err code conversion

unvoodoo:
Issue: grpc client plugin uses grpc code as micro error status. 
if client gets an error from grpc server it will not convert grpc error code to micro http status.

So, if you use go-micro/api with grpc server/client, you get a crash because rpc handler would try to write this code as http status in headers (https://github.com/asim/go-micro/blob/master/api/handler/rpc/rpc.go::460)
```
       ...
       default:
            w.WriteHeader(int(ce.Code))
       }
      ...
```

**current state in asim/go-micro** https://github.com/asim/go-micro/blob/master/plugins/client/grpc/error.go:
```
        ...
        // fallback
	return &errors.Error{
		Id:                   "go.micro.client",
		Code:                 int32(s.Code()),     // issue here, because its a grpc code
		Detail:               s.Message(),
		Status:               s.Code().String(),
	}
```

**current state in micro/micro**: it was fixed by returning of InternalServerError  https://github.com/micro/micro/blob/master/service/client/grpc/error.go
```
// fallback
return errors.InternalServerError("go.micro.client", s.Message())
```

In this PR I propose to convert grpc status codes to micro http statuses similar to  what grpc server plugin does, but vice versa
https://github.com/asim/go-micro/blob/master/plugins/server/grpc/error.go


asim:
Not sure about this. Need someone else to take a look. You also have merge conflicts.

unvoodoo:
@asim thanks, resolved the conflict

unvoodoo:
@asim should I resubmit new PR without merge?

PR #2223, opened 2021-08-30
Add gomu CLI tool

AuditeMarlow:
This PR migrates the [Gomu][1] CLI tool to Go Micro. It has features like some of the functionality found in the Micro toolkit.

* Generate project templates for both services and functions
* Continuously build and run services
* List running services
* Describe running services
* Call running services
* Stream requests from and to services
* Optionally integrates with [Jaeger][2]
* Optionally generate [Skaffold][3] configuration for continous development in the cloud

Closes #2220.

[1]: https://github.com/AuditeMarlow/gomu
[2]: https://jaegertracing.io
[3]: https://skaffold.dev

Issue #1154, opened 2020-02-03
`go run  main.go --help` print help and continue running the app


xmlking:
**Describe the bug**

1. What are you trying to do?
> run my apps with `--help`  arg
2. What did you expect to happen?
> should print help and exit.
3. What happens instead?
> print help and continue running app

**How to reproduce the bug:**

clone https://github.com/xmlking/micro-starter-kit

run
```bash
go run srv/greeter/main.go --help
```

**Environment:**
Go Version: please paste `go version` output here
```
go version go1.13.6 darwin/amd64
```

> go run srv/greeter/main.go --help                       

```
                      
NAME:
   greetersrv - a go-micro service

USAGE:
   main [global options] command [command options] [arguments...]

COMMANDS:
   help, h  Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --client value                  Client for go-micro; rpc [$MICRO_CLIENT]
   --client_request_timeout value  Sets the client request timeout. e.g 500ms, 5s, 1m. Default: 5s [$MICRO_CLIENT_REQUEST_TIMEOUT]
   --client_retries value          Sets the client retries. Default: 1 (default: 1) [$MICRO_CLIENT_RETRIES]
   --client_pool_size value        Sets the client connection pool size. Default: 1 (default: 0) [$MICRO_CLIENT_POOL_SIZE]
   --client_pool_ttl value         Sets the client connection pool ttl. e.g 500ms, 5s, 1m. Default: 1m [$MICRO_CLIENT_POOL_TTL]
   --register_ttl value            Register TTL in seconds (default: 60) [$MICRO_REGISTER_TTL]
   --register_interval value       Register interval in seconds (default: 30) [$MICRO_REGISTER_INTERVAL]
   --server value                  Server for go-micro; rpc [$MICRO_SERVER]
   --server_name value             Name of the server. go.micro.srv.example [$MICRO_SERVER_NAME]
   --server_version value          Version of the server. 1.1.0 [$MICRO_SERVER_VERSION]
   --server_id value               Id of the server. Auto-generated if not specified [$MICRO_SERVER_ID]
   --server_address value          Bind address for the server. 127.0.0.1:8080 [$MICRO_SERVER_ADDRESS]
   --server_advertise value        Used instead of the server_address when registering with discovery. 127.0.0.1:8080 [$MICRO_SERVER_ADVERTISE]
[...]

PR #2224, opened 2021-08-30
Move from micro/cli/v2 to urfave/cli/v2

AuditeMarlow:
@asim mentioned it may be favourable to move away from using github.com/micro imports in #2223, so I thought in undoing my own errors, I might as well elevate the Go Micro source code. I'm opting to use github.com/urfave/cli/v2, since github.com/micro/cli/v2 was forked from that, so it's pretty simple to move to without extending too much effort.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -22,7 +22,7 @@ import (
 	"github.com/asim/go-micro/v3/server"
 	"github.com/asim/go-micro/v3/store"
 	"github.com/asim/go-micro/v3/transport"
-	"github.com/micro/cli/v2"
+	"github.com/urfave/cli/v2"
 )
 
 type Cmd interface {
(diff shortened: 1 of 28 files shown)


AuditeMarlow:
Unfortunately this PR reopens #1154.

asim:
Thanks, much appreciated. When we forked, there was an idea to create a go-micro/cli package of our own but its quite a bit of work. Potentially something to think about in the future since we could abstract over various CLI packages.

AuditeMarlow:
This PR requires a new release I think. Either that or I have to figure out another way to fix Gomu. Gomu now breaks because it depends on the updated `github.com/urfave/cli/v2` imports, but is stuck using the `github.com/micro/cli/v2` imports in `github.com/asim/go-micro/v3@v3.6.0`.

asim:
It's a very easy fix. Go modules can use any commit hash, not just release versions. So you quite simply. Set it to use the latest commit in the gomu go.mod or specify the latest commit hash of go-micro. Then run `go fmt ./...` or `go test -v ./...` in gomu directory and it will fix the go.mod to be in the format it expects.

```
-       github.com/asim/go-micro/v3 v3.6.0
+       github.com/asim/go-micro/v3 v3.6.1-0.20210831082736-088ccb50019c
```

I'll commit it now but you can do so in future if there's ever a need to update.

PR #2231, opened 2021-08-31
Add simple in-memory cache

### LINKED ISSUE #2227: [FEATURE] Cache package
Furthering our building block packages, it's clear a Cache is a design pattern that's frequently used in distributed systems with a simple Get/Put/Delete interface. For anyone looking to contribute this, we welcome that plus some initial implementations. While the store package enables Read/Write/Delete semantics it also focuses on more long term storage much like a database and operates on byte values. With the cache package, it would be nice to use just simple Go interfaces and implementations such as redis, memcache and memory.

Here's an example interface:

```
type Cache interface {
  Get(key string, val interface{}) error
  Put(key string, val interface{}) error
  Delete(key string) error
}
```

We can additionally add a method for context use like so..

```
type Cache interface {
   ...
   Context(ctx context.Context) Cache
}
```

This returns a copy of Cache with the context set. This is good design practice so we don't pollute all methods with context as the first arg. Usage looks like so.

```
err := cache.Context(ctx).Put("foo", "bar")
```

AuditeMarlow:
This PR adds a `cache` package with a default in-memory cache implementation registered in the `micro` and `cmd` packages. The `cache` package is fully documented and tested.

Closes #2227.

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -51,8 +51,15 @@ func (i *Item) Expired() bool {
 
 // NewCache returns a new cache.
 func NewCache(opts ...Option) Cache {
+	options := NewOptions(opts...)
+	items := make(map[string]Item)
+
+	if len(options.Items) > 0 {
+		items = options.Items
+	}
+
 	return &memCache{
-		opts:  NewOptions(opts...),
-		items: make(map[string]Item),
+		opts:  options,
+		items: items,
 	}
 }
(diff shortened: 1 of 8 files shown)


asim:
Very clean code. Thanks for the contribution. We can then go on to add memcache/redis implementations in the plugins directory or allow others to contribute those as and when its relevant.

PR #2232, opened 2021-09-01
Add cache example

AuditeMarlow:
This PR adds a cache example. It demonstrates the usage of the recently added `cache` package. It also contains shameless advertising of the Gomu CLI tool, although I'm wondering if this is the right place to do so.


asim:
I think it it's fine to mention the CLI tool. I've also added it to the main readme.

PR #2243, opened 2021-09-03
Add new client cmd

AuditeMarlow:
With this change, Gomu users will be able to generate template projects
for clients to services. Additionally vendor support has been built in
so Gomu users can now generate projects using fully qualified package
names, for example:

```bash
gomu new service github.com/auditemarlow/helloworld
```

This will create a new service project `helloworld` with its module name
already set to `github.com/auditemarlow/helloworld`. Likewise, Gomu
users can then generate client projects in the same manner:

```bash
gomu new client github.com/auditemarlow/helloworld
```

This will create a `helloworld-client` project that uses the protobufs
found in the `github.com/auditemarlow/helloworld` service. This removes
at least some strain in configuring these module dependencies yourself;
you can just scaffold them outright from the start.

Although the default client project is highly opinionated, it works
straight out of the box and has Skaffold in mind. Gomu users should be
able to get going in a matter of seconds.

### DIFF
--- a/cmd/gomu/README.md
+++ b/cmd/gomu/README.md
@@ -197,6 +197,38 @@ Skaffold pipeline using the `skaffold` command.
 skaffold dev
 ```
 
+## Creating A Client
+
+To create a new client, use the `gomu new client` command. The name is the
+service you'd like to create a client project for.
+
+```bash
+$ gomu new client helloworld
+creating client helloworld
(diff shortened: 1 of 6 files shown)


AuditeMarlow:
Some stuff doesn't exactly work in combination with Skaffold just yet, I'm working on resolving issues there.

PR #2273, opened 2021-09-18
fix(cache): only watch calling service in registry

### LINKED ISSUE #2272: [BUG] cache module will watch every services  when i use consul as a registry
in cache.go , it not say what service  want to watching, 

https://github.com/asim/go-micro/blob/master/registry/cache/cache.go#L345

So, when I use the consul registry plugins, it can't know what service should to watching , it will watch every services in consul instance. This will waste a lot of resources, especially in large consul instances

https://github.com/asim/go-micro/blob/master/plugins/registry/consul/watcher.go#L188-L215

https://github.com/asim/go-micro/blob/master/plugins/registry/consul/watcher.go#L198-L200

 this section was originally used to filter irrelevant services.

PR in here: #2273

Kuri-su:
closes #2272

### DIFF
--- a/registry/cache/cache.go
+++ b/registry/cache/cache.go
@@ -180,7 +180,7 @@ func (c *cache) get(service string) ([]*registry.Service, error) {
 
 		// only kick it off if not running
 		if !c.running {
-			go c.run()
+			go c.run(service)
 		}
 
 		c.Unlock()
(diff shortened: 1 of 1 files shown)


asim:
It's flip flopped like this before and there was a specific reason it's been changed so I need to remember why before merging.

Kuri-su:
ok, i saw the last modify in this section is #919 , to makes a couple optimisations to the mdns registry and cache registry, I will check the impact of this PR on mdns registry.

asim:
We'll merge and see if anyone presents issues

PR #2277, opened 2021-09-19
Command add Run method

simon28082:
**background:**
When I tried to increase the built-in config option, I found that I must increase before cmd.init, because when I execute service.init, execute cmd.init before executing service.init will cause cmd.init Repeat execution.

Of course, I can prioritize the options option in the newcmd (options) method, but this will cause another problem, I can't get the current CMD INSTANCE inside Options, as shown below:
```go
	c := cmd2.NewCmd(
		cmd2.NewConfig(`consul`, func(option ...config.Option) (config.Config, error) {
			// like this
			// I hope to get c var instance
			// but at this time, I am initializing me.
			return config.NewConfig(option...)
		}),
		cmd2.NewConfig(`file`, func(option ...config.Option) (config.Config, error) {
			return config.NewConfig(option...)
		}),
	)

	// execute init method
	c.Init()

	// Will execute cmd.init again
	service.Init()
```
I think this may be unreasonable, I have changed this option.

This may cause incompatibility under scenes that do not use Defaultcmd, but I think this still needs to be updated.


cmd/cmd.go
```diff
@@ -26,15 +26,18 @@ import (
...
-type Cmd interface {
-	// The cli app within this cmd
-	App() *cli.App
-	// Adds options, parses flags and initialise
-	// exits on error
-	Init(opts ...Option) error
-	// Options set within this command
-	Options() Options
-}
+type (
```
asim: Can you leave this type definition on one line please

simon28082: Ok, I will change

asim:
You may need to look at the test output for failures caused by the change, may require a change in the web package.

simon28082:
> You may need to look at the test output for failures caused by the change, may require a change in the web package.

OK, I have been working for a long time, after rest, I will come to FIX this problem.

simon28082:
@asim I'm sorry, I have tried did not find Test Panic in the local (maybe because I was based on the GO1.17 version)
But some Test Case still has no running, I have spent a few hours, trying to find the reason is not found for the time being.
So I will temporarily turn off this PR, but I still think this is an optimized function.
I will spend more time to find problems, until the problem solved, and resubmit Pr

PR #2280, opened 2021-09-21
logger: caller's skip correction

unvoodoo:
**Issue**
Inconsistency with `CallerSkipCount` logger.Option. 
Every logger should be called with **same wrappers' depth** to treat caller's skips correctly, but Helper doesn't  wrap two methods:
```
func (h *Helper) Log(level Level, args ...interface{})
func (h *Helper) Logf(level Level, template string, args ...interface{})
```

**Logger Usage Scenarios**
According to the fact that we have a logger.Logger interface and logger.Helper struct-decorator currently, there are 3 patterns of using logger:
1. **[global]** Using logger's methods like logger.Info, logger.Infof, etc.  - caller skip should be 2 (1 for logger package, 1 for implementation)
2. **[contextual]** In Handlers, while getting logger from context:  logger.Extract(ctx context.Context) returns new Helper. - caller skip should be 2 (1 for helper, 1 for implementation)
3. **[direct]** build minimal logger for any needs (like passing to some package as external dependency) - caller skip should be 1 (1 for implementation).

**Current Easy Solution. What's in commit** 
Add Log and Logf functions to Helper, make a more consistent caller's skip policy

1. added Log/Logf functions to *Helper
```
func (h *Helper) Log(level Level, args ...interface{}) {
	h.logger.Log(level, args...)
}
func (h *Helper) Logf(level Level, template string, args ...interface{}) {
	h.logger.Logf(level, template, args...)
}
```

2. walked through implementations: 
 - **zap**: ok. Changed zap plugin to respect common CallerSkipCount option, no need to have its own option in context.
 - **windowseventlog**: ok. Caller skip is not used.
 - **zerolog**: ok. It uses global zerolog logger, and used to set caller's skip globally to 4. Anyway, I added logic to use for skip passed through options and treat 4 as a default.
 - **apex**: ok. Caller skip is not used.
 - **logrus**: not ok and it wasn't ok for a long time. Caller skip is correct with this plugin only in Scenario 3. Logrus doesn't support caller's skip passing as far as I see. Possible solution: to add custom field "caller"  to logrus, calculated same way like in the default logger.

### DIFF
--- a/logger/helper.go
+++ b/logger/helper.go
@@ -6,11 +6,11 @@ import (
 )
 
 type Helper struct {
-	Logger
+	logger Logger
 }
 
-func NewHelper(log Logger) *Helper {
-	return &Helper{Logger: log}
+func NewHelper(logger Logger) *Helper {
+	return &Helper{logger: logger}
 }
 
 // Extract always returns valid Helper with logger from context or with DefaultLogger as fallback.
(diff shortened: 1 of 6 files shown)


unvoodoo:
I've got also some thoughts/questions about Logger/Helper. Started a discussion here https://github.com/asim/go-micro/discussions/2282

PR #2283, opened 2021-09-23
logger helper: add "inject" method, to make a pair with "extract"

unvoodoo:
Adding "Inject" method to Helper, to work in pair with "Extract" while building log wrappers.

Now you can do following:

```
func LogSomething() server.HandlerWrapper {
	return func(h server.HandlerFunc) server.HandlerFunc {
		return func(ctx context.Context, req server.Request, rsp interface{}) error {
			l := logger.Extract(ctx)

			l.Infof("logger from context")

			l = l.WithFields(map[string]interface{}{
				"additional_field": "123",
			})

			ctx = l.Inject(ctx)
			
			return h(ctx, req, rsp)
		}
	}
}

```

### DIFF
--- a/logger/helper.go
+++ b/logger/helper.go
@@ -14,14 +14,18 @@ func NewHelper(logger Logger) *Helper {
 }
 
 // Extract always returns valid Helper with logger from context or with DefaultLogger as fallback.
-// Can be used in pair with function NewContext(ctx context.Context, l Logger) context.Context.
-// (e.g. propagate RequestID to logger in service handler methods).
+// Can be used in pair with function Inject.
+// Example: propagate RequestID to logger in service handler methods.
 func Extract(ctx context.Context) *Helper {
 	if l, ok := FromContext(ctx); ok {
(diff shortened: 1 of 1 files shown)


asim:
Can you just call it Context?

unvoodoo:
Thanks for reply, @asim.

Currently we have:
- 2 methods to work directly with Logger already: NewContext / FromContext. 
- 1 method for Helper - "Extract". That's why I went with "Inject" naming. 
This way we have two pairs:
Logger.NewContext/FromContext
Helper.Extract/Inject

Actually, after this patch, typical usage of logger can be limited to 2 scenarios:
1. logger package: logger.Info, logger.Infof...
2. helper from context, that provides similar to logger package methods - Extract/Inject

but Logger's related NewContext/FromContext - should be in place for anyone who preferred minimalistic Logger.

Wouldn't it be confusing to have Context method working with Helper and NewContext working with Logger? What do you think?

asim:
Sure lets go with that.

PR #2290, opened 2021-09-30
Errors

xpunch:
Add errors.As to finds the first error in err's chain that matches *Error.
If target err don't match *Error, return false,nil.


errors/errors.go
```diff
@@ -144,3 +145,12 @@ func FromError(err error) *Error {
 
 	return Parse(err.Error())
 }
+
+// As finds the first error in err's chain that matches *Error
+func As(err error) (bool, *Error) {
```
asim: Should return (*Error, bool) ?

xpunch: Yeah, it's better to be comma, ok style, already updated.

PR #2311, opened 2021-10-15
Plugins

xpunch:
update plugins release script


asim:
Do you want a separate windows script? Or maybe to detect OS? Because I don't use windows.

asim:
When you get a chance can you also look at cmd/protoc-gen-micro, I think it might suffer from similar issues and maybe examples repo too

xpunch:
Ok,I'll handle those modules too.

asim:
Thank you

asim:
If you learn of how to make vanity URLs and sub modules work let me know. I fought that quite hard and lost

xpunch:
The offical document about [Multi Module Repositories](https://github.com/golang/go/wiki/Modules#what-are-multi-module-repositories) didn't mention vanity URLs for submodules too, seems there is no very good way to do so.

Issue #2314, opened 2021-10-19
[FEATURE] util/pool should limit client max open connections at least.


j2gg0s:
**Is your feature request related to a problem? Please describe.**
A crazy client will create a lot of connections.

**Describe the solution you'd like**
Support limit MaxOpenConns, like golang http's [MaxConnsPerHost](https://github.com/golang/go/blob/master/src/net/http/transport.go#L206).

**Additional context**

xpunch:
You can use [hystrix](github.com/asim/go-micro/plugins/wrapper/breaker/hystrix/v3) to limit connections by setting MaxConcurrentRequests.

j2gg0s:
@xpunch link is not found, do you mean https://github.com/asim/go-micro/tree/master/plugins/wrapper/breaker/hystrix ?

xpunch:
Yeah, hystrix can used to limit the max concurrent count for requests. It usually used to be circuit breaker in distributed systems, so it may provided more function then you need.
You can disable timeout by setting a big timeout duration, disable circuit breaker by setting custom filter, like this:
```
package main

import (
	"context"

	hystrix "github.com/asim/go-micro/plugins/wrapper/breaker/hystrix/v3"
	"github.com/asim/go-micro/v3"
)

func main() {
	hystrix.ConfigureDefault(hystrix.CommandConfig{Timeout: 100000, MaxConcurrentRequests: 100})
	service := micro.NewService(micro.WrapClient(hystrix.NewClientWrapper(hystrix.WithFilter(func(c context.Context, e error) bool {
		return true
	}))))
	service.Init(micro.Name("test.srv"), micro.Address(":80"))
	if err := service.Run(); err != nil {
		panic(err)
	}
}
```
BTW, the max concurrent request is set to every separate request name, not the total concurrent request of the client. You can also set custom config for specific request by using ConfigureCommand(name string, config CommandConfig) , if you know which request may become crazy.

j2gg0s:
hytrix is ok and wonderful!
Improve pool maybe necessary but not urgent.

asim:
If you'd like limits make a PR

PR #2322, opened 2021-10-25
feat(util/pool): new pool limit maxIdleConns and maxConns

j2gg0s:
Link issue #2314 

I'm trying to add new pool suport maxIdleConns and maxConns, like golang's http/1 transport.

As a newbie of go-micro, I hope to gather some opinions before I actually done it.

### DIFF
--- a/util/pool/default.go
+++ b/util/pool/default.go
@@ -4,8 +4,8 @@ import (
 	"sync"
 	"time"
 
-	"go-micro.dev/v4/transport"
 	"github.com/google/uuid"
+	"go-micro.dev/v4/transport"
 )
 
 type pool struct {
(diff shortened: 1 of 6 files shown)


asim:
@xpunch review?

xpunch:
OK, I'll take a look.

xpunch:
I'm not sure whether a connection pool should provide limitation for max concurrent connection, the use of connection pools is primarily to avoid the cost of setting up new connections all the time, so the intention is to re-use them as frequently as possible.
BTW, the unit tests failed and there is no examples demonstrate how the limit pool works and what scenario it can handle.

asim:
It's not clear to me who's going to maintain this code and with the build not passing and no response in 9 days I'm basically going to say it won't get merged. What's more appropriate is probably if we just accept that client.Pool can be swapped out. So we'll work towards that instead.

Issue #2319, opened 2021-10-21
[FEATURE]How to use client.Stream like grpc.CloseSend()


bootfirst:
I use the plugins/client/grpc to send request, but I don't know how to close the send
(the server know the client's close and will close too, so I can get a io.EOF to finish the stream request)

https://github.com/asim/go-micro/issues/2212
xpush said "I used a workaround in my project by sending an empty object to present EOF, not a very good solution, but solved my issue."
SO, HOW? WHAT DOES "empty object" MEAN?

xpunch:
Currently client.Stream do not support grpc.CloseSend(), cause this only supported by grpc which uses http 2.
"empty object" is used to present EOF, which can be an empty Chunk Data or something should not be empty.
For example, you can send an empty byte slice after client data sending finished,  when the server received an empty byte slice, if start to handle received chunk data.

xpunch:
I've created an example in [stream](https://github.com/xpunch/go-micro-example/tree/main/v3/stream), you can take a look.

bootfirst:
👍🏻

Issue #2318, opened 2021-10-21
how to send message after EOF


GaloisZhou:
I use `gomu new service user` to generate a user service project.

In the file `handler/user.go`, I see the following statements:

```go
		if err == io.EOF {
			log.Infof("Got %v pings total", count)
			return stream.SendMsg(&pb.ClientStreamResponse{Count: count})
		}
```

I don't understand how to `send msg` after `EOF`. 
My understanding is :

1. client close the stream
2. server get `EOF`

At this time. The stream is closed, server cannot send message and client cannot get message.

I look forward to your help to answer my doubts. Thank you!

bootfirst:
if you use origin google.golang.org/grpc client, you can use grpc.CloseSend().
Then the server stream.Recv() will return io.EOF.
Then send the response back.（the stream is bidirectional)

gz1115:
Thanks a lot! I find the method(in stream.go):

```go

func (as *addrConnStream) CloseSend() error {
	if as.sentLast {
		// TODO: return an error and finish the stream instead, due to API misuse?
		return nil
	}
	as.sentLast = true

	as.t.Write(as.s, nil, nil, &transport.Options{Last: true})
	// Always return nil; io.EOF is the only error that might make sense
	// instead, but there is no need to signal the client to call RecvMsg
	// as the only use left for the stream after CloseSend is to call
	// RecvMsg.  This also matches historical behavior.
	return nil
}
```

Issue #2212, opened 2021-08-18
[FEATURE] Support client streams


AuditeMarlow:
I've searched the documentation, plugins and existing issues for examples on how to implement client streams, but haven't found a conclusive way to achieve this with the Go Micro framework just yet. Are there examples out in the wild that I can learn from?

For clarification purposes, I think I've got the server implementation down. Below is some boilerplate code I'm using to test with.

`proto/helloworld.proto`
```proto
syntax = "proto3";

package helloworld;

option go_package = "./proto;helloworld";

service Helloworld {
        rpc ClientStream(stream ClientStreamRequest) returns (ClientStreamResponse) {}
}

message ClientStreamRequest {
        int64 stroke = 1;
}

message ClientStreamResponse {
        int64 count = 1;
}
```

`handler/helloworld.go`
```go
package handler

import (
    "context"
    "io"

    log "github.com/asim/go-micro/v3/logger"

    pb "helloworld/proto"
)

type Helloworld struct {}

func (e *Helloworld) ClientStream(ctx context.Context, stream pb.Helloworld_ClientStreamStream) error {
        var count int64
        for {
                req, err := stream.Recv()
                if err == io.EOF {
                        rsp := &pb.ClientStreamResponse{Count: count}
                        log.Infof("Sending response: %v", rsp)
                        return stream.SendMsg(rsp)
                }
                if err != nil {
                        return err
                }
                log.Infof("Got ping %v", req.Stroke)
                count++
        }
}
```

I can't seem to figure out how to implement the client as there seems to be no way for the client to send `EOF` without closing the stream.

xpunch:
[github.com/asim/go-micro/client/client.go#L66](https://github.com/asim/go-micro/blob/49eccbc85a0f6d64c95a4afb59e75c290ac29db4/client/client.go#L66)
```go
// Stream is the inteface for a bidirectional synchronous stream
type Stream interface {
	// Context for the stream
	Context() context.Context
	// The request made
	Request() Request
	// The response read
	Response() Response
	// Send will encode and send a request
	Send(interface{}) error
	// Recv will decode and read a response
	Recv(interface{}) error
	// Error returns the stream error
	Error() error
	// Close closes the stream
	Close() error
}
```
The abstraction of Stream do not support CloseSend, so there is no way to support send EOF currently I think.

xpunch:
[...]

PR #2323, opened 2021-10-26
[feature] stream CloseSend

xpunch:
Currently there is no way for grpc to send EOF without closing the stream(#2212 #2318 #2319).

Adding CloseSend to client.Stream can solve this issue, but it disobey [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle), streamer which based on http/1 do not support close send.

Although it's not a good design, but I think grpc is a better choice for people who want to use stream, support CloseSend for grpc is more valuable.

Changes:
- add CloseSend to client Stream interface
- support CloseSend for plugins/client/grpc
- grpc stream support connection pool
- update protoc-gen-micro to support CloseSend
- call CloseSend of mucp/http streamer, will throw error "streamer not implemented" 
- examples for grpc stream in examples/stream/grpc


asim:
Can you do something like

```
type Stream interface {
    Closer
}

type Closer interface {
    CloseSend() error
}
```

xpunch:
Which name do you think is the proper name for this interface?
 CloseSend, ClientCloser, SendCloser, Closer?

asim:
I think because its in the client package there's no need to do ClientCloser. Closer is fine for now. Because it'll be an embedded interface in the stream the name can change, what we don't want to do is break the method name. So type Closer interface with CloseSend() method makes sense.

xpunch:
OK, I've moved CloseSend into Closer.

PR #2354, opened 2021-11-18
[fix] ignore unexported field

xpunch:
ignore unexported field when register endpoints

some protobuf unexported field was registered into service endpoints


asim:
Good fix. I'm gonna need similar in Micro

isfk:
This fix need update golang to 1.17

xpunch:
Thanks for your notice, I'll try to fix that.

xpunch:
@isfk Fixed in #2382.

PR #2368, opened 2021-11-29
Adding start-up documentation

JeffreyBool:
Adding start-up documentation


asim:
The goal of the readme example was to be bare minimum. Feel free to add to the examples directory but not this main readme.

PR #2379, opened 2021-12-05
update grpc default server client

JeffreyBool:
update grpc default server client


asim:
It's currently a discussion point, we're not actually actively doing it. @xpunch and I are discussing the right long term approach in an issue which you can comment in.

JeffreyBool:
> 目前这是一个讨论点，我们实际上并没有积极地去做。@xpunch和我正在

Any suggestions? Remove the default rpc implementation? And then make the grpc implementation the default rpc?

asim:
We are discussing what a v5 looks like. Otherwise we may borrow from what Micro does and define the notion of a "profile" which can group plugins and introduce the notion of profile loading for that.

JeffreyBool:
> We are discussing what a v5 looks like. Otherwise we may borrow from what Micro does and define the notion of a "profile" which can group plugins and introduce the notion of profile loading for that.

Stai parlando del `micro profilo` [https://github.com/micro/micro/blob/master/profile/profile.go](https://github.com/micro/micro/blob/master/ profilo/profilo.go)? Non ho familiarità con il micro, ma sono disposto a guardare nel porting su di esso

JeffreyBool:
```
// default the profile for the server
	prof := ctx.String("profile")

	// if no profile is set then set one
	if len(prof) == 0 {
		switch ctx.Args().First() {
		case "service", "server":
			prof = "local"
		default:
			prof = "client"
		}
	}

	// apply the profile
	if profile, err := profile.Load(prof); err != nil {
		logger.Fatal(err)
	} else {
		// load the profile
		profile.Setup(ctx)
	}
```

Simile a qui? Profili di microservizi, giusto? E poi grpc come implementazione rpc predefinita, giusto?

asim:
Something like that yea. Named profiles for different environments and purposes. local, dev, production, or different tech stacks like aws, hashicorp, google, etc.

JeffreyBool:
OK

JeffreyBool:
Vede come questo risponde alle sue aspettative? Non so quali modifiche debbano essere apportate a kubernetes, service e test, quindi lo lascerò vuoto di default e lo vorrà in seguito

asim:
We won't be copying profiles verbatim and it won't happen in this PR. It will likely go throw a design discussion first. Feel free to open an issue to discuss.

Issue #2348, opened 2021-11-15
[help me] migrating from go micro v2 to v3 returns this error {"id":"go.micro.client.codec","code":500,"detail":"EOF","status":"Internal Server Error"}


willpang:
Hi, I am recently updating my go micro from version 2 (v2.9.1) to version 3 (v3.7.0). My code is using Grpc to pass the data to Graphql service.

It was working properly when I was using version 2, but it immediately get this error in version 3.

I didn't change anything in the code, except how I pass the Protobuf object, previously some used copy by value which is treated as an error by go vet, so I change it to all pass by reference.

I tried it on my local, and it is working properly, but when I deployed it to my cloud environment, my Graphql returns this error. What is the cause of this? Is there any guide to properly migrate from version 2 to version 3?

Thank you.

xpunch:
I can't get enough useful information form your description, can you provide a small example to reproduce your issue?

bhoriuchi:
Not sure if its the same issue, but I also ran into grpc issues that were resolved by setting the contentType explicitly per https://github.com/asim/go-micro/issues/2123

willpang:
Thank you @xpunch and @bhoriuchi for replying. I have checked in the logs, the content type is right, it recognized the data as protobuf and set the type to proto-rpc.

After doing some digging, the root cause of my problem is my connection to Spanner DB is context cancelled. If I try to repeatedly send the request, a few may go through. So is there any changes related to timeout from micro v2 to micro v3?

willpang:
This is what my code looks like. Do you have any feedback if there is anything wrong with it?

Here is a service where I put my micro

```
package service

import (
	"context"
	"os"
	"os/signal"
	"syscall"

	k8s "github.com/asim/go-micro/plugins/registry/kubernetes/v3"
	"github.com/asim/go-micro/plugins/selector/static/v3"
	micro "github.com/asim/go-micro/v3"
	"github.com/asim/go-micro/v3/client"
	"github.com/asim/go-micro/v3/server"
	logger "../internal/log"
)

// RPCService is a struct with go-micro service and shutdown function
type RPCService struct {
	Service  micro.Service
	shutdown context.CancelFunc
}

// NewRPCService create a new main web service
func NewRPCService(ctx context.Context, name string) *RPCService {
	ctx, cancel := context.WithCancel(ctx)

	optionsService := []micro.Option{
		micro.Context(ctx),
		micro.Name(name),
	}
	optionsAction := []micro.Option{}

[...]

PR #2380, opened 2021-12-05
go-micro v3 update grpc default server client

JeffreyBool:
Change `go-micro` DefaultServer and `DefaultClient` to `grpc`

#2348


asim:
It's currently a discussion point, we're not actually actively doing it. @xpunch and I are discussing the right long term approach in an issue which you can comment in.

asim:
Same as other issue

PR #2382, opened 2021-12-07
[fix] fixing f.IsExported undefined issue

xpunch:
IsExported needs go1.17

#2354 Added IsExported, but it needs go1.17.


isfk:
Nice! 😀

PR #2385, opened 2021-12-13
[feature] add publish wrapper

xpunch:
add publish wrapper for client

### DIFF
--- a/client/options.go
+++ b/client/options.go
@@ -38,6 +38,9 @@ type Options struct {
 	// Default Call Options
 	CallOptions CallOptions
 
+	// Default Publish Options
+	PublishOptions PublishOptions
+
 	// Other options for implementations of the interface
 	// can be stored in a context
 	Context context.Context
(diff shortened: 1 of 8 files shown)


asim:
Don't do that. We have a Client wrapper, anyone can wrap the client and define a Publish method.

asim:
The only reason a Call wrapper existed was for convenience but if we have wrappers for all methods its messy. Wrappers should likely be an external construct, not one baked into the client.

xpunch:
I've try the client wrapper, it will create client.Call trace span, which makes jaeger think the service dependent on itself.
Be like:
a.srv /abc
     - a.srv b.srv.Service.Call
     - b.srv b.srv.Service.Call

asim:
It's entirely dependent on how you use it. If you're just publishing you don't implement the Call method, you just return the original client Call. If there is a current client wrapper that doesn't work, it just needs fixing but otherwise it's a better alternative than having a specific publish wrapper, this just makes the client messy and as you can see every client implementation has to do it.

My advice, create a wrapper package or put something in util/wrapper and just wrap the client.

JeffreyBool:
Is it going to be a link trace for mq? It could be based on micro's default wrapper and handler? But is it possible to remove the default trace wrapper? It feels like a lot of the configuration is mandatory

PR #2397, opened 2021-12-27
nats config plugin

isfk:
## Add Nats config plugin

[Nats Jetstream](https://docs.nats.io/nats-concepts/jetstream/key-value-store/kv_walkthrough)


dh1tw:
that's a cool idea. We should investigate if the NATS Jetstream Key/Value store could also replace the [nats registry plugin](https://github.com/asim/go-micro/tree/master/plugins/registry/nats)

PR #2409, opened 2022-01-08
Add Kafka asynchronous send support

simon28082:
Now add Kafka asynchronous write messages
In Kafka, synchronization messages are meant to be low performance, perhaps in the production environment, maybe we need to write asynchronous messages.
Now add Options to set the return result of asynchronous messages to ensure that asynchronous messages can be implemented.
use `AsyncProducer(errors chan<- *sarama.ProducerError, successes chan<- *sarama.ProducerMessage)`


plugins/broker/kafka/context.go
```diff
@@ -2,6 +2,7 @@ package kafka
 
 import (
 	"context"
+	"github.com/Shopify/sarama"
```
asim: We format as stdlib imports groups together and then external imports grouped together
 Please move the import

plugins/broker/kafka/options.go
```diff
@@ -24,6 +24,13 @@ func ClusterConfig(c *sarama.Config) broker.Option {
 	return setBrokerOption(clusterConfigKey{}, c)
 }
 
+type asyncProduceErrorKey struct{}
+type asyncProduceSuccessKey struct{}
+
+func AsyncProduce(errors chan<- *sarama.ProducerError, successes chan<- *sarama.ProducerMessage) broker.Option {
```
asim: Call it AsyncProducer?

simon28082:
@asim I understand what you mean, I have corrected the relevant code problem.

And add readme.md to briefly describe how the method you can use

PR #2410, opened 2022-01-09
Fix wrong usage of RWMutex in default cache implementation

733amir:
I saw the usage of `RWMutex` in default cache implementation and wanted to fix it. It should use `RLock` and `RUnlock` mechanism in `Get`.


asim:
It's not wrong, just opted not to use read lock.

PR #2433, opened 2022-02-21
add NATS JetStream events plugin

wkloucek:
Introduce a NATS JetStream events plugin, since go-micro currently only supports NATS Streaming, which is depreceated.


plugins/events/natsjs/nats.go
```diff
@@ -0,0 +1,227 @@
...
+
+	// publish the event to the topic's channel
+	if _, err := s.natsJetStreamCtx.PublishAsync(event.Topic, bytes, pubOpts...); err != nil {
+		return errors.Wrap(err, "Error publishing message to topic")
+	}
+
+	return nil
+}
+
+// Consume to a topic
```
kobergj:
```suggestion
// Consume from a topic
```

plugins/events/natsjs/nats.go
```diff
@@ -0,0 +1,227 @@
...
+func (s *stream) Consume(topic string, opts ...events.ConsumeOption) (<-chan events.Event, error) {
+	// validate the topic
+	if len(topic) == 0 {
+		return nil, events.ErrMissingTopic
+	}
+
+	// parse the options
+	options := events.ConsumeOptions{
+		Group: uuid.New().String(),
```
kobergj, line 136: Group name should be defineable. I want multiple services listening in the same group

kobergj (changes requested):
Looks good in general :+1: Just some smaller comments...

asim:
Thanks for the contribution. Think the test is failing on something unrelated. Looking for a reviewer.

plugins/events/natsjs/nats.go
wkloucek: That's just a "default" group if you don't provide one via the opts.

If you provide one, it'll be set in the next few lines:
```
	for _, o := range opts {
		o(&options)
	}
```

plugins/events/natsjs/nats.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	// serialize the event to bytes
+	bytes, err := json.Marshal(event)
+	if err != nil {
+		return errors.Wrap(err, "Error encoding event")
+	}
+
+	pubOpts := []nats.PubOpt{
+		// TODO: to make de-duplication work, we need to pass the event from the outside as an option
```
wkloucek, line 115: @asim would be de-duplication a thing we want to support? I'm not sure if it can be supported by other event implementations.

https://docs.nats.io/using-nats/developer/develop_jetstream/model_deep_dive#message-deduplication

plugins/events/natsjs/nats.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	// setup the subscriber
+	c := make(chan events.Event)
+	handleMsg := func(m *nats.Msg) {
+
+		ctx, cancel := context.WithCancel(context.TODO())
+		defer cancel()
+
+		// TODO: not supported by go-micro interface
```
wkloucek, line 149: @asim NATS Jetstream supports "stretching" the message ACK timeout (`options.AckWait`) but that would need exposing a `InProgressFunc` to the events channel consumer. Is that something we want and can support with other implementations? see https://docs.nats.io/using-nats/developer/develop_jetstream/model_deep_dive -> `AckProgress`

asim: I think its best to do it in an isolated case specifically for this. You can see the rabbitmq broker plugin for implementation specific options that can be passed through. It's not clear that other implementations would support it so we'd inherit an option nothing can deal with.

Davincible:
Thanks a lot for this pull <3

PR #2439, opened 2022-02-24
Add tls option for sync and etcd plugin

alkaidchen:
add tls option for sync and etcd plugin


asim:
Please remove your vendor dir and reopen a pull request

PR #2441, opened 2022-02-24
[fix] http transport deadlock

Ak-Army:
Fixed dead lock, when try to receive message after http transport closed


xpunch (approved):

xpunch:
related #2374

Issue #2453, opened 2022-03-16
[BUG] mdns registry formats ipv6 Node address incorrectly


chrismoran-blockfi:
Using mdns registry and services listening on tcp6, the registry.Node created for the entry/entries is wrapping the address with an extra [] pair.

Context:
mdns_registry.go:355
```
				// prefer ipv4 addrs
				if len(e.AddrV4) > 0 {
					addr = e.AddrV4.String()
					// else use ipv6
				} else if len(e.AddrV6) > 0 {
					addr = "[" + e.AddrV6.String() + "]" //Line 360
				} else {
					if logger.V(logger.InfoLevel, logger.DefaultLogger) {
						logger.Infof("[mdns]: invalid endpoint received: %v", e)
					}
					continue
				}
				s.Nodes = append(s.Nodes, &Node{
					Id:       strings.TrimSuffix(e.Name, "."+p.Service+"."+p.Domain+"."),
					Address:  net.JoinHostPort(addr, fmt.Sprint(e.Port)),
					Metadata: txt.Metadata,
				})
```
The wrapping at line 360 is not necessary. `net.JoinHostPort` does the necessary wrapping if the address is an ipv6 address. This extra wrapping causes later `net.Dial` attempts to fail because the address is invalid.

I've fixed this is my fork and have included a PR.

PR #2454, opened 2022-03-16
#2453 Fix with associated test update

chrismoran-blockfi:
I did not see a develop branch (per the PR template).

This PR fixes an issue with mdns_registry wrapping ipv6 addresses with an extra `[]`. This causes the address to be invalid and `net.Dial` calls fail because of the extra `[]`.


xpunch (approved):

xpunch:
Currently this repo do not follow the git-flow, so there is mainly two brach: master and v3. Master used to maintain the latest version of go-micro, and the v3 used to maintain the go-micro@v3, mostly do some bug fixing.

PR #2468, opened 2022-04-04
Add headers handling for kafka broker plugins

grissom1:
Submitted PR are following these lines

```
if err := h.kopts.Codec.Unmarshal(msg.Value, &m); err != nil {
	p.err = err
	p.m.Body = msg.Value
	if eh != nil {
		eh(p)
	} else {
		log.Errorf("[kafka]: failed to unmarshal: %v", err)
	}
	continue
}
```
I'm not even entirely sure why this code is here. Because it doesn't make sense to me, we are actually unmarshal []bytes to Broker.Message which is
```
{
    Header: map[string]string
    Body: []byte
}
```

But just to be conservative(these line usually doesn't generate error, and even it does, can be handled through eh()), I'm adding lines to handle headers right after it.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module go-micro.dev/v4
+module github.com/grissom1/go-micro/v4
 
 go 1.17
 
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/grissom1/go-micro/v4
+module go-micro.dev/v4
 
 go 1.17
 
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -1,4 +1,4 @@
-module go-micro.dev/v4
+module github.com/grissom1/go-micro/v4
```
asim, line 1: This needs to be changed back

PR #2469, opened 2022-04-04
Update kafka broker plugin to support headers

grissom1:
Revert root module back to go-micro.dev/v4


asim:
You need to merge master. There's an additional unnecessary change

grissom1:
I see. Let me try to clean it up.

PR #2474, opened 2022-04-07
Support direct generation of grpc method

bosima:
Support direct generation of grpc method when the package and service names of proto files are different.

The 'method' generated by protoc-gen-micro does not contain 'package', when the proto file contains a package declaration. When the package and service names of proto files are different, and use service discovery, the correct method cannot be accessed.

1. 
proto example:

> syntax = "proto3";
> option go_package="/hello";
> package HelloA;
> 
> service HelloA{
> ....
> }

2. 
But the registered service name is: Hello.

3. 
Use the following way,  the corresponding method cannot be accessed, when accessing grpc services developed by other frameworks or languages.
`client := NewHelloAService("Hello", service.Client())`

Because the package name is not included when the method is called in the generated proxy.
 `req := c.c.NewRequest(c.name, "HelloA.XXX", in)`

--------------
protoc-gen-go-grpc handles this case correctly, so I think support for this should be added.

The code generated by protoc-gen-go-grpc is:
`err := c.cc.Invoke(ctx, "/HelloA.HelloA/XXX", in, out, opts...)`

After modification, we can: 
protoc --go_out=. --micro_out=. --micro_opt=use_grpc=1 xxx.proto

Set the parameter 'use_grpc', regardless of the value, the package name will be added when the method is generated.

This modification can also be used to solve the problem of calling gRPC services developed by other frameworks or languages.

### DIFF
--- a/cmd/protoc-gen-micro/plugin/micro/micro.go
+++ b/cmd/protoc-gen-micro/plugin/micro/micro.go
@@ -131,7 +131,8 @@ func (g *micro) generateService(file *generator.FileDescriptor, service *pb.Serv
 
 	origServName := service.GetName()
 	serviceName := strings.ToLower(service.GetName())
-	if pkg := file.GetPackage(); pkg != "" {
+	pkg := file.GetPackage()
+	if pkg != "" {
 		serviceName = pkg
 	}
 	servName := generator.CamelCase(origServName)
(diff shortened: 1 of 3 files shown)


xpunch:
closes #2476

bosima:
@xpunch I have rollbacked two unrelated commits.

xpunch:
> @xpunch I have rollbacked two unrelated commits.

You don't need to revert your commits, I would have approved your PR.
Closes #XXX means it'll close target issue once PR was merged.

xpunch (approved):

bosima:
> > @xpunch I have rollbacked two unrelated commits.
> 
> You don't need to revert your commits, I would have approved your PR. Closes #XXX means it'll close target issue once PR was merged.

Sorry, i will resubmit them.

xpunch (approved):

PR #2479, opened 2022-04-14
Make Grpc server support service name with package.

bosima:
At present, gRPC clients developed by other frameworks or languages cannot access the gRPC services developed by go-micro, because the gRPC method in go-micro does not contain the package name.

After modification, we can:
protoc --go_out=. --micro_out=. --micro_opt=use_grpc=1 xxx.proto

Then the struct of the service will contain a fullName field:

    func RegisterHelloHandler(s server.Server, hdlr HelloHandler, opts ...server.HandlerOption) error {
    
	    type hello interface {
		    Say(ctx context.Context, in *SayRequest, out *SayResponse) error
	    }
	    type Hello struct {
		    hello
		    fullName string
	    }
	    h := &helloHandler{hdlr}
	    rh := &Hello{
		    hello:    h,
		    fullName: "Business.Hello",
	    }
	    return s.Handle(s.NewHandler(rh, opts...))
    }

Then the gRPC handler will try to get the fullName, and if so, use it as the prefix of the method name, otherwise continue to use the previous scheme.

    func newRpcHandler(handler interface{}, opts ...server.HandlerOption) server.Handler {
	   ...
    
	    typ := reflect.TypeOf(handler)
	    hdlr := reflect.ValueOf(handler)
	    hdlrv := reflect.Indirect(hdlr)
	    name := hdlrv.Type().Name()
	    fullNameField := hdlrv.FieldByName("fullName")
	    if fullNameField.IsValid() {
		    fullNameValue := fullNameField.String()
		    if fullNameValue != "" {
			    name = fullNameField.String()
		    }
	    }
    
	    ...
    }

Now gRPC clients developed by other frameworks or languages can access the gRPC services developed by go-micro.

Demo is here: https://github.com/bosima/go-demo/tree/main/go-micro-grpc-hello-compatible


asim:
Get @xpunch to review

cmd/protoc-gen-micro/plugin/micro/micro.go
```diff
@@ -137,6 +137,11 @@ func (g *micro) generateService(file *generator.FileDescriptor, service *pb.Serv
 	}
 	servName := generator.CamelCase(origServName)
 	servAlias := servName + "Service"
+	fullServName := servName
+	useGrpc := g.gen.Param["use_grpc"]
```
xpunch: useGrpc := g.gen.Param["use_grpc"] != ""
Do you think it's better to define userGrpc as bool?

xpunch, line 141: useGrpc := g.gen.Param["use_grpc"] != ""
Do you think it's better to define userGrpc as bool?

bosima: Originally, I thought I only needed this parameter, but it must be assigned when executing the command, so I wrote use_ grpc = 1, in fact, any value is OK.
However, since the assignment is made, it is better to have a clearer range of values. I'll modify it later.

xpunch: I mean define local variable as bool, like;
useGrpc := g.gen.Param["use_grpc"] != ""
No need to limit value range for use_grpc, but it's better to document it somewhere.

grissom1:
I'm not sure if I understand this correctly. But I was able to call go-micro gRPC server from Node.js client and Python client with no problem.
> At present, gRPC clients developed by other frameworks or languages cannot access the gRPC services developed by go-micro, because the gRPC method in go-micro does not contain the package name.

cmd/protoc-gen-micro/plugin/micro/micro.go
bosima: All right, I get it.

bosima:
> I'm not sure if I understand this correctly. But I was able to call go-micro gRPC server from Node.js client and Python client with no problem.
> 
> > At present, gRPC clients developed by other frameworks or languages cannot access the gRPC services developed by go-micro, because the gRPC method in go-micro does not contain the package name.

I tested the proxy generated by protoc-gen-go-grpc could not call GRPC services for Go-Micro.
Demo is :https://github.com/bosima/go-demo/tree/main/go-micro-grpc-hello

Is your proto file definition proto package?  Can you share  xxx.proto and xxx.pb.micro.go?

bosima:
> I'm not sure if I understand this correctly. But I was able to call go-micro gRPC server from Node.js client and Python client with no problem.
> 
> > At present, gRPC clients developed by other frameworks or languages cannot access the gRPC services developed by go-micro, because the gRPC method in go-micro does not contain the package name.

I will test it again.

asim:
The solution is a bit convoluted. I didn't really understand in the beginning but yes you can call grpc services from go-micro and go-micro services from grpc. The grpc client has a function that converts methods to grpc methods. But it requires that you are sending a request to the same name as the package when you initialise that client with NewXService. So maybe that's your issue? The fact that we don't include package names in the generated client can be an issue but you can call grpc services right now because of the way go-micro handles it.

Hopefully @xpunch and others can help resolve

bosima:
@asim @grissom1 You are right. 

At first, I  got a problem when go-micro client access grpc services developed by other languages, because package is not included in grpc method.  This problem has been solved by the last PR.

Then I failed to access the go-micro grpc server by grpc clients of other frameworks, so I tried to modify it. But now I can't repeat this problem. No problem now. Maybe I had a brain-fart.

@xpunch 
I read the code again, and found that my submitted modification was useless. 

in plugins/server/grpc/server.go:
the 'register' method still registers the service name without package.

    s := new(service)
    s.typ = reflect.TypeOf(rcvr)
    s.rcvr = reflect.ValueOf(rcvr)
    sname := reflect.Indirect(s.rcvr).Type().Name()
    ...
    server.serviceMap[s.name] = s

in plugins/server/grpc/grpc.go:
the 'handler' method erases the package.

    serviceName, methodName, err := mgrpc.ServiceMethod(fullMethod)
    service := g.rpc.serviceMap[serviceName]

xpunch:
So we can close this PR?

bosima:
> So we can close this PR?

useGrpc := g.gen.Param["use_grpc"] != ""
Does this need to be modified? The param is also included in the code that generated the client API in the last PR.

xpunch:
> > So we can close this PR?
> 
> useGrpc := g.gen.Param["use_grpc"] != "" Does this need to be modified? The param is also included in the code that generated the client API in the last PR.

If use_grpc is useless, should we remove it?

bosima:
> > > So we can close this PR?
> > 
> > 
> > useGrpc := g.gen.Param["use_grpc"] != "" Does this need to be modified? The param is also included in the code that generated the client API in the last PR.
> 
> If use_grpc is useless, should we remove it?

It is useful for the go micro grpc client to call grpc services developed by other frameworks.
asim also mentioned this issue. 

We also can close this PR.  It's just a matter of writing.

PR #2492, opened 2022-04-27
fix: consume and publish blocked after rabbitmq reconnecting

bosima:
As the title says, now consume and publish blocked after reconnecting. This can be replicated by restart rabbitmq.

For publish and consume , one go Channel may be blocked after the connection is disconnected.
For consume only, 'waitConnection' maybe the old value, and will blocked forever.

reference issue:
#https://github.com/rabbitmq/amqp091-go/issues/18

---
In addition 'amqp.Channel.NotifyReturn' is not a connection or channel failed, so i removed the 'channelNotifyReturn'.

[amqp-0-9-1-quickref](https://www.rabbitmq.com/amqp-0-9-1-quickref.html)
This method returns an undeliverable message that was published with the "immediate" flag set, or an unroutable message published with the "mandatory" flag set.


xpunch (approved):

xpunch:
Could you also update to github.com/go-micro/plugins/v4/broker/rabbitmq?

bosima:
> Could you also update to github.com/go-micro/plugins/v4/broker/rabbitmq?

ok

PR #2513, opened 2022-06-21
feat: redis implementation of cache

xpunch:
Update cache interface and add redis implementation.


asim:
I'm not a fan of prefixing Context as the first arg to all methods. I think its mostly just interface bloat. A better pattern is to have one method available in a package that returns an interface with the context set.

```
func Context(ctx context.Context) Cache {
   return &cache{
     ctx: ctx,
   }
}
```

This can then be used by the underlying implementation while keeping the interface clean

xpunch:
I'll change back to previous Context pattern, but do you think this will increase GC pressure in high concurrent scenario?

asim:
I think that's a non issue considering how much else in the system will first cause GC pressure. I think it's fine. Just leave it as is. It seems most people are stuck on this pattern even though I think it's pretty crap. Polluting interfaces with context isn't great.

xpunch:
OK, thanks for your explanation.

PR #2519, opened 2022-07-01
feat(web): add start and stop functions to the web

JellyTony:
feat(web): add start and stop functions to the web

@asim 
@xpunch

### DIFF
--- a/web/service.go
+++ b/web/service.go
@@ -34,6 +34,7 @@ type service struct {
 	running bool
 	static  bool
 	exit    chan chan error
+	ex      chan bool
 }
 
 func newService(opts ...Option) Service {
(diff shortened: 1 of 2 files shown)


asim:
So now we have Start/Stop/Run. Is that not a bit confusing? What am I supposed to use? Start or run? Would it just make more sense to have Run and Shutdown? What's the reason for this change?

JellyTony:
> So now we have Start/Stop/Run. Is that not a bit confusing? What am I supposed to use? Start or run? Would it just make more sense to have Run and Shutdown? What's the reason for this change?

Consistent with the implementation of the server, the run method is internally called to start and stop. If you want to manage the start and stop by yourself, you can directly use the start and stop methods, so that you can flexibly manage the life cycle

JellyTony:
> So now we have Start/Stop/Run. Is that not a bit confusing? What am I supposed to use? Start or run? Would it just make more sense to have Run and Shutdown? What's the reason for this change?

Run is easy to use, regardless of turning off signal monitoring. However, if I want to start and close myself, I don't need to use run mode, but directly use start and stop methods

PR #2522, opened 2022-07-02
api package

asim:
Making the api package work as an api gateway

- Enables creation of gateway through `api.NewApi()`
- Gateway runs on ":8080" by default
- Uses registry router by default meaning routes for any service on /[service]/[endpoint]
- Requires some level of testing

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -11,6 +11,8 @@ import (
 	"go-micro.dev/v4/server"
 )
 
+// The Api interface provides a way to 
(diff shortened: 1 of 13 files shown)


ivybridge-3c33:
@asim did you have any example api gateway  for grpc and rest api example rest? 
thank you very much

asim:
Not yet, still working on it. I think the interface/package needs to be rewritten again just to make it easier to use.

PR #2531, opened 2022-07-22
Fix http transport panic

Ak-Army:
Read the buffer before reset it in close method.


asim:
Can you please test your changes before submitting random PRs.

PR #2554, opened 2022-09-07
fix: log.Fatal() and log.Fatalf() will not exit with code 1

msxfXF:
log.Fatal() will not exit with code 1

If you execute the following code.
`
if err!=nil{
    log.Faltal(err)
}
`
The expected situation is that the program prints an error and exits. Otherwise, execution with errors may cause the program to crash. 

it seems to be missing `os.exit(1)` after logging.


asim:
Don't use this package. It's an internal deprecated package. Use top level logger. Is there any reason you're directly importing this?

msxfXF:
Oh sorry, I may have watched an outdated video course. 
It is true that it can be replaced by top level logger, Thanks.

PR #2555, opened 2022-09-19
Allow passing context to store operations

butonic:
This allows passing a context with opentelemetry tracing information to remote services / store implementations.

Regarding the pr template:
> 1. Make sure this PR targets the `develop` branch. We follow the git-flow branching model.

there is no develop branch ;-)

### DIFF
--- a/store/options.go
+++ b/store/options.go
@@ -65,6 +65,9 @@ func WithClient(c client.Client) Option {
 
 // ReadOptions configures an individual Read operation
 type ReadOptions struct {
+	// Context for the read
+	Context context.Context
+	// Read from the following
 	Database, Table string
 	// Prefix returns all records that are prefixed with key
 	Prefix bool
@@ -78,6 +81,13 @@ type ReadOptions struct {
 
 // ReadOption sets values in ReadOptions
 type ReadOption func(r *ReadOptions)
+
+// ReadContext to use for the read
+func ReadContext(context context.Context) ReadOption {
+	return func(r *ReadOptions) {
+		r.Context = context
+	}
+}
 
 // ReadFrom the database and table
 func ReadFrom(database, table string) ReadOption {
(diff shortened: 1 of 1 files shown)


butonic:
That failing test seems unrelated ... is it flaky?

Davincible:
Can you explain a bit more about your thoughts behind this PR?

Any core options to go-micro would still have to be implemented in each implementation of the store interface as found in `github.com/go-micro/plugins`, adding them here would mean adding lots of implementations.

For which specific implementation of the store interface (aka plugin) are you planning on using this with?

You can also add implementation-specific options to plugins, e.g. the nats store [plugin](https://github.com/go-micro/plugins/tree/main/v4/store/nats-js) has a few options specific to nats as lined out in the readme. I'm thinking it could be more appropriate to add these options to the particular implementation you plan on using this with.

Only options that are relevant to all or the majority of store implementations should be added to the core of go-micro.

About the tests, I'm not sure

xpunch (approved):

butonic:
I'd like to pass the opentelemetry traceid between services. The easiest way to do that is to use the context. And yes, every implementation will have to be changed to make use of that. For us redis and etcd are most important.

Even for the in memory implementation tracing could be added.

Unless I am blind on how to do tracing with go micro ...

Davincible:
@butonic etcd doesn't have a store implementation, so might be better for you to extend the plugins directly and add a context option for the redis and etcd plugin, and create a PR at `github.com/go-micro/plugins`.

Adding the options here would add tech debt for the other store plugin implementations. As this PR would just be used in 1/8 store plugins.

About whether this is the best way to do tracing, I can't say, haven't used it myself yet

xpunch:
I've added the open telemetry plugin to handling client/server tracing, you can take a look at https://github.com/go-micro/plugins/tree/main/v4/wrapper/trace/opentelemetry. This only handle rpc request and broker messages, if you want to pass trace id from micro service to store plugins, maybe add the operation context is the only way, but you should also implement the store plugins.
The tracing feature is not required by everyone who are using store plugins, maybe this feature should be optional?

Davincible:
@butonic recently we've started working on updating part of the plugin options, and they way loggers are handled. This could be taken into consideration with that as well, so we could update the plugins.

I'm not exactly sure however how you planned on implementing this. Could you create a PR for the Redis plugin with how you envisioned the changes?

Davincible:
Closing as it has become stale.

PR #2557, opened 2022-09-23
feat(config): add withFS option to file source 

asynxc:
This PR introduce a new option to the [file](https://github.com/go-micro/go-micro/tree/master/config/source/file) module that allow a user either to pass in a file path or FS handler. as argument.  (if both options provided, FS handler is used)

This options allow the use of a raw or embed.FS as data source. (e.g: as default config data, overridden by subsequent provided sources).

Any feedback is very welcome ^^

### DIFF
--- a/config/source/file/file.go
+++ b/config/source/file/file.go
@@ -3,12 +3,14 @@ package file
 
 import (
 	"io"
+	"io/fs"
 	"os"
 
 	"go-micro.dev/v4/config/source"
 )
 
 type file struct {
+	fs   fs.FS
 	path string
 	opts source.Options
 }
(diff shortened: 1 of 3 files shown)


Davincible (approved):

Davincible:
Thanks for the PR!

Just curious, what is your use case for this?

asynxc:
@Davincible Thanks ^^

working on a project that have a lot of configurations and we want it to have sane defaults. this require us to have a default configurations that gets merged/overridden by other (user) provided files...

with this PR, now we can embed a default config via embed.FS and use it as a source file, without the hustle of handling that in the project codebase.

PR #2559, opened 2022-09-24
feat(logger): add logger option to all micro components (override DefaultLogger) closes #2556

### LINKED ISSUE #2556: [FEATURE] allow passing custom logger to "server" component
**Is your feature request related to a problem? Please describe.**

in a context where we instantiate multiple [server](https://github.com/go-micro/go-micro/tree/master/server) in same process, we may need to assign specific fields to each instance.
So that it's easier to distinguish their logs.

**Describe the solution you'd like**

add new option `server.WithLogger(logger log.Logger)` to server options

If this's ok with you, i can create a PR in both [go-micro](https://github.com/go-micro/go-micro) and [plugins](https://github.com/go-micro/plugins) projects ?

asynxc:
closes #2556

(to get a better understanding of the reason bedind this PR, please checkout this issue [#2556](https://github.com/go-micro/go-micro/issues/2556)

A PR to allow providing a custom logger to every micro component (client, server, store, broker,...).

the guide lines of this implementation are :

- do not change components API (for retro compatibility)
- add Logger option to every component
- fallback to DefaultLogger if no logger option provided (for retro compatibility)

Any feedback is welcome ^^

### DIFF
--- a/service.go
+++ b/service.go
@@ -152,7 +152,7 @@ func (s *service) Run() (err error) {
 		}()
 	}
 
-	logger.Logf(log.ErrorLevel, "Starting [service] %s", s.Name())
+	logger.Logf(log.InfoLevel, "Starting [service] %s", s.Name())
 
 	if err = s.Start(); err != nil {
 		return err
--- a/service.go
+++ b/service.go
@@ -125,6 +125,8 @@ func (s *service) Stop() error {
 }
 
 func (s *service) Run() (err error) {
+	logger := s.opts.Logger
+
 	// exit when help flag is provided
 	for _, v := range os.Args[1:] {
 		if v == "-h" || v == "--help" {
@@ -150,9 +152,7 @@ func (s *service) Run() (err error) {
 		}()
 	}
 
-	if logger.V(logger.InfoLevel, logger.DefaultLogger) {
-		logger.Infof("Starting [service] %s", s.Name())
-	}
+	logger.Infof("Starting [service] %s", s.Name())
 
 	if err = s.Start(); err != nil {
 		return err
--- a/service.go
+++ b/service.go
@@ -7,7 +7,7 @@ import (
 	"sync"
 
 	"go-micro.dev/v4/client"
-	mlogger "go-micro.dev/v4/logger"
+	log "go-micro.dev/v4/logger"
 	"go-micro.dev/v4/server"
 	"go-micro.dev/v4/store"
 	"go-micro.dev/v4/util/cmd"
@@ -58,14 +58,14 @@ func (s *service) Init(opts ...Option) {
 			cmd.Store(&s.opts.Store),
 			cmd.Profile(&s.opts.Profile),
 		); err != nil {
-			s.opts.Logger.Log(mlogger.FatalLevel, err)
+			s.opts.Logger.Log(log.FatalLevel, err)
 		}
 
 		// Explicitly set the table name to the service name
 		name := s.opts.Cmd.App().Name
 		err := s.opts.Store.Init(store.Table(name))
 		if err != nil {
-			s.opts.Logger.Log(mlogger.FatalLevel, err)
+			s.opts.Logger.Log(log.FatalLevel, err)
 		}
 	})
 }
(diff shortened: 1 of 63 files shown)


Davincible:
@asynxc Really nice PR, thanks!

Two thoughts; 

In the log helper, you implemented the checks to see if a level is enabled, I know this was present in the original code, but this should be the responsibility of the logger implementation (plugin), so I think we can remove those checks.

Also, if I set the logger for the service globally, I also expect it to be set for all sub-modules/plugins unless specifically specified otherwise. I don't think this is currently implemented right?

asynxc:
> In the log helper, you implemented the checks to see if a level is enabled, I know this was present in the original code, but this should be the responsibility of the logger implementation (plugin), so I think we can remove those checks.

I agree with you, the original code does unnecessarily a double log level check (in the logger and in inline)
Also the logger.Helper does the log level check, but i dit not want to change its behavior/api since it's a public API...
 
the default logger [default](https://github.com/go-micro/go-micro/blob/master/logger/default.go#L97) and plugins logger does check log level, so i guess i can ditch using the Helper.  i have added a commit to do that, please take a look. @Davincible 

> Also, if I set the logger for the service globally, I also expect it to be set for all sub-modules/plugins unless specifically specified otherwise. I don't think this is currently implemented right?

no it's not the case at the moment, maybe can be done in a second iteration ^^
Edit: **Service** has no direct access to most components, so in order to pass in the service Logger, components must forward  it in to "child" components, and must be careful doing that to not override already setup"child" option logger... so for those reasons i don't think auto setup logger for all components using **Service** logger is good idea...
The user can always customize Service/global level logger by changing the **DefaultLogger**
@Davincible thoughts ?

service.go
```diff
@@ -145,14 +147,12 @@ func (s *service) Run() (err error) {
...
-				logger.Error(err)
+				logger.Log(log.ErrorLevel, err)
 			}
 		}()
 	}
 
-	if logger.V(logger.InfoLevel, logger.DefaultLogger) {
-		logger.Infof("Starting [service] %s", s.Name())
-	}
+	logger.Logf(log.ErrorLevel, "Starting [service] %s", s.Name())
```
Davincible: You accidentally changed info level to error level here, please double check them to make sure you didn't change the level

Davincible:
About the setup of the logger; the problem is that components can be initialized at different stages, which makes it difficult to reliably pass down the logger to all components, unfortunately.

The downside of this is that you might need to define it multiple times.

I'm afraid that passing in a logger will raise the expectation that it will be used everywhere, where this will only be used in the service.

```go
micro.NewService(
    micro.WithLogger(logger)
)
```

asynxc:
> I'm afraid that passing in a logger will raise the expectation that it will be used everywhere, where this will only be used in the service.

I agree with you, it's a valid assumption.
to accommodate this we can auto set `DefaultLogger` by the service logger set by the user via `micro.WithLogger` ?
this way the service logger will used everywhere...

@Davincible thoughts ?

Davincible:
The problem with just setting the global logger in the option is that there is no guarantee of order. The option can be executed before or after defaults are initialized. The only solution I see, is to make the logger in all services pointers, and by default point to the global variable, which we can then change at any point in time in the setup phase. If a user passes in a custom logger, then the logger wont be pointing to the global anymore

Davincible:
Turns out that's not how it works. I've been trying to implement it but can't come up with any consistent way. So for now the user will still have to manually set the `DefaultLogger`, at least you can now manually specify the loggers for the different interfaces

Davincible (approved):

Davincible:
@asynxc this is ready to be merged now. Have you been able to work on the plugins PR?

asynxc:
> @asynxc this is ready to be merged now. Have you been able to work on the plugins PR?

that's great !

I'm waiting for this PR to land and a go-micro release to be able to reuse the interfaces in plugins

Davincible:
@asynxc released!

Davincible:
> I'm waiting for this PR to land and a go-micro release to be able to reuse the interfaces in plugins

@asynxc  Any progress?

PR #2579, opened 2022-10-03
feat: add test framework & refactor RPC server

Davincible:


jochumdev:
Looks like everything is cleaner now, `server/rpc_server.go` changes are so big that they are hard to understand on the first sight.

Davincible:
> Looks like everything is cleaner now, `server/rpc_server.go` changes are so big that they are hard to understand on the first sight.

`rpc_server.go` is one massive file with massive functions. Trying to break it up to make it more digestable.

Davincible:
todo: don't retry on business logic error
Edit: done

PR #2584, opened 2022-10-20
fix: store table initialization

### LINKED ISSUE #2583: Store table initialization
[This code ](https://github.com/go-micro/go-micro/blob/c25dee7c8a7053743266c95b6a63a1b11ef3c1fb/service.go#L64) redefines store table provided via --store_table or $MICRO_STORE_TABLE for no reason.
And forces us to re-init store table later once again.

I think the previous approach (before [the commit ](https://github.com/go-micro/go-micro/commit/71538adfdcb00cb8e240c3b57471c55e74756e82)) was a correct one. What was the reason for this change?

sgnrslv:
Set the table name to the service name only if table name is empty.

closes #2583

### DIFF
--- a/service.go
+++ b/service.go
@@ -61,11 +61,14 @@ func (s *service) Init(opts ...Option) {
 			s.opts.Logger.Log(log.FatalLevel, err)
 		}
 
-		// Explicitly set the table name to the service name
-		name := s.opts.Cmd.App().Name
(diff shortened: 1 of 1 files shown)


Davincible (approved):

Davincible:
Thanks for the PR!

PR #2592, opened 2022-11-09
feat(api): add extra form content type and extract endpoints from path

leoujz:
api gateway:
- handing http request adds Content-Type application/x-www-form-urlencoded support
- extract endpoints from path if no endpoint matched, only in api case

### DIFF
--- a/api/handler/rpc/rpc.go
+++ b/api/handler/rpc/rpc.go
@@ -315,7 +315,7 @@ func requestPayload(r *http.Request) ([]byte, error) {
 		}
 
 		return raw.Marshal()
-	case strings.Contains(myCt, "application/www-x-form-urlencoded"):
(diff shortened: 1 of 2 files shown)


xpunch (approved):

Davincible:
Nice!

Could you explain a bit how the endpoint extraction works?

If I'm not mistaken it now limits you to only have one slash in your URL for your method

leoujz:
The original code use rsp.Method as the endpoint name. As the strategy mentioned in https://github.com/go-micro/api, I extract the endpoint name from rsp.Path. If any endpoint matched before this code, this would never happen, this only runs when no endpoint mathced, as a final default value. when using rsp.Method, like POST, GET, would never work.

leoujz:
And 3 failing checks relative to my code? I don't know how to fix that

Davincible:
Makes sense, thanks for the PR!

Checks are not relative to your change.

PR #2600, opened 2022-11-24
fix(registry/cache): start watcher loop for other services

Scukerman:
The registry watcher loop starts only for the first service and does not for others.

Now it starts for all services and there are no "connection refused" errors when dependent services are rolling out.


Davincible:
Nice, thanks for the PR.

PR #2624, opened 2023-04-03
fix(api/handler): avoid zombie goroutine when connection unexpectedly closes

tchandelle:
In API handler, when serving stream through web-socket, the `writeLoop` goroutine may never finishes if the websocket connection closes unexpectedly.

Solution: Pass a cancellable context when creating a new stream.


Davincible:
Nice solution, thanks!

PR #2628, opened 2023-04-12
fix(config/source/cli): mergo.Map error, src and dst must be of same …

slowrookie:
config/source/cli error

read config/srouce/cli return errr (src and dst must be of same type error).
use the Merge function with the WithOverride option to merge two structs or maps with different types .


Davincible (approved):

Davincible:
Nice catch, thanks for PR

PR #2630, opened 2023-04-12
fix(config): fix file source watcher stop behavior when Stop is called

### LINKED ISSUE #2629: [BUG] config file source watcher do not stop correctly when Stop() is called
## Describe the bug
when calling Stop() on file source watcher, the watcher do not exit immediately and do not return the correct error `source.ErrWatcherStopped`

1. What are you trying to do?
stop the file source watcher on demand by calling Stop() method

3. What did you expect to happen?
Next() method of the watcher instance return `source.ErrWatcherStopped`

5. What happens instead?
the underlying fs watcher (fsnotify) close Events & Errors channels when Close() is called, but we don't handle it correctly

asynxc:
This PR should fix a minor bug in config source file watcher as described in the Issue #2629 

closes #2629

any feedback is very welcome

### DIFF
--- a/config/source/file/watcher.go
+++ b/config/source/file/watcher.go
@@ -13,8 +13,7 @@ import (
 type watcher struct {
 	f *file
 
-	fw   *fsnotify.Watcher
-	exit chan bool
+	fw *fsnotify.Watcher
 }
 
 func newWatcher(f *file) (source.Watcher, error) {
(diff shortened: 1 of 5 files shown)


Davincible (approved):

Davincible:
Hey! looks good, added tests are nice. Thanks for the work

PR #2659, opened 2023-10-04
bump urfave cli & fix race condition

dragonchaser:
We are running into race conditions when running our software. Those seem to originate in the urfave-cli you are using, as a first step (to tackle them) this PR bumps urfave/cli from the (ancient) 2.3.0 to the current 2.25.7

### DIFF
--- a/config/default.go
+++ b/config/default.go
@@ -2,14 +2,14 @@ package config
 
 import (
 	"bytes"
-	"sync"
-	"time"
-
+	"fmt"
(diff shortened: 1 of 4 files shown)


dragonchaser:
Looking into necessary follow-up work :eyes:

dragonchaser:
@asim this should fix the issue (for now) any follow up work from my side needed?

asim (approved):

PR #2673, opened 2023-11-19
util/addr: Fixes findIP to return the correct public IP

guillaumebour:
When run on a host with no private IP, but with a public IP available, `go-micro` incorrectly figures out its service's address and uses the loopback address. This is due to the `findIP` function in `utils/addr` returning the wrong IP address when run on a host with a Public IP and no Private IP, leading to `go-micro` advertising the wrong address for its service (e.g. when using `etcd` as the registry).

## What was done

- Update `util/addr_test.go` with a test for `findIP`. 
- Update `findIP` to return the first public IP (if available) if no private IP is available.
- Update `findIP` to return the first loopback IP (if available) if no public IP is available.
- Fix typos in the `Extract` function documentation.

## How to reproduce the initial issue

To reproduce the initial issue, you can setup a docker-compose with `etcd` and a `go-micro` service and check the information available in `etcd`.

<details><summary>Details</summary>
<p>
<br/>

**Step 1.** Create a `go-micro` app with the following `main.go`:

```go
package main

import (
	"context"
	"fmt"
	"go-micro.dev/v4"
	"os"

        _ "github.com/go-micro/plugins/v4/registry/etcd"
)

type Request struct {
	Name string `json:"name"`
}

type Response struct {
	Message string `json:"message"`
}

type Helloworld struct{}

func (h *Helloworld) Greeting(ctx context.Context, req *Request, rsp *Response) error {
	rsp.Message = "Hello " + req.Name
	return nil
}

func main() {
	serviceName := os.Getenv("SERVICE_NAME")

	fmt.Printf("--- Starting %s ---\n", serviceName)

	// create a new service
	service := micro.NewService(
		micro.Name(serviceName),
		micro.Handle(new(Helloworld)),
	)

	// initialise flags
	service.Init()

	// start the service
	err := service.Run()
	if err != nil {
		return
	}
}

```

**Step 2.** Use the following Dockerfile to build the application's image with the tag `helloworld:latest`:

```Dockerfile
# syntax=docker/dockerfile:1
FROM golang:1.20 as builder

WORKDIR /app

COPY go.mod go.sum ./
RUN go mod download

COPY . .

RUN CGO_ENABLED=0 GOOS=linux go build -o helloworld

FROM debian

WORKDIR /app

COPY --from=builder /app/helloworld /bin/helloworld

CMD ["/bin/helloworld"]
```

**Step 3.** Use the following `docker-compose.yaml` to deploy `etcd` alongside the app:

```yaml
version: "3"
services:
  etcd-test:
    container_name: etcd-test
    image: bitnami/etcd:latest
    environment:
      - ALLOW_NONE_AUTHENTICATION=yes
    networks:
      test_net:
        ipv4_address: 109.0.0.8
  helloworld1:
    container_name: helloworld
    image: helloworld:latest
    environment:
      - SERVICE_NAME=helloworld-1
      - MICRO_SERVER_ADDRESS=:8080
      - MICRO_REGISTRY=etcd
      - MICRO_REGISTRY_ADDRESS=etcd-test:2379
    depends_on:
      - etcd-test
    networks:
      test_net:
        ipv4_address: 109.0.0.10
networks:
  test_net:
    ipam:
      driver: default
      config:
        - subnet: 109.0.0.0/24
```

**Step 4.** Use the following command to get the address advertised by the service:

`docker exec -it etcd-test "/opt/bitnami/etcd/bin/etcdctl" get "" --prefix | sed -n '2 p' | jq '.nodes | first.address'`

With the docker network setup with a public IP range, it will be: `"127.0.0.1:8080"`, while without, it will use the container private IP. 

</p>
</details>

### DIFF
--- a/util/addr/addr.go
+++ b/util/addr/addr.go
@@ -35,8 +35,8 @@ func IsLocal(addr string) bool {
 }
 
 // Extract returns a valid IP address. If the address provided is a valid
-// address, it will be returned directly. Otherwise the available interfaces
-// be itterated over to find an IP address, prefferably private.
+// address, it will be returned directly. Otherwise, the available interfaces
+// will be iterated over to find an IP address, preferably private.
 func Extract(addr string) (string, error) {
 	// if addr is already specified then it's directly returned
 	if len(addr) > 0 && (addr != "0.0.0.0" && addr != "[::]" && addr != "::") {
@@ -115,10 +115,12 @@ func IPs() []string {
 	return ipAddrs
 }
 
-// findIP will return the first private IP available in the list,
-// if no private IP is available it will return a public IP if present.
+// findIP will return the first private IP available in the list.
+// If no private IP is available it will return the first public IP, if present.
+// If no public IP is available, it will return the first loopback IP, if present.
 func findIP(addresses []net.Addr) (net.IP, error) {
 	var publicIP net.IP
+	var localIP net.IP
 
 	for _, rawAddr := range addresses {
 		var ip net.IP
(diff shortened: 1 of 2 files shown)


jochumdev:
Hi Guillaume,

This is great work! Many thanks.

Kind regards,
René

jochumdev (approved):

PR #2736, opened 2024-10-28
invalidate service if node was not updated

butonic:
[nats does not yet emit events when a key is evicted due to ttl](https://github.com/nats-io/nats-server/issues/3268). This might cause the cache to never receive an update for the node. This PR will invalidtate a service if any node has not been updated in time.

Let me know what you think and why node TTL was not taken into account, yet?

### DIFF
--- a/registry/cache/cache.go
+++ b/registry/cache/cache.go
@@ -42,6 +42,7 @@ type cache struct {
 	sg      singleflight.Group
 	cache   map[string][]*registry.Service
 	ttls    map[string]time.Time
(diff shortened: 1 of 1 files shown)


asim:
I think there are ttls per service but not node. I have a feeling it comes down to. There's a lot of edge cases to deal with and caching can be quite hard to do properly. So the idea is that you don't want to just throw away a node in case there's some sort of upstream issue with the registry. I'm not really maintaining this anymore so I will accept your PR if you think it works

kobergj (approved):

asim (approved):

PR #2740, opened 2024-11-15
update ttl for updated nodes only

butonic:
Followup to https://github.com/micro/go-micro/pull/2736

obviously, you were right: ~~edge cases are~~ hard to do properly 😞


kobergj (approved):

asim:
> Followup to #2736
> 
> obviously, you were right: ~edge cases are~ hard to do properly 😞

Yea no worries it takes some attempts and seeing it in action. Eventually I found simplicity beat out trying to do clever things because the complexity of maintenance became a burden and I was not in the environments it was being used. If you have first hand exposure to the usage then you can tweak it in the real world and everyone else will benefit.

PR #2761, opened 2025-05-15
Re-add events package

asim:


bketelsen:
@asim let me know what you think of my changes.

PR #2767, opened 2025-05-16
refactor: move MDNS registry to root of registry package

BombartSimon:
- Removed the mdns subpackage and integrated its logic directly into the root of the registry package.

- Updated the default registry to use the refactored MDNS implementation.

- Refactored tests accordingly.


bketelsen (approved):
Perfect, thank you!

PR #2772, opened 2025-05-19
feat: re-add profiles

bketelsen:
One-shot code generation from Code with Copilot.

This isn't complete, but checking in to show what happened.


bketelsen:
same old test flake, merging

PR #2773, opened 2025-05-20
Create docs readme

asim:


asim:
Doing it in https://github.com/go-micro/docs

PR #2786, opened 2025-06-04
Topic wildcard

Ak-Army:
Ability to use filter subject in topic name e.g. factory-events.*.*


asim:
I merged your previous PR so you now have a merge conflict here

Ak-Army:
Thanks, I will resolved it

Ak-Army:
Resolved.

asim (approved):

asim:
Weirdly no tests are being run for this. Are you taking into account the need for grpc server changes since you've changed the router interface?

Ak-Army:
I admit I never looked at the grpc part because we don't use it. And since there is no test for it, I didn't touch it, because we can't test it with our own system.
But if it's enough that if the ProcessMessage function is adapted to it, I'll be happy to commit the patch.

asim:
OK well I think firstly you probably need to resync master and squash all this because theres like commits from 4 years ago, and I need to see the tests run and you'll see grpc server fail because of the change in interface.

PR #2892, opened 2026-03-31
server/grpc: improve graceful stop behavior

jejefferson:
## Summary

  This changes the gRPC server shutdown path to behave more like a real graceful stop for long-lived RPCs and streams.

  Current behavior in `server/grpc/grpc.go` is:

  1. wait on internal `wg`
  2. only then call `grpc.Server.GracefulStop()`
  3. force `Stop()` after `1s`

  That makes graceful shutdown difficult for services with long-running gRPC calls.

  ## What this patch changes

  - starts `grpc.Server.GracefulStop()` immediately when shutdown begins
  - adds a configurable graceful timeout via `GracefulStopTimeout(time.Duration)`
  - only force-calls `Stop()` after that timeout
  - waits on the internal waitgroup after transport shutdown has started, not before

  ## Why

  For gRPC graceful shutdown, the expected behavior is:

  - stop accepting new RPCs as soon as shutdown begins
  - allow already-running RPCs/streams to continue
  - only force-stop after an explicit timeout

  With the previous order, `GracefulStop()` starts too late and the hardcoded `1s` fallback is too aggressive for long-lived calls.

  ## Compatibility

  - default behavior is preserved if `GracefulStopTimeout(...)` is not configured
  - this is a minimal transport-level change; it does not change service lifecycle APIs

  ## Validation

  I verified the package still compiles with:

  ```bash
  go test ./server/grpc -run '^$'

### DIFF
--- a/examples/graceful-stop/README.md
+++ b/examples/graceful-stop/README.md
@@ -1,0 +1,38 @@
+# Graceful Stop Demo
+
+This example demonstrates the intended shutdown behavior after the gRPC graceful-stop patch.
+
+## Run
+
+```bash
+go run ./examples/graceful-stop
+```
+
+## Expected behavior
+
+- one long-running RPC starts
+- shutdown begins while that RPC is still running
+- new RPCs stop being accepted shortly after shutdown starts
+- the in-flight RPC is allowed to finish
+
+Typical output:
+
+```text
+long RPC is running; starting shutdown
+new RPC rejected after shutdown began: ...
+long RPC completed: slept for 1500ms
+done
+```
+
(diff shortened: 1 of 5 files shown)


Grantmartin2002:
+1 on this. I independently ran into the same issue and submitted a similar fix to the plugins repo [(go-micro/plugins#163) ](https://github.com/go-micro/plugins/pull/163) before learning the plugins were merged here. The hardcoded 1s timeout was causing ECONNRESET errors during rolling updates for our long-lived streaming RPCs.

asim (approved):

PR #3006, opened 2026-06-24
Add provider-conformance harness, contract test, and reframe docs to 'agent harness'

asim:
### Motivation

- Provide a repeatable cross-provider conformance check so the same end-to-end harnesses can be exercised against every configured live model provider without silently burning API credits. 
- Make the documented 0→1 getting-started path impossible to accidentally break by locking the `micro new` output with an automated contract test. 
- Clarify product messaging and internal strategy by repositioning the project as an "agent harness" and capturing the roadmap/assessment in docs.

### Description

- Add a new provider conformance harness at `internal/harness/provider-conformance/main.go` that runs specified harnesses (`universe`, `agent-flow`, `plan-delegate`) across a configurable list of providers and skips providers without API keys. 
- Update CI workflow `.github/workflows/universe.yml` to call `internal/harness/provider-conformance` for scheduled/manual runs and export common provider env vars so configured providers are checked; the previous Atlas-only live check was replaced. 
- Add a `provider-conformance` target to the `Makefile` and document `harness`/`provider-conformance` usage, and wire the deterministic harness targets to `make harness`. 
- Add an automated contract test `cmd/micro/cli/new/contract_test.go` to validate `micro new helloworld` produces a buildable Go service and update many docs/website files and `internal/docs/DEVELOPMENT_STRATEGY_ASSESSMENT.md` to adopt the "agent harness" framing and guidance.

### Testing

- Ran unit tests with `go test ./...`, including the new `TestZeroToOneContract`, and they completed successfully. 
- Executed the deterministic harnesses with `make harness` (which runs `go run ./internal/harness/universe`, `agent-flow`, and `plan-delegate`) and observed they passed locally. 
- Verified the provider conformance command runs deterministically against the `mock` provider via `go run ./internal/harness/provider-conformance -providers mock` and succeeds, while live provider runs are skipped in environments lacking API keys (CI scheduled runs will skip missing keys and fail if a configured provider fails).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3b8b661588833088e906687ec394b5)

### DIFF
--- a/.github/workflows/harness.yml
+++ b/.github/workflows/harness.yml
@@ -1,10 +1,9 @@
-name: Universe (E2E)
+name: Harness (E2E)
 
-# Spins up a mini go-micro world end-to-end — services, a durable flow
-# that crashes and resumes, and an agent — then shuts it down. Everything
-# is real except the LLM, which is mocked, so the default job is
-# deterministic and needs no secrets. A second job runs the same scenario
-# against a live model when an API key secret is configured.
+# Runs the end-to-end harnesses for agents, services, flows, and provider
+# conformance. The default job uses deterministic mock LLMs and needs no
+# secrets. A second job runs the same harnesses against any live providers
+# whose API key secrets are configured.
 
 on:
   push:
@@ -16,8 +15,8 @@ on:
   workflow_dispatch:
 
 jobs:
-  universe:
-    name: Mini universe (mock LLM)
+  harness:
+    name: Harnesses (mock LLM)
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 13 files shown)


asim:
Review (from a second pass with Claude Code). Verified all three harnesses (`universe`, `agent-flow`, `plan-delegate`) accept `-provider` and handle `mock`, so the conformance runner is wired correctly. Lint/mock-universe/etcd are green.

**Good:** the conformance harness (skip-on-missing-key, per-run timeout, runs all and reports every failure before exiting non-zero) is solid; the "lock the `micro new` path" contract test is a valuable idea; and the docs thoughtfully reconcile the old "no harness" language with the new positive framing.

**Mechanical fixes pushed as #3007** (stacked on this branch):
- `provider-conformance`: `go run` launches the harness as a child it doesn't kill on context cancellation, so a timed-out harness (which starts local services) can be orphaned. #3007 builds each harness to a temp binary and execs that, so the timeout is honest.
- `contract_test.go`: it shells out to `go mod tidy` (via `Run`) + compiles a generated service, so it makes the default `go test ./...` slow and network-dependent (likely why Unit Tests runs long). #3007 skips it under `-short` and uses `go build ./...` (the contract is that it *builds*).

**Left for you to decide (judgment, not mechanical):**
- **Scope** — this bundles the conformance harness, the contract test, and a broad **"agent harness" rebrand** (README headline, landing hero, `<title>`, docs, comparison, + a 194-line strategy doc). I'd split the rebrand into its own PR you sign off on explicitly — it's a positioning change that flips a term the docs previously used negatively, and it shouldn't ride in on test infra.
- Minor: `localRPCEnv` sets `NO_PROXY=*`, which also bypasses the proxy for provider API calls (fine on CI runners, would bite behind a corporate proxy). And the strategy doc quotes the pre-rebrand README line as the current positioning.

No correctness blockers — land the conformance + contract test (with #3007), and consider pulling the rebrand out.

---
_Reviewed with Claude Code._

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #3154, opened 2026-06-27
fix: retract phantom modules for cmd paths

### LINKED ISSUE #2987: [BUG] go install @latest
## Long story short
First, the module name in the go.mod file must be go-micro.dev/v6. Second, create a new branch named "v6" and add the "v6.x.x+1" tag to this branch. The next step is to run "go install go-micro.dev/v6/cmd/micro@latest" and ensure that the documentation has been parsed at https://pkg.go.dev/go-micro.dev/v6.

Fix #2812, #2838, #2796, and #2985

## Battle Proof
- Redirect url with meta headers: https://micro.labqa.pp.ua/v5/
- Go Packages: https://pkg.go.dev/micro.labqa.pp.ua/v5
- go install micro.labqa.pp.ua/v5/cmd/micro@latest

alex-dna-tech:
## Summary
Retracts phantom module versions for `cmd`, `cmd/micro`, and `cmd/protoc-gen-micro`. 

These paths were incorrectly cached as separate modules by the Go proxy due to a previous vanity domain misconfiguration. By publishing a version that retracts all existing versions, the proxy will stop treating them as module roots, forcing `go install ...@latest` to correctly fall back to the root module `go-micro.dev/v6`.

Fixes #2987

## Test plan
- Run `go list -m -retracted go-micro.dev/v6/cmd@latest` (after publishing tags) to verify retractions are active.
- Verify `go install go-micro.dev/v6/cmd/micro@latest` resolves to the root module.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cmd/go.mod
+++ b/cmd/go.mod
@@ -1,6 +1,0 @@
-module go-micro.dev/v6/cmd
-
-go 1.24
-
-// wipe out phantom module cache
-retract [v0.0.0, v1.18.1]
--- a/cmd/go.mod
+++ b/cmd/go.mod
@@ -1,0 +1,6 @@
+module go-micro.dev/v6/cmd
+
+go 1.24
+
+// wipe out phantom module cache
+retract [v0.0.0, v1.18.1]
--- a/cmd/micro/go.mod
+++ b/cmd/micro/go.mod
@@ -1,6 +1,0 @@
-module go-micro.dev/v6/cmd/micro
-
-go 1.24
-
-// wipe out phantom module cache
-retract [v0.0.0, v1.18.1]
--- a/cmd/micro/go.mod
+++ b/cmd/micro/go.mod
@@ -1,0 +1,6 @@
+module go-micro.dev/v6/cmd/micro
+
+go 1.24
+
+// wipe out phantom module cache
+retract [v0.0.0, v1.18.1]


alex-dna-tech:
The attempt to land `go.mod` files in this PR broke the build because Go treats any directory with a `go.mod` as a separate module, breaking internal imports.

To fix the proxy cache without breaking `master`, we need a "tag-only" retraction. Because the proxy treats `cmd` and `cmd/micro` as separate phantom modules, each requires its own tag:

1. **Temporarily add** the `go.mod` files with the `retract` directives.
2. **Create specific tags** for the phantom paths:
   - `git tag cmd/v1.18.1`
   - `git tag cmd/micro/v1.18.1`
3. **Push the tags**: `git push origin --tags`
4. **Delete the `go.mod` files** immediately.

The Go proxy will process the retractions from the tags, forcing `go install ...@latest` to fall back to the root module while keeping the codebase healthy.

alex-dna-tech:
As mentioned in golang/go#80100, we need to use the  [retract derective](https://go.dev/ref/mod#go-mod-file-retract). Can we use this weekend to temporarily break the micro-cli build and cure the "poisoning" issue?

asim:
I've merged this thinking I need to do that and then retract but I'm not sure do the tags but I think from your instructions its not quite right as I was supposed to put the go mod with retracts in, then tag and then delete

alex-dna-tech:
Yeh, I will split to separate pull requests. But you should add tags after merge first one that brake build.

PR #3166, opened 2026-06-27
fix: remove temporary go.mod for phantom module retraction

alex-dna-tech:
## Summary
Removes the temporary `go.mod` files added in the previous PR. This restores the build by ensuring these paths are once again part of the root module.

This should be merged immediately after the tags are pushed and the previous PR is landed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cmd/go.mod
+++ b/cmd/go.mod
@@ -1,6 +1,0 @@
-module go-micro.dev/v6/cmd
-
-go 1.24
-
-// wipe out phantom module cache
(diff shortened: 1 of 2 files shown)


asim:
@alex-dna-tech I used Claude to do the work. Want to confirm it works.

alex-dna-tech:
Oh. I missed the added tags. Will try to install to trigger process.

PR #3651, opened 2026-07-02
docs(cli): fix loop init flag example

asim:
## Summary
- Fix the micro loop init README example so the multiline shell command remains copy-pasteable.
- Move inline flag explanations into a Markdown list below the command.

## Testing
- go build ./...
- go test ./cmd/micro/loop/...
- golangci-lint run ./...
- go test ./... (fails in existing environment-dependent packages: atlascloud stream conformance and loopback-forbidden gRPC transport/client tests)


asim:
Closing as a stale duplicate. This branch was cut from `master` **before #3649 merged**, so alongside the intended README fix it re-adds the entire `micro loop` feature (`cmd/micro/loop/loop.go`, the templates, the `main.go` import, the CHANGELOG entry) — all of which already landed in #3649. That overlap is the merge conflict (`mergeable_state: dirty`): the 8 duplicated files collide with what's already on `master`.

The one piece of genuine new value here is the **README readability fix** — the `micro loop init` example currently has `\` line-continuations followed by inline `# comments`, which isn't copy-pasteable; moving the flag explanations into a list below a clean command is the right call. I'm carrying that fix forward into the in-flight PR that reworks `micro loop` (prompt-file-driven roles + `coherence`/`release`), which rewrites that README section anyway. So nothing is lost — just landing it without the duplicate-file conflict.

Thanks Codex.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #3769, opened 2026-07-03
feat: add MiniMax provider

octo-patch:
## Summary
Add MiniMax as a new LLM provider, compatible with the existing OpenAI-compatible provider architecture.

## Changes
- Add `ai/minimax/minimax.go` implementing `ai.Model` (mirrors the `groq` provider — self-contained OpenAI-compatible chat completions + tool calls, streaming delegated to `ai/internal/openaiapi`)
- Register `minimax` in `init()` with `ai.Register` + `ai.RegisterStream`; default model `MiniMax-M3`, default base URL `https://api.minimax.io`
- Add `minimax` case to `AutoDetectProvider`
- Add unit tests in `ai/minimax/minimax_test.go` (mirrors `groq_test.go`)
- Document the provider in `ai/README.md` and the root README provider table

## Testing
- `go test ./ai/...` passes
- `go vet ./ai/minimax/ ./ai/` clean

## Note
Diff is slightly over 300 lines because the new provider and its test file are near-verbatim clones of the existing `groq` provider, keeping the integration consistent with the sibling OpenAI-compatible providers.


asim:
Thanks for the contribution

PR #4856, opened 2026-07-16
Add retract-phantom.sh for proxy cache retraction of phantom module paths

alex-dna-tech:
## Summary

Adds `retract-phantom.sh` — a bash script that retracts phantom module paths cached by the Go proxy that don't exist on master.

## Problem

The Go proxy has cached module paths like `go-micro.dev/v5/api/handler`, `go-micro.dev/v4/a`, etc. as separate modules. When users run `go install go-micro.dev/v5/api/handler@latest`, they get:

```
go: go-micro.dev/v5/api/handler@latest: version constraints conflict:
    go-micro.dev/v5/api/handler@v1.18.0: parsing go.mod:
    module declares its path as: github.com/micro/go-micro
            but was required as: go-micro.dev/v5/api/handler
```

## Solution

Tag-only retraction using orphan commits:
1. Creates empty orphan commits with `retract [v0.0.0, v1.18.2]` directives
2. Tags each phantom path (e.g., `api/handler/v1.18.2`)
3. Pushes tags to origin — master is never touched

After retraction, `go install ...@latest` falls back to the root module.

## Phantom Paths Covered

| Base | Paths |
|------|-------|
| v6 | api |
| v5 | api/api, api/gateway, api/genproto, api/handler, api/handler/http |
| v4 | a, a/codec/proto, a/transport/memory |

## Usage

  ./retract-phantom.sh [--dry-run] [--pilot] [--push]

|Flag | What it does|
|------|-------|
|(none) | Create all 10,211 phantom tags locally (no push).|
|--dry-run | Print what would happen; touches nothing.|
|--pilot | Process only the first path per base (3 tags total). Validate the mechanism cheaply.|
|--push | Create tags and push them to origin.|

  Typical flow:
  ./retract-phantom.sh --dry-run          # sanity-check the enumeration
  ./retract-phantom.sh --pilot --push     # tiny live test (3 tags)
  ./retract-phantom.sh --push             # full run

  What you need to know

  - Idempotent. Re-running skips already-existing tags, so a partial/failed run can just be
  re-run.
  - Invalid names are skipped. A path whose tag name fails git check-ref-format (e.g. the old
  protoc-gen-micro~) is logged and skipped — it can never be retracted by this mechanism.
  - Push target is origin. Right now origin is the alex-dna-tech fork, so the tags have no effect
   on the go-micro.dev proxy (that reads micro/go-micro/upstream). To actually retract on the
  live proxy, point the push at upstream — either git remote set-url origin <upstream> first, or
[...]

PR #4873, opened 2026-08-01
fix(retract): dedupe phantom paths and tag refs before push

alex-dna-tech:
## What

Follow-up to #4856 ("Add retract-phantom.sh for proxy cache retraction of phantom module paths").

`retract-phantom.sh` failed its `--push` with:

```
error: dst ref refs/tags/<rel>/v1.18.2 receives from more than one src
```

Duplicate entries in `PHANTOM_PATHS` and cross-base path collisions (e.g. `go-micro.dev/v4/client` vs `go-micro.dev/v5/client`) map to the same tag `<rel>/v1.18.2`, so the single `git push` invocation passed the same destination ref multiple times.

## Change

- Dedupe each per-base path array (`BASES_V6/V5/V4/V0`, first-seen order), cutting the list from ~21k to 11,064 unique paths.
- Dedupe `CREATED_TAGS` with `sort -u` before the push so git never sees the same dst ref twice.
- Base execution order remains v6 -> v5 -> v4 -> v0 (high to low), so on a cross-base tag collision the highest-version base's go.mod module name wins.

## Testing

`bash -n retract-phantom.sh` passes; `./retract-phantom.sh --push` exits 0 and publishes 10,311 tags (267 new, rest already up-to-date) with zero "receives from more than one src" / rejected refs. Verified on origin via `git ls-remote`.

### DIFF
--- a/retract-phantom.sh
+++ b/retract-phantom.sh
@@ -21310,6 +21310,11 @@ for p in "${PHANTOM_PATHS[@]}"; do
     esac
 done
 
+for b in V6 V5 V4 V0; do
+    declare -n ba="BASES_$b"
+    mapfile -t ba < <(printf '%s\n' "${ba[@]}" | awk '!seen[$0]++')
+done
+
 # ── Sanity checks ───────────────────────────────────────────────────────────
 for p in "${PHANTOM_PATHS[@]}"; do
     case "$p" in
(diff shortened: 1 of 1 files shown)


alex-dna-tech:
@asim The code didn't trigger the last “full run”; I optimized the script so that it runs less frequently and works faster. Try again

asim:
@codex try do full run with this script again on this branch and if it works we'll merge

PR #4878, opened 2026-08-04
fix: preserve AtlasCloud conformance markers across text tool fallbacks

### LINKED ISSUE #4729: Fix AtlasCloud agent conformance marker handling
Root cause:
- CI run https://github.com/micro/go-micro/actions/runs/29163774031 failed only in Harness (E2E), specifically `atlascloud / agent (model call + tool call)`.
- The live agent provider conformance test failed in `go-micro.dev/v6/agent`: `TestAgentProviderConformanceMatrix/atlascloud` reported `Ask: provider conformance incomplete after 4 attempts: missing conformance marker`.
- Other AtlasCloud harnesses in the same run passed (`universe`, `agent-flow`, `plan-delegate`, `a2a-stream-fallback`), and seven other providers were skipped because their API keys were not configured. This points to a scoped AtlasCloud basic agent/tool-call conformance regression or harness prompt/response parsing gap rather than a broad workflow failure.
- Dedupe check: searched open `codex` issues for AtlasCloud/conformance/provider/marker and found only #4713, which covers mock plan-delegate resume scenarios and does not cover the failing basic agent conformance marker path.

Where:
- Workflow run: https://github.com/micro/go-micro/actions/runs/29163774031
- Job: Provider harnesses (live LLM conformance) / Provider conformance against live models
- Harness: `atlascloud / agent (model call + tool call)`
- Test: `TestAgentProviderConformanceMatrix/atlascloud`

Acceptance criteria:
- The AtlasCloud `agent` provider-conformance harness reliably obtains the expected conformance marker within the existing retry budget, or the harness/provider adapter is adjusted so valid AtlasCloud tool-call responses are normalized and recognized.
- Add or update deterministic coverage where possible for the marker extraction/normalization path so malformed or missing markers fail clearly without requiring live credentials.
- A rerun of `go run ./internal/harness/provider-conformance -providers atlascloud -harnesses agent -summary-json provider-conformance-summary.json -summary-markdown provider-conformance-summary.md -capabilities-markdown provider-capabilities.md` passes with AtlasCloud credentials configured.
- `go build ./...`, `go test ./...`, and `golangci-lint run ./...` pass.

mvanhorn:
## Why

The live AtlasCloud agent conformance harness keeps failing with a missing conformance marker, even though the structured echo and the guarded delegate both run.

In `ai/atlascloud/atlascloud.go`, a follow-up response that carries a text-encoded tool call is correctly kept in `Reply` so the agent's text-fallback parser can find it. That branch does not, however, retain the results of structured tools the provider already executed. When AtlasCloud emits `conformance_echo` as a structured call and the `delegate` as tagged text, the echo result — the one containing `agent-conformance-ok` — is dropped at the structured-to-text transition, before `agent.askLocked` combines the reply, answer, and additional text-tool result. All four attempts then exhaust on a marker that was produced and thrown away.

Fixes #4729

## What changed

`ai/atlascloud/atlascloud.go`: when the follow-up response is kept for text-fallback parsing and structured tool results have accumulated, those results are now joined into `resp.Answer` instead of being discarded. `Reply` still carries only the executable markup, so the existing split between executable text and prior tool output is preserved and `agent.executeAdditionalTextToolCalls` behaves exactly as before.

`ai/atlascloud/atlascloud_test.go`: locks down both halves of that response contract — a structured `conformance_echo` followed by a tagged-text `delegate` must leave the tagged call in `Reply` and the earlier echo result in `Answer`.

`agent/conformance_test.go`: the fake provider now models the real AtlasCloud shape (tagged delegate in `Reply`, echo marker in `Answer`, structured echo in `ToolCalls`) rather than concatenating the echo into `Reply`. The delegate assertion moved from a boolean to an exact call count, so a regression that double-executes the fallback fails instead of passing.

## Testing

- The AtlasCloud provider test asserts the mixed-response contract directly: tagged call retained in `Reply`, echo result carried in `Answer`.
- The agent conformance test asserts the delegate executes exactly once, both tool calls are recorded, and the final reply still contains the marker.
- A follow-up with no accumulated structured-tool content does not invent an answer, and ordinary non-text follow-up answers continue through the existing normalization path untouched.
- The deterministic fake-provider conformance matrix still rejects genuinely missing markers, so this does not weaken the check it fixes.

I could not run the Go suite in this workspace, so CI is the authority on the full run.

## Scope

Three files, +76/-13, and the only non-test change is the `Answer` assignment guarded by `len(toolResults) > 0`. Providers other than AtlasCloud are untouched.

AI was used for assistance.

### DIFF
--- a/agent/conformance_test.go
+++ b/agent/conformance_test.go
@@ -846,7 +846,7 @@ func TestAgentExecutesTextToolCallFallbackAfterStructuredToolCall(t *testing.T) 
 			Input: map[string]any{"value": "agent-conformance"},
 		})
 		return &ai.Response{
-			Reply:  echo.Content + "\n<tool_call name=\"delegate\">{\"task\":\"summarize the conformance marker\",\"to\":\"blocked-reviewer\"}</tool_call>",
+			Reply:  "<tool_call name=\"delegate\">{\"task\":\"summarize the conformance marker\",\"to\":\"blocked-reviewer\"}</tool_call>",
 			Answer: echo.Content,
 			ToolCalls: []ai.ToolCall{
 				{ID: "structured-echo-1", Name: "conformance_echo", Input: map[string]any{"value": "agent-conformance"}, Result: echo.Content},
@@ -856,7 +856,7 @@ func TestAgentExecutesTextToolCallFallbackAfterStructuredToolCall(t *testing.T) 
 	defer func() { fakeGen = nil }()
 
 	var sawTool bool
-	var sawBlockedDelegate bool
+	var delegateCalls int
 	a := New(
 		Name("conformance-mixed-text-tool"),
 		Provider("fake"),
@@ -865,7 +865,7 @@ func TestAgentExecutesTextToolCallFallbackAfterStructuredToolCall(t *testing.T) 
 		WithMemory(NewInMemory(4)),
 		ApproveTool(func(tool string, input map[string]any) (bool, string) {
 			if tool == "delegate" {
-				sawBlockedDelegate = true
+				delegateCalls++
 				return false, "cross-provider conformance blocks delegate side effects"
 			}
 			return true, ""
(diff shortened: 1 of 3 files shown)


asim:
Thanks for the contribution

mvanhorn:
Thanks @asim for landing the AtlasCloud conformance marker fix. Good to keep those markers stable across text updates.

PR #4943, opened 2026-09-21
fix(natsjs): prevent durable consumers from skipping history

### LINKED ISSUE #4929: [BUG] `events/natsjs` `Consume` silently skips history on first connect and defaults to a random group
**Labels:** bug, events, jetstream, footgun
**Package:** `go-micro.dev/v6/events/natsjs`
**Version:** v6.12.0 / v6.13.0
**Confidence:** source-verified (`events/natsjs/nats.go`) + observed (KNOWN_ISSUES #17, 8 events permanently missing)

## Summary

Two defaults combine into silent event loss:

1. When `ConsumeOptions.Offset` is the zero value, `Consume` applies `nats.DeliverNew()`. A durable consumer's first-ever
   connection receives only events published after that moment. Anything already on the stream is skipped and never delivered.
2. `Group` defaults to `uuid.New().String()` on every call. Without `WithGroup`, each restart creates a brand-new durable
   consumer, which per (1) also starts at "new only". Durability is lost without any warning.

## Evidence

```go
// events/natsjs/nats.go, Consume()
options := events.ConsumeOptions{ Group: uuid.New().String() }   // random per call
...
if !options.Offset.IsZero() {
    subOpts = append(subOpts, nats.StartTime(options.Offset))
} else {
    subOpts = append(subOpts, nats.DeliverNew())                 // silent skip of history
}
```

Live: two new consumer groups (`audit-communication-events`, `audit-billing-events`) connected against streams that already had
history. `psql` showed 1 of 11 published communication events recorded. The other events were never delivered, not delayed.
Nothing logged, nothing errored.

## Expected

- A stable, named consumer resumes from its persisted position, and a new one starts from the beginning of the stream (or from a
  documented, explicit default).
- `Consume` without a `Group` either fails or logs loudly instead of minting a throwaway durable consumer.

## Suggested fix

- Default to `DeliverAll` for durable named consumers and `DeliverNew` only for explicitly ephemeral ones. Or expose
  `events.WithDeliverPolicy(All|New|ByStartTime|Last)` and make the default explicit in the docs.
- Require `WithGroup` for durable streams, or log a warning when the random default is used.
- Document that a start offset only applies the first time a durable consumer name is created.

## Workaround here

Every `Consume` call passes `events.WithOffset(time.Unix(0, 0))` and a fixed `WithGroup`. JetStream only honours the offset when the
durable consumer is first created, so restarts resume from the persisted ack position. We also had to delete and recreate the
10 audit consumers to backfill.

asim:
## Summary
- require an explicit group for durable JetStream consumers instead of generating a throwaway durable name
- replay all retained events when a named durable consumer is first created while preserving new-only delivery for ephemeral consumers
- document and test the durable consumer defaults

## Testing
- `go build ./...`
- `go test ./...`
- `$(go env GOPATH)/bin/golangci-lint run ./...`

Closes #4929


asim:
@codex address the comments

